PR #3, opened 2017-05-08
Support for `danger` buttons.

jvalente:


henningmu:
Awesome. Thanks for integrating the @doist/reactist package :)

PR #14, opened 2017-05-30
Adds `large` prop to `Modal.Box`

jvalente:


src/components/styles/modal.less
```diff
@@ -28,10 +28,16 @@
...
+
+        &.large {
+            width: 60%;
+            max-width: 1000px;
+        }
 
         // only applies for devices smaller than a desktop (defined by bootstrap)
         @media(max-width: 992px) {
-            width: calc(~'100% - 120px'); // 60px on each side (incl. top via padding)
+            width: calc(~'100% - 120px') !important; // 60px on each side (incl. top via padding)
```
henningmu: Why do we need the `!important` here? It should be ok to change the width by overriding these styles, right?

jvalente: right, it is a better idea to add an explicit media query to `.large`. 👍

src/components/styles/modal.less
```diff
@@ -33,11 +33,14 @@
         &.large {
             width: 60%;
             max-width: 1000px;
+
+            @media @small_screen {
```
jvalente: @henningmu here I've added a dedicated `small_screen` media query for large. this way we avoid the override. 👍

PR #21, opened 2017-07-03
Component/dropdown

henningmu:


src/components/Dropdown.jsx
```diff
@@ -30,9 +30,10 @@ class Box extends React.Component {
 
     _handleClickOutside(event) {
         const dropdown_dom_node = ReactDOM.findDOMNode(this)
+
         if (!dropdown_dom_node.contains(event.target))
```
jvalente: @henningmu shouldn't this condition be enough to prevent the dropdown from closing when click are inside the dropdown?

henningmu: No :(
The first condition
```
if (!dropdown_dom_node.contains(event.target)) { ...
```
checks whether the user clicked outside the dropdown and if so the dropdown is closed.

The second condition
```
else if (!this.props.allowBodyInteractions) { ...
```
is the counterpart and checks whether the user clicked inside the dropdown and if so closed it, too. Although it waited 100ms to do so.

This commit prevents the dropdown from being closed when the body is clicked in case `allowBodyInteractions` is supplied.

In most cases (i.e. menus or details) it is totally fine and expected to close the dropdown when clicking on it. In some cases it's not (e.g. when you want to copy something from it)

PR #67, opened 2017-12-10
Commentary

jgradim:


src/index.js
```diff
@@ -42,3 +42,4 @@ export {
     Avatar,
     Icon
 }
+
```
jgradim: No need to `import` just to `export`. You can just `export X from './components/X`. Also it would be nice to sort the export alphabetically

henningmu: Good to know, I will give that a try. Thank you very much! 😊


My initial reasoning was to be able to do `import { X } from '@doist/reactist'` and wasn't sure if that works without an index.js exporting all components. Especially when it comes to duplicate names (e.g. two components exporting an `Item` as sub components)

jgradim: @henningmu I started this review before realizing sitting down with @jvalente and go through the code would be way more productive, trying to get that to happen :)

henningmu:
Hi @jgradim,

the change you requested is now live 👍
I'll close this PR as it triggers my OCD 😅 
However, if you've any more feedback I'm more than happy to hear it. Just open a new PR or issue :)

jgradim:
@henningmu 🤦‍♂️ I ended up postponing the review with @jvalente and forgot all about this! Thank you! :)

PR #71, opened 2018-01-10
Tabs: add new onChange props and also new 'spread layout' option

stkao05:
Hi @henningmu 

I would like to integrate Reactist for the first time for the new Todoist Karma UI 🎉 This PR contains some new features to the Tabs component needed for the new UI:

- New `onChange` Tabs props for tab change event. Each Tab could also have identifier props `value`
- Add a new tab layout style `spreadLayout`

Let me know if there is any further change you would like me to make 👍
  

screen shot 2018-01-10 at 3 43 19 pm


src/components/Tabs.jsx
```diff
@@ -7,19 +7,40 @@ import classNames from 'classnames'
 class Tabs extends React.Component {
     constructor(props, context) {
         super(props, context)
-        this.state = { activeTab: props.defaultTab || 0 }
+        const children = React.Children.toArray(this.props.children)
+
+        this.valueToTab = children.reduce((map, c, i) => {
```
henningmu: This won't work as soon as you add / remove children dynamically, right?

What do you think about adding a function `_getValueForTab(tabs /* children */, value)` which uses `children.find( ... )` to return the correct tab?

src/components/Tabs.jsx
```diff
@@ -7,19 +7,40 @@ import classNames from 'classnames'
...
+            const v = "value" in c.props ? c.props.value : i
```
henningmu: We're using single quotes in reactist. Running `npm run check` or even `npm run check -- --fix` should help here 👍

src/components/Tabs.jsx
```diff
@@ -7,19 +7,40 @@ import classNames from 'classnames'
...
         return children.map((child, index) => {
             const { title, disabled } = child.props
+            const value = child.props.value || index
             const className = classNames('reactist tabs__header--item',
-                { disabled, active: index === this.state.activeTab }
+                { disabled, active: value === this.state.activeTabValue }
             )
+
             return (
                 <a
```
henningmu: We should now use the `value` as key instead of the index correct?

src/components/Tabs.jsx
```diff
@@ -42,25 +63,44 @@ class Tabs extends React.Component {
...
+        let activeTab = null
+        // in case of invalid activeTab value, render first tab instead
+        if (children.length > 0) {
+            activeTab = this.valueToTab[this.state.activeTabValue] || children[0]
+        }
+
+        const cls = classNames(
+            "reactist",
+            "tabs",
+            this.props.spreadLayout ? "tabs-spreadlayout" : null
```
henningmu: Could we try to adhere to BEM and use `tabs--spreadlayout` here?

.storybook/stories/components/TabsStory.js
```diff
@@ -32,6 +35,7 @@ const TabsChapter = {
 // Story setup ================================================================
 const Story = () =>
 storiesOf('Tabs', module)
+    .addDecorator(withKnobs)
```
henningmu: Normally I tried to separate "playgrounds" (with knobs) from overview stories. However, I think it's fine in this case. Thansk for adding it 👍

src/components/Tabs.jsx
```diff
@@ -42,25 +63,44 @@ class Tabs extends React.Component {
...
+
 Tabs.propTypes = {
-    /** Index of the tab that should be selected on initial render. */
-    defaultTab: PropTypes.number
+    /** selects the tab whose value prop matches this prop */
+    defaultTab: PropTypes.oneOfType([
+        PropTypes.string,
+        PropTypes.number,
+    ]),
+    spreadLayout: PropTypes.bool,
```
henningmu: Could you add a jsdoc comment so a small description appears in the storybook a la `Whether the tabs should take all available space and distribute it evenly or use the minimum required`

stkao05:
Hi @henningmu 
Thanks for the feedback. Just made the above mention change. 👍 Let me know if I miss anything

src/components/Tabs.jsx
```diff
@@ -38,29 +52,56 @@ class Tabs extends React.Component {
         })
     }
 
+    _getTabForValue = (value) => {
+        const c = React.Children.toArray(this.props.children)
+        if (typeof value == 'number') return c[value] || c[0]
```
henningmu: I think this will break as soon as someone uses numeric values (e.g. ids) as `value` prop. I think there are two possible solutions:
1. We mark the `value` prop as required
2. We are cloning every child that doesn't have a `value` prop and set it to it's index

Advantage of *2* is that it's a non-breaking change but I'm a bit worried about possible performance implications if we cloning children on every render. 
What's your opinion? Or do you have a better idea?

stkao05: @henningmu 
What do you think about the solution: If `onChange` or `defaultTab` prop is set, each child is required to have `value` set. 

This way, we either get children by index or by matching value, but not both at the same time.

henningmu: Sounds like a plan 👍
Should we log / error to console in case this condition isn't satisfied? Seems to be too complex to be modeled with propTypes 😋

stkao05: @henningmu I tend to prefer throwing an exception so that developer could spot this error during development time. What do you think

stkao05:
Hi @henningmu
Just made another update 🙂 Did some refactoring to make it cleaner as well.

src/components/Tabs.jsx
```diff
@@ -7,28 +7,48 @@ import classNames from 'classnames'
 class Tabs extends React.Component {
     constructor(props, context) {
         super(props, context)
-        this.state = { activeTab: props.defaultTab || 0 }
+
+        const { defaultTab, onChange } = props
+        const children = React.Children.toArray(this.props.children)
+
+        if (defaultTab || onChange) {
```
henningmu: We should add something like
```
const isDefaultTabDefined = defaultTab || defaultTab === 0
```
and use that in the if-conditions otherwise it will result in unwanted behaviour

src/components/Tabs.jsx
```diff
@@ -7,28 +7,48 @@ import classNames from 'classnames'
...
+            const missing = children.find(c => !c.props.value)
+            if (missing) throw new Error('(Tab) Missing property: all Tab must have "value" set if "defaultTab" or "onChange" is used')
+        }
+
+        if (defaultTab) {
+            const c = children.find((x) => x.props.value === defaultTab)
```
henningmu: nitpick: This can be done in one call with `children.findIndex(...)`

stkao05: I think `findIndex` require some polyfill. Does Reactist in general requires use of polyfill?

henningmu: I think babel is taking care of that, right? You shouldn't need to import any other polyfills

src/components/Tabs.jsx
stkao05: Good catch 👍 just fixed it.

src/components/Tabs.jsx
stkao05: Hmm I think for new object types and prototype methods, you will need to include polyfill explicitly; the same way you need polyfill if you use Promise.

Anyway, I don't think it is a major issue since both Todoist/Twist already use polyfill anyway 🙂

PR #136, opened 2018-11-26
Customizable Loading component

jvalente:
Allows for custom colors and size on the `Loading` Component.


jvalente:
@henningmu did not generate the distribution assets since the tests are failing for me.

Issue #164, opened 2019-05-28
Tooltip Issue and Proposal


stkao05:
This issue outlines some issues with the current Tooltip component implementation, and also suggest some proposed change.

## Issues

__1. The Tooltip HTML ins't friendly for layout.__

Suppose you have a button which you want to add a tooltip.

```js
<div className="button_container">
    <Tooltip text={_("Edit")}>
        <EditButton />
    </Tooltip>
</div>
```

The Tooltip component will wrap the button component with its HTML.

```js
<div className="button_container">
    <span class="reactist popover__wrapper reactist tooltip__wrapper">
        <button />
        <span class="reactist popover" style="top: 276.25px; left: 1033.81px;">
        <span class="reactist popover__content reactist tooltip__text">Edit</span>
        </span>
    </span>
</div>
```

`button` is no longer a direct child of `button_container`, but rather the `tooltip__wrapper`. This could make CSS layout trickily in some scenario. Ideally, the tooltip should not insert extra HTML to the tooltip target and also its container.

__3. The HTML mockup structure is not friendly for browser rendering performance__

Each tooltip is placed beside the target element. 

The more optimize HTML structural will be to place tooltip as an immediate child of `<body>`
```js
<body>
    ....
<div className="tooltip">Edit</div>
</body>
```

__2. Tooltip is always rendered even when not show__

We want to support the lazy rendering of a tooltip - only render tooltip when it is showing, because some tooltip could be expensive to render.

For instance, Todoist's due date tooltip

<img width="658" alt="Screen Shot 2019-05-28 at 5 03 45 PM" src="https://user-images.githubusercontent.com/1556390/58465742-e6310700-816a-11e9-8abf-fad682d89dd3.png">
 

#### Proposal

- The Tooltip does not introduce any mockup to the target component and it will be rendered as an immediate child of `<body>`

- The Tooltip should be lazy rendered: it's only rendered when the target component is on-hovered. Should be great if Tooltip could accept a function() if lazy rendering is needed

```js
<Tooltip content={renderExpensiveComponent}>
</Tooltip>
```

Reference: [Material UI Tooltip](https://material-ui.com/components/tooltips)

henningmu:
I fully agree with the outlined problems and proposed solution. 
One thing: this should be fixed on the level of the underlying `Popover` component ideally the Tooltip component wouldn't even need to change 👍

stkao05:
[...]

PR #178, opened 2019-09-28
Tooltip with portal (WIP)

frankieyan:
**Link to issue:** #164 

**Short description:**

Hey @stkao05, I hope you don't mind that I took a stab at this. This attempts to address some of the issues you've listed in #164 but it will introduce some breaking changes depending on how it's being used in your apps right now.

These are some of the changes made so far:
- The popover now uses `React.createPortal` to attach its content to the `body` element as you suggested
- No extra wrappers are added around trigger elements unless we encounter a plain text node
- Event handlers are attached directly to the trigger element

As the Popover used to depend on its wrapper's ref in order to calculate the correct placement for its content, which is no longer available for components that do not forward their underlying DOM elements as refs. So when we use other components in this library as our trigger, we receive their component instance instead. [Material-ui also has the same limitation](https://material-ui.com/guides/composition/#caveat-with-refs), but they've forwarded refs within all of their own components. In order to continue support for other components within this library to be used as triggers, we'll most likely need to do the same thing here.

Since the tooltip is no longer a sibling to the trigger, we'll likely want to shift the focus to the tooltip content on trigger, or attach an aria label to the triggering element to maintain accessibility too.

I'm not sure if this component is still being used internally in Todoist, as I've noticed the class names used in the tooltip over there is quite different. Let me know if this is still something you'd like to push forward or have any suggestions! I'd love to help out with this if it's still needed :)

**PR Checklist:**

-   [ ] Described changes in CHANGELOG.md
-   [x] Executed `npm run check` and made sure no errors / warnings were shown
-   [ ] Bumped version in package.json

### DIFF
--- a/.storybook/stories/components/TooltipStory.js
+++ b/.storybook/stories/components/TooltipStory.js
@@ -29,7 +29,10 @@ const TooltipPropTypesChapter = {
 
 const TooltipStory = () => (
     <section className="story tooltip">
-        <Tooltip text="Very helpful content in this tooltip">
+        <Tooltip
+            text="Very helpful content in this tooltip"
+            tooltipClassName="storybook-tooltip-content"
+        >
             <div className="tip_item">
                 Hover me for an automatically positioned Tooltip
             </div>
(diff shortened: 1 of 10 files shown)


frankieyan:
Closing this in favour of using https://github.com/reach/reach-ui as the backing library for the tooltip

PR #179, opened 2019-11-07
Automatic Time refresh

jvalente:
**Short description:**

Added a `refresh` prop to `<Time />` component in order to keep the information shown up-to-date when the component displays "time ago". By default, the component will re-render every 60 seconds.

**PR Checklist:**

-   [x] Described changes in CHANGELOG.md
-   [x] Executed `npm run check` and made sure no errors / warnings were shown
-   [x] Bumped version in package.json


src/components/Time.jsx
```diff
@@ -43,6 +65,12 @@ class Time extends React.Component {
         return TimeUtils.timeAgo(this.props.time, config)
     }
 
+    _refresh() {
+        this.refresh_timeout = setTimeout(() => {
+            this.setState(() => this.state)
+        }, DELAY)
```
gnapse: Isn't the intention that this component continually refreshes the time? If so I think we need to use `setInterval` rather than `setTimeout`.

src/components/Time.jsx
```diff
@@ -43,6 +65,12 @@ class Time extends React.Component {
         return TimeUtils.timeAgo(this.props.time, config)
     }
 
+    _refresh() {
+        this.refresh_timeout = setTimeout(() => {
+            this.setState(() => this.state)
```
gnapse:
```suggestion
            this.forceUpdate()
```

src/components/Time.jsx
```diff
@@ -7,17 +7,39 @@ import Tooltip from './Tooltip'
...
+        }
+    }
+
+    componentDidUpdate() {
+        clearTimeout(this.refresh_timeout)
+
+        if (this.props.refresh) {
+            this._refresh()
+        }
+    }
```
gnapse: We only want to do this if the `refresh` prop changed from `false` to `true`, and conversely to do the opposite if it changed from `true` to `false`:

```js
componentDidUpdate(prevProps) {
  if (!prevProps.refresh && this.props.refresh) {
    this._refresh()
  }
  if (prevProps.refresh && !this.props.refresh) {
    clearTimeout(this.refresh_timeout)
  }
}
```

src/components/Time.jsx
jvalente: I don't like using `forceUpdate` since it skips `shouldComponentUpdate`. Seems like a bit of an anti-pattern. In any case, I leave it the final call to you guys. @henningmu @gnapse.

jvalente:
Thank you very much for the review @gnapse! 🙏
I've updated the PR accordingly.

src/components/Time.jsx
gnapse: If this component had a `shouldComponentUpdate`, which it does not, would you want it to prevent this update? I don't think that's the case. You want this component to refresh at the given interval no matter what.

I don't like to use it in general because it is a signal that the component's output depends on something that's not part of its props or state (meaning is not pure), but in this case this component isn't anyway.

But I too have no hard feelings about it. It's only that reading `setState(this.state)` made me think for a couple of seconds, while `forceUpdate` makes it more explicit the intention.

jvalente: Agreed. `forceUpdate` represents better what `refresh` does, a forced update that doesn't relate to state change. 👍

jvalente:
All the issues addressed.

jvalente:
All the issues addressed. 
Ready for a final evaluation @henningmu @gnapse. 👍

PR #181, opened 2019-12-19
Remove use of style-loader. Output css file instead

stkao05:
In preference of enforcing a strong Content Security Policy, the PR remove usage of inline-style and require CSS file to be included for the library usage.


README.md
```diff
@@ -17,8 +17,21 @@ You can add Reactist to your project by installing it from npm:
...
+To import a component within your code:
+
+```js
+import loading from "@doist/reactist/lib/loading"
+```
+
+You also need to load the CSS file of the component somewhere in your app. For instance:
+
+```
+<link rel="stylesheet" type="text/css" href="./node_modules/@doist/reactist/lib/Loading.css">
```
henningmu: Maybe we also mention that there's one `reactist.css` that includes css for all components?

stkao05:
@henningmu  Added minification to the CSS. 

I also made another change to the build setting to only apply minification for the `dist/` build and not the build output from `lib/`. It seems like most modules don't minify the codes because minification could be taken cared for by the library user's bundling tool. What you think?

stkao05:
> Will release this as version 2 (breaking change) tomorrow if you have no further reservations

@henningmu Great. Don't have anything else to add here. 
Will let you handle the merging of this PR then 👌

PR #186, opened 2020-02-26
Popper and TooltipNew component

stkao05:
The PR implements a new tooltip component TooltipNew to address [issues](https://github.com/Doist/reactist/issues/164) with the existing Tooltip. A new component is introduced here because we cannot fix the issue in a backward-compatible way with the existing Tooltip, as the resulting HTML has also changed.

The PR also fork the Popper component we use internally in Todoist for the tooltip implementation.

### DIFF
--- a/.storybook/stories/components/PopperStory.js
+++ b/.storybook/stories/components/PopperStory.js
@@ -1,0 +1,39 @@
+import React, { useState, useRef } from 'react'
+import { storiesOf } from '@storybook/react'
+import Popper from '../../../src/components/Popper'
+
+function PopperStory() {
+    const btn = useRef()
+    const [isShown, setIsShown] = useState(false)
+
+    return (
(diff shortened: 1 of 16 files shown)


gnapse:
I once recommended in the referenced issue that we could use a better, more accessible alternative. One that, for instance, also appears when you focus on the element, and not just on hover. See https://github.com/Doist/reactist/issues/164#issuecomment-555685065

The objections raised against it at the time no longer apply. And I'd rather use that. Also, not sure how convenient is this approach of introducing a new Tooltip component with a different name. If we insisted in going down the path of developing our own, I'd rather replace Reactist's tooltip with this one you're adding here, and issue a breaking changes release instead.

stkao05:
Close as we have concluded to use https://reacttraining.com/reach-ui/tooltip/

PR #184, opened 2020-02-07
CSS namespace

Nazeh:
**Link to issue:** https://github.com/Doist/reactist/issues/174

**Short description:**

 Changed namespacing from `.reactist.popover {}` to `.reactist_popover {}`.

**PR Checklist:**

-   [ ] Added tests for bugs / new features
-   [x] Updated docs (storybooks, readme)
-   [x] Described changes in CHANGELOG.md
-   [x] Executed `npm run check` and made sure no errors/warnings were shown
-   [x] Executed `npm run test` and made sure all tests are passing
-   [x] Bumped version in package.json
-   [x] Updated all static build artifacts (`npm run build-all`)

PR #187, opened 2020-03-04
Nazeh/css namespace

henningmu:
Kinda failed to push to #184 so this one closes #184 

@Nazeh thank you for the contribution. It goes live as version 3.0.0 in a sec

Celebrations


Nazeh:
Thanks @henningmu 🎉

PR #188, opened 2020-04-22
Typescript support

proxi:
Adds exportable TypeScript type declarations. It doesn't aim to fix all the problems and tsconfig is non-strict, but the problems do not prevent successfully exporting types.


henningmu:
How much more effort would it be to rewrite this to "real" TS over type annotations in JS?

 Since that's the direction we want to go anyway it seems unnecessary to touch each file twice 👍

gnapse:
FWIW, I've been thinking of porting Reactist to use tsdx.

proxi:
> How much more effort would it be to rewrite this to "real" TS over type annotations in JS?

The problem is that there are issues present that would prevent TS from compiling and I don't want to change semantics of existing code. Personally I really don't see much if any difference between jsdoc vs .ts and so I don't see much value in converting something that works.

This PR is now ready - adds all types - but doesn't try to fix all the internal issues. Could someone knowledgeable with npm help me finish to to publish the types? Running `npx tsc` will create `dist/index.d.ts` with the types. It already works with `npm link @doist/reactist`.

proxi:
BTW conversion from annotated js to ts can probably be done automatically, for example with [gents](https://github.com/angular/clutz#gents---closure-to-typescript-converter). Ref. https://www.lucidchart.com/techblog/2017/11/16/converting-600k-lines-to-typescript-in-72-hours/

frankieyan:
> The problem is that there are issues present that would prevent TS from compiling

I think one of the biggest disadvantages in having to ensure everything goes through the compiler is that it delays us from turning on strict mode, meaning any new code we write may not adhere to it and would be building on tech debt. 

That said, since Reactist is considerably smaller than our apps with less inter-dependencies - maybe we can one-shot this all the way to strict mode using JSDoc only. I'll take some time to take a look at this tomorrow 👍

proxi:
> maybe we can one-shot this all the way to strict mode using JSDoc only

Of course we can. I fixed some issues already here, the remaining ones are mostly with prop types or obvious missing null checks.

proxi:
I have added `npm run declaration:build` script and fixed all the issues against Twist (found [this](https://github.com/Doist/twist-web/pull/2082)).

src/components/Avatar.jsx
```diff
@@ -36,7 +50,7 @@ const Avatar = ({
     colorList = AVATAR_COLORS
 }) => {
     const userInitials = getInitials(user.name) || getInitials(user.email)
-    const avatarSize = AVATAR_SIZES.includes(size) ? size : 'l'
+    const avatarSize = size && AVATAR_SIZES.includes(size) ? size : 'l'
```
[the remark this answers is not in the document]
proxi: The check is required because array of non-undefined can't include undefined:
> Type 'undefined' is not assignable to type '"xxs" | "xs" | "s" | "m" | "l" | "xl" | "xxl" | "xxxl"'.ts(2345)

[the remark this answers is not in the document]
gnapse: I wasn't referring to the added `size &&`. I was referring to the fact that we can remove this line altogether. What this line is doing (checking that the provided `size` prop is one of the accepted valid values) is already provided to us by the type checking mechanism.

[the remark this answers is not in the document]
proxi: Ah. But this is also untrue because of https://medium.com/@martin_hotell/react-typescript-and-defaultprops-dilemma-ca7f81c661c7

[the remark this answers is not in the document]
gnapse: That article refers to when you have class components, but in function components there's no need to use `defaultProps`. The default values for props are better expressed in destructuring right in the arguments, as done in this component.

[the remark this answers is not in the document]
proxi: Try with `strict: true` and you'll see what I mean. I don't see how this could be fixed.

package.json
```diff
@@ -6,6 +6,7 @@
     "scripts": {
         "build": "NODE_ENV=production ./node_modules/.bin/webpack --progress --color",
         "build-watch": "npm run build -- --watch",
+        "declaration:build": "rm -f dist/**/*.d.ts lib/**/d.ts && npx tsc -p tsconfig.lib.json; npx tsc -p tsconfig.index.json; mv dist/index.d.ts dist/reactist.d.ts",
```
frankieyan, line 9: Let's stick with the current pattern of `build-x` for these npm script names

tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+        "moduleResolution": "Node",
+        "noImplicitReturns": true,
+        "jsx": "react",
+        "allowSyntheticDefaultImports": true,
+        "skipLibCheck": true,
+        "baseUrl": ".",
+        "declaration": true,
+        "emitDeclarationOnly": true,
+        "noImplicitAny": true,
+        "strict": false
```
frankieyan, line 16: Can we add the strict rules that we haven't enabled but leave them commented out? It'd be clearer what they are and what we need to work towards

package.json
frankieyan: Also, can we add a ["types" field](https://www.typescriptlang.org/v2/docs/handbook/declaration-files/dts-from-js.html#editing-the-packagejson) for completeness? TS is able to find reactist.d.ts but it always helps to be explicit 👍

proxi:
> Before we push this through, can we ensure that the d.ts files build without errors? It looks like most of them just need further narrowing of the types

It doesn't seem so to me, most problems are with prop types. I don't know enough about proptypes or reactist to fix them and I think it might be hard to fix them without changing external semantics.

proxi:
I've created https://github.com/Doist/reactist/pull/190 to fix constants issue. There are 14 other issues left.

package.json
proxi: > Let's stick with the current pattern of `build-x` for these npm script names

Renamed to `build-types`. Please use github "suggestion" for these simple fixes.

proxi: > Also, can we add a ["types" field](https://www.typescriptlang.org/v2/docs/handbook/declaration-files/dts-from-js.html#editing-the-packagejson) for completeness? TS is able to find reactist.d.ts but it always helps to be explicit 👍

TBH I don't understand why we have two different kinds of output and how this relates. it we include it won't it mess up `lib` usage?

frankieyan: We don't use the "libs" dist build in Twist or Todoist, but you can also consume the library like `import { Avatar } from "@doist/reactist", and that will tell webpack/typescript to point to the "main" and "types" fields in package.json. The difference is the entire library will end up being bundled into the app's output.

proxi:
Hey team, I fixed most of the issues, except 2 I need your help with:

- Checkbox label proptype is messed up. I don't understand what it was supposed to be, but given that it's passed to input.value, I suspect only a string is valid. Or maybe passing it to value is a mistake?

- in Modal: event.target.id, id does not exist on EventTarget. I believe a valid fix would be to check `instance of Element`, but it will break the test. It's quite a common problem that we use partial dummy objects in tests. Can we somehow mock target so that it passes Element instanceof check?

src/components/Checkbox.jsx
```diff
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types'
  * @property {boolean | undefined} [checked]
  * @property {boolean | undefined} [disabled]
  * @property {(checked: boolean) => void} [onChange]
- * @property {string | React.ReactNode | React.ReactNode[]} [label]
+ * @property {string} [label]
```
gnapse, line 11: Why. Seems we loose some versatility here.

proxi: For the reason I described above (input valye can't be ReactNode nor ReactNode[]). I've no idea what the original intention could have been, open to suggestions.

gnapse: Yup, I also do not understand what does it mean for the value to be assigned the label. I think there may be a bug here.

proxi: @henningmu please?

henningmu: We only use string 👍

proxi: @henningmu So proptypes are wrong and the correct fix is to limit them to string, right?

package.json
proxi: According to documentation it will use `main` field, no need to duplicate it.

proxi:
> in Modal: event.target.id, id does not exist on EventTarget. I believe a valid fix would be to check instance of Element, but it will break the test. It's quite a common problem that we use partial dummy objects in tests. Can we somehow mock target so that it passes Element instanceof check?

This is the last problem, please help.

src/components/Checkbox.jsx
henningmu: Correct

gnapse: I agree that this PR should be concerned with what we have now, so not meaning to say we should change this here, but...

Why do we pass the checkbox label as its `value` attribute? To the best of my knowledge, I think that the `value` attribute in a `<input type="checkbox" />` element is usually not needed, and it is only used when you submit a form via the browser submit mechanism. But in a SPA we rarely, if ever, do that.

From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/checkbox#Value)

> This is never seen on the client-side, but on the server this is the `value` given to the data submitted with the checkbox's `name`.

Could we stop passing it as the `value`, and make it a `React.ReactNode`?

proxi: @gnapse How about you open another issue about it? Let's focus on getting this shipped here.

gnapse:
It is safe to add a check `event.target instanceof Element` to that condition.

> but it will break the test

The problem is not this code, it is the test. It is testing an implementation detail.

I pushed a commit fixing it, but the fix involved more changes in code than might have been expected for a "fix/add type definitions" kind of PR. However, I did not see a way around it other than doing a rather hacky thing in the tests. Which I did not.

proxi:
Thank you @gnapse Looks okish to me, but someone with better understanding of code has to review it.

@Doist/frontend The PR now builds (`npm run build-types`). Can you all please review it?

Can someone with knowledge of npm publishing process advise on how to make sure types will be visible in the imported package?

henningmu:
@proxi Released as 3.1.0 🙌

PR #191, opened 2020-05-05
Fix label proptype to allow number.

proxi:
We discussed it [here](https://github.com/Doist/reactist/pull/188#discussion_r419493977) but the conclusion was wrong: based on its usage in Twist-Web (in ConfirmDialog) the label type should also allow number as value.


proxi:
Fixed to also allow undefined (as preliminary step to fix strict mode).

proxi:
@henningmu @gnapse Just a reminder that this fix and bumping twist-web reactist version to include types will be needed to complete typing effort next week.

gnapse:
I still maintain it can and should be a `React.ReactNode`. But I am not saying this is blocking anything, just saying.

proxi:
@gnapse Please open a separate issue so we don't lose track of it.

PR #194, opened 2020-05-11
Remove legacy context parameter.

proxi:
@henningmu I don't understand exactly why `context` parameter messes up `ComponentProps`, but it seems that we don't need it. After removing it in Twist `timeConfig?: React.ComponentProps<typeof Time>['config']` will work as expected.


proxi:
> Looks good, will release as 3.1.2

It is very cumbersome having to release for each little change. Isn't there any way around it?

henningmu:
As I told you in the twist-web repo you can link reactist into twist-web and use the local version 👍

PR #196, opened 2020-05-19
Rework reactist to be based on TSDX

vishnugopal:
**Short description:**

This reworks reactist to be based on tsdx, getting a lot of typescript sweetness in the process. This is not yet release-ready, but should achieve current feature parity.


tsconfig.json
```diff
@@ -1,20 +1,27 @@
...
-        "checkJs": true,
-        "target": "es6",
-        "moduleResolution": "Node",
+        "module": "esnext",
+        "lib": ["dom", "esnext"],
+        "importHelpers": true,
+        "declaration": true,
+        "sourceMap": true,
+        "rootDir": "src",
+        "strict": false,
```
proxi, line 11: Hi @vishnugopal Did you consider enabling `strict` first, perhaps independent of this PR? This is where the biggest value is.

vishnugopal: Yup I did, there's too many errors to fix quickly with `strict: true`. I'll definitely tackle them in a future PR. It's easier to find these now once I get the infrastructure rolling.

gnapse:
I second @henningmu's comment about everything under lib. I'd also expect things to adapt more to tsdx's default folder structure, which if I recall correctly, it's to have things under `src/`, isn't it? But anyway, this is still a great first step. I will try to make some time to try this out locally with Twist tomorrow.

@henningmu any tips on how to better try a local twist-web with this checked-out version of Reactist also found locally? Is it enough to have this repo as sibling of twist-web, check out this branch, and then run some npm script command in twist-web?

henningmu:
> @henningmu any tips on how to better try a local twist-web with this checked-out version of Reactist also found locally? Is it enough to have this repo as sibling of twist-web, check out this branch, and then run some npm script command in twist-web?

Yes, that's exactly the process 👍
You can run `npm run build-watch` in the reactist repo to get live updates even 👍

vishnugopal:
> I think there are way too many things ending up in the `lib` directory, e.g. stories and tests doesn't seem to be needed there 👍

I've removed extra files from here, thanks for the find! 🙂

> I'd also expect things to adapt more to tsdx's default folder structure, which if I recall correctly, it's to have things under src/, isn't it? 

Because of the way Twist imports reactist, there's no easy way to remove lib/ entirely at least at this stage without changing Twist too. I also had to bring part of the webpack compilation back so that lib would be re-generated correctly under this new build. 🤷🏽‍♂️

One of my next PRs would be to change Twist to import these files from dist/ instead, that will make sure we get all the advantages that a tsdx build brings to the table. But with the philosophy that we go with parity first, I've made no breaking changes here. This is now ready for a re-review!

henningmu:
>  next PRs would be to change Twist to import these files from dist/ instead,

What does this mean? I think preferably we'd import like
```js
import { Button } from '@doist/reactist'
```
or something like
```js
import Button from '@doist/reactist/Button'
```
Is your proposal
```js
import Button from '@doist/reactist/dist/Button'
```
?

package.json
```diff
@@ -1,28 +1,46 @@
 {
     "name": "@doist/reactist",
-    "version": "3.1.3",
     "description": "Open source React components by Doist",
-    "main": "./dist/reactist.js",
+    "author": "Henning Muszynski <henning@doist.com> (http://doist.com)",
+    "version": "4.0.0",
```
henningmu, line 5: btw: do you want to work against a v4 branch and do all changes there or release multiple (major) versions as you go?

vishnugopal:
> `import { Button } from '@doist/reactist'`

Should work! With the `Button` component then responsible for importing its own assets & styles, or perhaps as a stop-gap until we properly do CSS modules, a single reactist CSS import. It also should support tree-shaking, and its also how [antd](https://ant.design/docs/react/introduce) does its imports.

package.json
vishnugopal: I want to make sure every release we do is compatible with Twist, so yeah, that might mean multiple releases. If we follow semver properly, I think this one should be a major one as we're thinking of changing the external API surface, but perhaps the next few need not be.

Before I cut a release though, I want to test a Twist PR that changes imports.

vishnugopal:
Just a note that for this to work, we'll probably need to change package.json's `main` to point to the `dist/index.js` file.

You can test this now by opening up a node repl, and doing:

```
const { Button } = require("./dist/index")
```

vishnugopal:
Am merging this now as I'd like to proceed with a release perhaps today/tomorrow and synchronise the Twist changes too. Please review when able as I'll definitely address any feedback in later PRs!

PR #197, opened 2020-05-22
Make a 4.0 release

vishnugopal:
**Short description:**

This is a release PR to make a 4.0 release to change the way files are imported in Twist. A few types have been added here to make Twist error-free.

This is still WIP!

**PR Checklist:**

-   [x] Added tests for bugs / new features
-   [x] Updated docs (storybooks, readme)
-   [x] Described changes in CHANGELOG.md
-   [x] Executed `npm run check` and made sure no errors / warnings were shown
-   [x] Executed `npm run test` and made sure all tests are passing
-   [x] Bumped version in package.json
-   [x] Updated all static build artifacts (`npm run build-all`)

PR #199, opened 2020-05-25
Feature/strict types 1

vishnugopal:
This is a sub-PR for:
https://github.com/Doist/reactist/pull/197

that does part of the typescript `strict: true` conversion for a few components.


src/components/__tests__/Tooltip.test.tsx
```diff
@@ -188,17 +199,18 @@ describe('Tooltip', () => {
...
         })
     })
 
     it('sets ref of tooltip and wrapper after mounting', () => {
-        const tooltip = mount(getTooltip()).instance()
+        const tooltip = mount(getTooltip()).instance() as Tooltip
 
-        expect(tooltip.wrapper).toBeTruthy()
-        expect(tooltip.tooltip).toBeTruthy()
+        expect((tooltip as any).wrapper).toBeTruthy()
```
proxi, line 212: I think in tests it makes more sense to disable checking for the entire file than do to incorrect casts.

src/components/Dropdown.tsx
```diff
@@ -185,51 +189,23 @@ class Box extends React.Component<any, any> {
...
 
-    /**
-     * @param {React.MouseEvent} event
-     */
-    _onClick(event) {
+    _onClick(event: React.MouseEvent) {
         event.preventDefault()
         event.stopPropagation()
-        this.props.onClick(event)
+        this.props.onClick && this.props.onClick(event)
```
proxi, line 208: I think it is more readable to use regular `if`s.

proxi:
@vishnugopal I would consider disabling typechecking for test files (or parts of them) if they are hard to fix, it's better than introducing broken typechecking with `any` casts.

In regular code there should not be any `as any` casts, they introduce typeholes and I believe are always wrong. If you have difficulty typing something, please ask the team, maybe we can come up with a proper solution. The eventual goal should be to not have any `any` :) in the code.

src/components/__tests__/Tooltip.test.tsx
vishnugopal: I really don't want to do this as it's just a few lines that are being casted here, and we lose type safety for the entire file because of that. I'll try to rework the Tooltip file to get proper types. I tried to go down that rabbit hole, but Tooltip has a dependency on Popover etc, and then I end up reworking a lot more files than I initially wanted to be in scope.

vishnugopal:
> In regular code there should not be any `as any` casts, they introduce typeholes and I believe are always wrong. If you have difficulty typing something, please ask the team, maybe we can come up with a proper solution. The eventual goal should be to not have any `any` :) in the code.

I agree with this completely @proxi, it's just that it's easy later on to identify these sniffs (let's say there are 20 instances of type-casted `any` in our code), a small linter rule change can identify these `explicitAnys`, and then we can go ahead and fix them. 

I think it's a philosophical discussion here as well, as I'd rather prefer to make small changes (even if parts of it are sub-optimal), rather than go ahead and make a 100% no-niggles refactor that might take more time. 

But since we already agreed to merge the parent branch only with `strict: true`, I'll ahead and make these changes here 🙂

src/components/__tests__/Tooltip.test.tsx
proxi: You can use @ts-ignore or @ts-expect-error (better) so that it works without introducing incorrect types.

src/components/Dropdown.tsx
vishnugopal: Changed!

src/components/Dropdown.tsx
```diff
@@ -2,37 +2,31 @@ import './styles/dropdown.less'
...
+
+interface BoxState {
+    top: boolean
+    show_body: boolean
+}
+
+class Box extends React.Component<BoxProps, BoxState> {
+    public static displayName: string
+
+    constructor(props: BoxProps, context: React.Context<any>) {
```
proxi, line 29: @vishnugopal We've discussed it in the past and it seems that if `context` is unused we can omit it. If it is needed, perhaps it can be `Context<unknown>` (safer)?

vishnugopal: Makes sense! Done!

vishnugopal:
> IMO Props types should not be exported. You can always get Props type via `React.ComponentProps<C>` without polluting type namespace.

That's pretty cool 🙂 Didn't know that!

src/components/ErrorMessage.tsx
```diff
@@ -84,17 +85,6 @@ ErrorMessage.displayName = 'ErrorMessage'
...
-        PropTypes.arrayOf(PropTypes.node),
-        PropTypes.node,
-    ]),
-    /** Timeout after the error message disappears (in ms). */
-    timeout: PropTypes.number,
-    /** Optional callback that is invoked when the error message disappears. */
-    onHide: PropTypes.func,
-}
 
+export { Props, State }
```
proxi, line 89: As I already wrote somewhere, I think `Props` types should not be exported. They can always be extracted with `React.ComponentProps` helper.

State is internal so it also should not be exported?

src/components/utils/__tests__/PositioningUtils.test.ts
```diff
@@ -164,6 +164,7 @@ describe('PositioningUtils', () => {
         })
         it('returns the wrapper position if position for calculation is invalid', () => {
             const position = calculatePosition(
+                // @ts-expect-error we are deliberately calling an invalid value.
```
proxi, line 167: Beautiful :)

src/components/utils/TextUtils.ts
```diff
@@ -1,7 +1,7 @@
 /**
```
proxi, line 1: You can / should remove `/**` JSDoc comment if they are no longer useful (like in this case).

src/components/Popover.tsx
```diff
@@ -110,15 +133,13 @@ class Popover extends React.Component<any, any> {
...
-        /** @type {("left" | "right" | "top" | "bottom")[]} */
-        const positionsToTry =
-            position === 'auto'
-                ? ['top', 'right', 'bottom', 'left', 'top']
-                : position === 'vertical'
-                ? ['top', 'bottom']
-                : position === 'horizontal'
-                ? ['left', 'right']
-                : [position]
+        const positionsToTry = ((position === 'auto'
```
proxi: `const positionsToTry: (RelativePosition | undefined)[] =`

`undefined ` here could be a mistake, is `position` really optional?

vishnugopal: Agreed, position can never be undefined.

src/components/ErrorMessage.tsx
vishnugopal: Done!

src/components/utils/TextUtils.ts
vishnugopal: Done!

src/components/Avatar.tsx
```diff
@@ -27,25 +26,31 @@ const AVATAR_COLORS = [
...
+    className?: string
+    colorList?: string[]
+    size?: AvatarSize
+    avatarUrl?: string
+    user: { name?: string; email?: string }
+}
 
-/** @type {React.FC<Props>} */
-const Avatar = ({
+const Avatar: React.FC<Props> = ({
```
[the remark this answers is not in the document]
vishnugopal: I have done this for the 3 affected components & will make sure this is the style followed for the rest. Is this more a stylistic choice or is there any other code advantage?

src/components/Avatar.tsx
```diff
@@ -27,25 +26,31 @@ const AVATAR_COLORS = [
...
 
-/**
- * @typedef {Object} Props
- * @property {string | undefined} [className]
- * @property {string[] | undefined} [colorList]
- * @property {AvatarSize | undefined} [size]
- * @property {string | undefined} [avatarUrl]
- * @property {{name?: string, email?: string}} user
- */
+interface Props {
```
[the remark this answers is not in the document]
vishnugopal: Done!

src/components/Avatar.tsx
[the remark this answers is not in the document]
proxi: I think style we use in Twist is more readable and less typing:

```
const Avatar = ({
    user,
    avatarUrl,
    size = 'l',
    className,
    colorList = AVATAR_COLORS,
}: Props) => {
```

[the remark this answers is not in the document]
vishnugopal: I noticed there is some hoisting advantages as well, and there is a linter rule too in case we'd like to enable it: https://eslint.org/docs/rules/func-style

vishnugopal:
I’m going to merge this back to the parent branch today and open a fresh PR for today’s progress. Hope that’s ok!

PR #200, opened 2020-05-26
Feature/strict types 2

vishnugopal:
This continues the work [started here](https://github.com/Doist/reactist/pull/199) of making more types strict and noImplicitAny.

### DIFF
--- a/src/components/Select.tsx
+++ b/src/components/Select.tsx
@@ -15,14 +15,14 @@ interface Option {
 }
 
 interface Props {
-    className: string
-    disabled: boolean
+    className?: string
+    disabled?: boolean
     /** Currently selected value. */
-    value: string | number
+    value?: string | number
     /** Callback for the change event. Will be called with the next value (not the full event). */
     onChange?: (value: Props['value']) => void
     /** Options that are rendered in the select. */
-    options: Option[]
+    options?: Option[]
 }
 
 function Select({
--- a/src/components/Select.tsx
+++ b/src/components/Select.tsx
@@ -1,28 +1,31 @@
 import './styles/select.less'
 
 import React from 'react'
-import PropTypes from 'prop-types'
 import classNames from 'classnames'
 
-/**
- * @typedef {Object} Option
- * @property {string | number | undefined} [key]
- * @property {string | number} [value]
- * @property {string | number | undefined} [text]
- * @property {boolean | undefined} [disabled]
- */
+interface Option {
+    /** Optional key for each option. If not provided `value` is used. */
+    key?: string | number
+    /** Value of the option. */
+    value: string | number
+    /** Text to display for the option. */
+    text?: string | number
+    /** Whether the options is disabled or not. */
+    disabled?: boolean
+}
 
-/**
- * @typedef {Object} Props
- * @property {string | undefined} [className]
- * @property {boolean | undefined} [disabled]
- * @property {string | number} value
- * @property {(value: Props['value']) => void} onChange
- * @property {Option[] | undefined} [options]
- */
+interface Props {
+    className?: string
+    disabled?: boolean
+    /** Currently selected value. */
+    value: string | number
+    /** Callback for the change event. Will be called with the next value (not the full event). */
+    onChange?: (value: Props['value']) => void
+    /** Options that are rendered in the select. */
+    options?: Option[]
+}
 
-/** @type {React.FC<Props>} */
-const Select = ({ value, options, onChange, disabled, className }) => {
+function Select({ value, options, onChange, disabled, className }: Props) {
     const selectClassName = classNames(
         'reactist_select',
         { disabled },
@@ -32,18 +35,21 @@ const Select = ({ value, options, onChange, disabled, className }) => {
         <select
             className={selectClassName}
             value={value}
-            onChange={(event) => onChange(event.target.value)}
+            onChange={(event) =>
+                onChange ? onChange(event.target.value) : undefined
+            }
             disabled={disabled}
         >
-            {options.map((option) => (
-                <option
-                    key={option.key || option.value}
-                    value={option.value}
-                    disabled={option.disabled}
-                >
-                    {option.text}
-                </option>
-            ))}
+            {options &&
+                options.map((option) => (
+                    <option
+                        key={option.key || option.value}
+                        value={option.value}
+                        disabled={option.disabled}
+                    >
+                        {option.text}
+                    </option>
+                ))}
         </select>
     )
 }
@@ -51,30 +57,6 @@ Select.displayName = 'Select'
 Select.defaultProps = {
     options: [],
     disabled: false,
-}
-Select.propTypes = {
-    /** Currently selected value. */
-    value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired,
-    /** Callback for the change event. Will be called with the next value (not the full event). */
-    onChange: PropTypes.func.isRequired,
-    /** Options that are rendered in the select. */
-    options: PropTypes.arrayOf(
-        PropTypes.shape({
-            /** Optional key for each option. If not provided `value` is used. */
-            key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-            /** Value of the option. */
-            value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
-                .isRequired,
-            /** Text to display for the option. */
-            text: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
-            /** Whether the options is disabled or not. */
-            disabled: PropTypes.bool,
-        })
-    ),
-    /** Whether the select is disabled or not. */
-    disabled: PropTypes.bool,
-    /** Additional css class applied to the select. */
-    className: PropTypes.string,
 }
 
 export default Select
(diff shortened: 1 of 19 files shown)


src/components/Tabs.tsx
```diff
@@ -97,9 +92,9 @@ class Tabs extends React.Component<any, any> {
 
     render() {
         // ensures that single or no child components don't throw
-        const children = /** @type {React.Component<TabProps>[]} */ React.Children.toArray(
+        const children = React.Children.toArray(
             this.props.children
-        )
+        ) as React.Component<TabProps>[]
```
vishnugopal, line 97: I don't think there's any way to extract the React Component type from the JSX Element type, or even to make the JSX Element more specific, so a casting is probably required here.

The closest I could come with googling was this issue: https://github.com/microsoft/TypeScript/issues/21699 which is still WIP.

vishnugopal: ^ cc @proxi Please let me know if you can find me a better alternative here. 🙏

src/components/Select.tsx
```diff
@@ -1,28 +1,31 @@
...
- */
+interface Props {
+    className?: string
+    disabled?: boolean
+    /** Currently selected value. */
+    value: string | number
+    /** Callback for the change event. Will be called with the next value (not the full event). */
+    onChange?: (value: Props['value']) => void
+    /** Options that are rendered in the select. */
+    options?: Option[]
```
proxi, line 25: Does it ever make sense to have Select without options? It seems to me that having optional Array is often bad style, and empty default array would be better.

I'm not sure about all consequences of this, but would `options = []` default work?

proxi: Ref.: https://twist.com/a/1585/ch/190200/t/1504732/c/58090586

src/components/Select.tsx
```diff
@@ -1,28 +1,31 @@
...
- */
+interface Option {
+    /** Optional key for each option. If not provided `value` is used. */
+    key?: string | number
+    /** Value of the option. */
+    value: string | number
+    /** Text to display for the option. */
+    text?: string | number
+    /** Whether the options is disabled or not. */
+    disabled?: boolean
```
proxi, line 14: Also ref.: https://twist.com/a/1585/ch/190200/t/1504732/c/58090586

(not sure if possible, but worth discussing a better solution)

src/components/Tabs.tsx
proxi: I think this is fine, there's no good solution.

src/components/Time.tsx
```diff
@@ -1,49 +1,49 @@
...
+    constructor(props: Props) {
         super(props)
         //eslint-disable-next-line @typescript-eslint/camelcase
         ;(this as any).refresh_interval = null
 
         this.state = {
             hovered: false,
-            mouseX: null,
-            mouseY: null,
+            mouseX: undefined,
```
proxi, line 45: @gnapse @henningmu Worth discussing as element of code style. Personally I think that `undefined` on object properties is confusing and it is better to use `null` in such cases. I would recomment having `mouseX: number | null` instead.

src/components/Tip.tsx
```diff
@@ -7,16 +7,15 @@ import classNames from 'classnames'
...
- * @property {boolean | undefined} [top]
- */
+interface Props {
+    className?: string
+    title?: React.ReactNode
+    message?: React.ReactNode
+    top?: boolean
+}
 
 /** @type {React.FC<Props>} */
```
proxi, line 17: After coonversion JSDoc should be removed.

src/components/Time.tsx
vishnugopal: Yes I agree that we should pick one and stick to it. I prefer `undefined` because it avoids an extra `| null` suffix in the type.

```
interface State {
    hovered: boolean
    mouseX: number | null
    mouseY: number | null
}
```

when there is great syntactic sugar available with the `?: number`. JS is kinda terrible in that it has 2 types of nulls 😶

src/components/Tooltip.tsx
```diff
@@ -101,26 +114,16 @@ class Tooltip extends React.Component<Props, State> {
         }, this.props.delayHide)
     }
 
-    /**
-     * @param {(...args: any[]) => void} actionFn
-     * @param {number} delay
-     */
-    _delayAction(actionFn, delay) {
+    _delayAction(actionFn: (...args: any[]) => void, delay: number) {
```
proxi, line 117:
```suggestion
    _delayAction(actionFn: Parameters<typeof setTimeout>[0], delay: number) {
```

src/components/utils/TimeUtils.ts
```diff
@@ -76,16 +68,13 @@ const TimeUtils = {
...
+    formatTimeLong(timestamp: number, config: Config = {}) {
+        const { locale = 'en', longFormat = this.LONG_FORMAT } = config
         const date = dayjs(timestamp * 1000)
         date.locale(locale)
         return date.format(longFormat)
     },
 }
 
 export default TimeUtils
+export { Config as TimeConfig }
```
proxi, line 80: Why not rename the type already?

src/components/utils/TimeUtils.ts
```diff
@@ -3,27 +3,23 @@ import LocalizedFormat from 'dayjs/plugin/localizedFormat'
...
- * @property {string} [locale]
- * @property {string} [longFormat]
- * @property {string} [shortFormatCurrentYear]
- * @property {string} [shortFormatPastYear]
- * @property {string} [daysSuffix]
- * @property {string} [hoursSuffix]
- * @property {string} [minutesSuffix]
- * @property {string} [momentsAgo]
- */
+type Config = {
```
proxi, line 6:
```suggestion
type TimeConfig = {
```

src/components/Time.tsx
proxi: The problem with undefined is that there's no distinction between type of `foo` in `{}` and `{foo: undefined}`. In a case like `State` above it loses all advantage of static typing (you could forget the property). My recommendation would be to never use undefined as explicit type (only use it to mark optionality of param).

src/components/Tip.tsx
vishnugopal: Done!

src/components/utils/TimeUtils.ts
vishnugopal: Done, that makes sense!

src/components/utils/TimeUtils.ts
vishnugopal: Done!

src/components/Tooltip.tsx
vishnugopal: Done!

src/components/Select.tsx
vishnugopal: Yeah, that makes sense. Makes the typing simpler to read and understand. Done!

src/components/Select.tsx
vishnugopal: Have moved everything to default values here, I'll try to do the same to other components. I'm not sure this is a good rule in every case, especially when string/number fields are involved (is ""/0 an acceptable default value?), so there is some nuance.

proxi: What I suggested is only true of empty collections and booleans. For other types it's only good if reasonable default can be found.

vishnugopal: This plays hell on tests as well as now we'll have to provide all the non-optional props. Semantically I feel this is wrong too as the component has this as an optional prop (as it has a default value set internally). So the callee Props interface does have this as optional.

Modelling this as non-optional is probably wrong, I've reverted this change in this PR.

proxi: Sad but true :/

src/components/Time.tsx
vishnugopal: I consider this a good thing 🙂 If you use undefined throughout then you've eliminated 2 classes of bugs here: the property not being present, and the property being undefined, and your type guards take care of the third one, the property being null. 

This is really stylistic preference though, but for what it's worth (& the big warnings), Typescript's code itself is [written to prefer undefined](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined).

src/components/Select.tsx
vishnugopal: Resolved as per here: https://twist.com/a/1585/ch/190200/t/1504732/c/58095833

vishnugopal:
Am going to merge this back today to the parent branch, and then open a new PR for the rest of the remaining changes. The only unresolved issue is null versus undefined above, where we have to take a team call on style. cc @henningmu

src/components/Time.tsx
dmiotti: I don‘t have a strong opinion to this, you both made great arguments. At first, I was somehow convinced by the use of `null`, like all the other languages does. But @vishnugopal provides good arguments for the use of `undefined` and it shifted my mind, so I would vote for `undefined`.

proxi: @vishnugopal Your arguments are solid, let's stay with undefined.

gnapse: Didn't get the chance to chime in, but Vishnu covered really well what I would've argued for as well. Even better, as I wouldn't have been able to express it so well.

gnapse:
@vishnugopal seems safe but I'll still ask: can we delete the branch?

PR #220, opened 2020-06-18
Expose event from KeyCapturer

frankieyan:
**Short description:**

The `event` object is now forwarded to the `onEnter` etc. handlers passed into `KeyCapturer`.

**PR Checklist:**

-   [x] Added tests for bugs / new features
-   [ ] Updated docs (storybooks, readme)
-   [x] Described changes in CHANGELOG.md
-   [ ] Executed `npm run check` and made sure no errors / warnings were shown
-   [x] Executed `npm run test` and made sure all tests are passing
-   [x] Bumped version in package.json
-   [x] Updated all static build artifacts (`npm run build-all`)

PR #223, opened 2020-06-26
KeyCapturer: Prevent onEnter from firing in the middle of an IME composition session

frankieyan:
**Short description:**

Replacing #220, this updates the `KeyCapturer` component to also prevent the `onEnter` handler from firing when an [IME composition session](https://developer.mozilla.org/en-US/docs/Glossary/input_method_editor) is happening.

**PR Checklist:**

-   [x] Described changes in CHANGELOG.md
-   ~~Executed `npm run check` and made sure no errors / warnings were shown~~
-   [x] Bumped version in package.json


src/components/KeyCapturer.tsx
```diff
@@ -90,33 +91,59 @@ type KeyCapturerProps = Record<
...
-    render() {
-        const { children, eventName = 'onKeyDown' } = this.props
-
-        return React.cloneElement(children as React.ReactElement, {
-            [eventName]: this._handleKeyEvent,
-        })
-    }
+    return React.cloneElement(children, {
+        [eventName]: handleKeyEvent,
+        ...composingEventHandlers,
```
stkao05, line 145: `composingEventHandlers` could be `undefined`?

gnapse: I think `undefined` can be safely spread into objects (but not into arrays, don't ask me why 🤷‍♂️).

image

frankieyan: Yes, it also seems to work for `null` as well as with `Object.assign`, but I can use a fallback empty object for `composingEventHandlers` 👍

stkao05: Ah okay I see. Just learned about it now 👍

https://github.com/tc39/ecma262/issues/687
https://stackoverflow.com/questions/46957194/javascript-es6-spread-operator-on-undefined


> Yes, it also seems to work for null as well as with Object.assign, but I can use a fallback empty object for composingEventHandlers 👍

No need to change the code. I think it is completely fine after knowing it is a defined behavior.

frankieyan: Sounds good 😊

I'll wait until it's fully tested on the Twist side before merging and cutting a release here 👍

PR #224, opened 2020-06-29
Documents release / publish workflow in readme

henningmu:
As discussed in our 1on1 here's a small doc on how to publish a new version of reactist 👍

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,3 +142,9 @@ npm run test-watch
 This also prints the current test coverage.
 
 MacOS users might need to upgrade watchman with `brew install watchman` when experiencing troubles with the watch mode. See this issue for details: https://github.com/facebook/jest/issues/1767
+
+# Releasing
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -142,3 +142,9 @@ npm run test-watch
 This also prints the current test coverage.
 
 MacOS users might need to upgrade watchman with `brew install watchman` when experiencing troubles with the watch mode. See this issue for details: https://github.com/facebook/jest/issues/1767
+
+# Releasing
+
+A new version of reactist is published both on npm and GPR whenever a new release on Github is created. A Github Action will automatically perform all the necessary steps.
```
frankieyan, line 148:
```suggestion
A new version of reactist is published both on npm and GitHub Package Registry whenever a new release on GitHub is created. A GitHub Action will automatically perform all the necessary steps.
```

I think the GPR acronym isn't used too often, at least I couldn't find many results when Googling, so it's probably better to be clear 👍

PR #226, opened 2020-07-02
Refactors the Button component's outer interface

gnapse:
## Description

A first refactor and redesign of the outer interface of the `Button` component. Here's a list of what's done and what's the plan:

1. Consolidated the `secondary`, and `danger` props into a single `variant` prop that can accept one of those names as a string value, plus `primary`. This prop is now required (more on why below) and the value when all flags were omitted is now `variant="primary"`.
1. The `white` prop now does not even exist as a `variant`. It is replaced by `variant="secondary"`.
1. Consolidated the `small` and `large` props into a single `size` prop that can accept one of those names as a string value. This prop is not required and defaults to `default`, which applies no special extra styling.
1. Renamed the `data_tip` prop as `tooltip`.
1. Consolidated a bit how both `disabled` and `loading` disable the button, now semantically in both cases (as opposed to programmatically ignore clicks when `loading={true}` but not disabling the button).
1. Replaced the use of `name` as the prop for providing the child elements of the button. It is now achieved using `children` as is most expected with buttons.
1. Changes all class names internally to follow a BEM-like approach. This is possibly temporary as it may as well be the case that with @vishnugopal's plans for theming we start using CSS modules, which will remove the need for this.
1. Removed the `LinkButton` component and replaced it with a new possible `variant` prop on the regular `Button`. So now `<LinkButton name="Click me" />` becomes `<Button variant="link">Click me</Button>`.

### What's the mid term plan

This is a first step in order to consolidate all our buttons. Regarding point (1) in the list above, the prop is made required because I want to introduce later a different semantics for when the prop is omitted. Omitting the prop will give us a bare-bones button with no special styles. But this first step does not yet enables that. It simply disallows the button to be used without this prop. So now if you want a primary-styled button you have to add `variant="primary"` to your buttons explicitly.

#### `variant="secondary"` vs. `variant="white"`

**Update:** this section discussion has been resolved. The `"white"` variant no longer exists. It should be replaced with `variant="secondary"`.

> One question for @henningmu: what's the deal around the fact that we have both a `white` and a `secondary` style modes (now called `variants`)? I wanted to deprecate `white` and consolidate them both under `secondary` but I see we use both in Twist. Buttons with them end up being very similar but not quite the same. Can we stick to one of these and drop the other? It'd be nice to define this now to streamline it all in this PR, and also take care of this in the upcoming PR in Twist to adopt these changes.

#### Button inner wrapper+span markup

Another question: I took the liberty to remove some inner markup we were generating inside the `button`. At first sight it seems to not have ay impact in how the buttons are rendered, but this is something I'll validate more tomorrow. But also wanted to highlight that to know what you think about this. Maybe there was a strong reason for doing this.

## PR Checklist


vishnugopal:
Hey @gnapse, I have worked on a PR to add [CSS module refactoring to this](https://github.com/Doist/reactist/pull/226). It's not a lot of changes, so if you'd like, this can go together, and especially since you'll be working on other components soon.

If you think that makes this PR too large, I'll refactor it to dev once you merge this in. Thanks!

gnapse:
@vishnugopal regarding Twist, I have a PR almost ready :) My plan is to try to keep it up with the changes in Reactist as they occur. This PR should prompt a new release. Maybe a beta release and we'll keep a beta branch. Though not yet sure about that.

Regarding the css modules, that's great. It's a pity that I did not coordinate with you before so we could avoid working on the same component independently at the same time. I'll check that out and see if I can and want to incorporate it here or else how we can go about releasing both things.

Let's talk on Monday please. I'll try to sneak something in your calendly.

gnapse:
Oh wait, I now realize that your PR is already on top of this one. I initially thought you worked on this independently and we'd have conflicts. Nice! Anyway, let's introduce this independently after this one. This because I'm not yet sure how we'll go about enabling client apps to customize the css when we introduce css modules.

henningmu:
> One question for @henningmu: what's the deal around the fact that we have both a white and a secondary style modes (now called variants)?

History 🕰

Totally okay with just keeping secondary and refactor all usages of `white` inside Twist to be secondary as well 👍
As you outlined with this new API it's much cleaner and also makes more sensed in a "themed world"

gnapse:
@vishnugopal about this

> I have one concern: how are we considering adopting this in Twist?

I opened the long-overdue PR in Twist: Doist/twist-web#2461. Been having some difficulty with tests failing that turn out it may not have anything to do with my chsnges but with tests not working when running against a locally linked version of Reactist.

gnapse:
Update: I realized we also have a `LinkButton`. I initially thought it was a component that rendered a link that looks like a button, but acts as a link (e.g. that it was for navigational purposes). But now I see it is actually a component that renders an anchor tag meant to work as a button semantically.

This means that I'll have to include the refactor of that in this PR. There will be no more `LinkButton` and we'll have a new `variant="link"` in the regular `Button` component. If it's meant to be a button, it has to be a `button` and not a `a`.

Will work on that to include it in this PR as well. Though is becoming bigger, but I do believe this is important. However, if we think this is better in separate PRs I am leaning towards creating a base branch for all this work and merging against it. If I do not push things here on time, let me know and we can do that tomorrow.

gnapse:
@frankieyan a question for you since you've been working lately in Reactist releases (@henningmu may also be able to help): what are your thoughts in how to release this. Is it possible to release beta versions of Reactist? Is it simply updating the version in `package.json` to have the beta version notation? And also, if we go as a beta release, shouldn't it be in a `beta` branch as well instead of the `dev` branch? Or maybe we decide not going beta at all, and simply make this the next major release. Thoughts? (I will dig into this tomorrow but wanted to leave the question anyway due to the time difference).

frankieyan:
I'd vote for cutting a major release and pulling it into Todoist and Twist immediately after rather than have a long-running beta branch 👍

To update the version numbers you can use `npm --no-git-tag-version version major` to update the lock file (or manually works too) as well.

henningmu:
If you're uncomfortable with releasing a lot of majors during your DO I'm very much okay with doing beta releases for a month and publishing the next major at the end of it.

We've never tried to publish a beta version but I'd assume it works by changing the package.json version field to `5.0.0-beta.0` and then creating the release on Github. In Twist we should then be able to pin to the respective beta version 👍

gnapse:
I think I'll go with Frankie's suggestions and make this a major release anyway. Overall I do not think there will be too many major releases due to my work on this DO, maybe 2 or 3.

@henningmu any comment on my plans for the link button?

proxi:
@gnapse Can you please make sure https://github.com/Doist/reactist/pull/249 and https://github.com/Doist/reactist/pull/250 make it into next release? The first one improves and unifies lint and the second one fixes several typing problems when used in Twist.

gnapse:
@henningmu @vishnugopal I followed through with the removal of the `LinkButton`. It was not difficult, including the changes on the Twist side, and I figured I might as well include it in this breaking change release that's all about the button.

I updated the PR description accordingly

> Removed the `LinkButton` component and replaced it with a new possible `variant` prop on the regular `Button`. So now `<LinkButton name="Click me" />` becomes `<Button variant="link">Click me</Button>`.

I will now proceed to update the CHANGELOG, increase the version to a new major one, and generate the build files and this will be fully ready for review and we can merge if you all approve.

gnapse:
@Doist/frontend-product given the scale of this change, and the fact that there's been some changes since the single approval I've got, I'd rather have another pair of eyes approve this once again before merging.

src/components/styles/button.less
```diff
@@ -22,64 +20,69 @@
...
-    &.white {
-        background-color: white;
-        .default_border(@color: @primary_color);
+    &--link {
         color: @primary_color;
+        background-color: transparent;
+        border: none;
+        text-decoration: none;
+        padding: 0;
+        font-size: inherit;
```
henningmu, line 65: This doesn't seem to work – at least in the storybook examples this overwrites the font-size set by the variant (e.g. `small`).

gnapse: Good catch. Fixing in a new commit. Will merge after it builds in CI and will make a new release later today.

gnapse:
> Can you please make sure #249 and #250 make it into next release? The first one improves and unifies lint and the second one fixes several typing problems when used in Twist.

@proxi I intentionally did not include them in the release for this PR because it was a big enough change already, big changelog entry, major release, and also those two PRs were not approved. But I also did so because I have a new release upcoming later today or by tomorrow with the new `KeyboardShortcut` component, and I'll include those PRs in it. Will also take a look and approve them myself to make sure they're ready.

PR #230, opened 2020-07-03
Fix publish action

frankieyan:
## Short description

Another attempt to fix our publish action. Basically, the problem is that we need to authenticate with npm again after authenticating with the GitHub registry in order to publish there. Not sure if the `setup-node` action inherits previous settings but its [demos](https://github.com/actions/setup-node#usage) seem to indicate so, so this is an attempt by resetting the scope and hoping an `.npmrc` will be properly created from it.

## PR Checklist

-   [ ] Executed `npm run lint` and made sure no errors / warnings were shown
-   [ ] Executed `npm run test` and made sure all tests are passing

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,9 +14,13 @@ jobs:
           node-version: 12
           registry-url: https://npm.pkg.github.com
           scope: '@doist'
+      # Skip post-install scripts here, as a malicious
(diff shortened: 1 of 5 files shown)


vishnugopal:
If this doesn't work, maybe you can try publishing to npmjs first, and Github second? I can't think of anything else 🤷🏽‍♂️

frankieyan:
Going to give this a shot now 🤞

PR #231, opened 2020-07-03
Fix and enhance KeyCapturer type definitions

gnapse:
**Link to issue:** related to #223 

Also, the motivation comes due to an upcoming PR to Twist to update it for #226, and in doing so I have troubles with a use of `KeyCapturer` in Twist because the type of the event handlers do not allow for passing a function that receives an event.

## Short description

I noticed that the type definitions for `KeyCapturer` had an incorrect type for the handlers, and overall a very relaxed typing on some dynamic props. The props for the handlers and the props for the `propagateXYZ` flags were all bundled inside a very liberal type that was not even allowing to call the handlers with a React event as an argument (check [here](https://github.com/Doist/reactist/blob/f3e33c2513e1b2fc1330802e40117b1f657cd1bf/src/components/KeyCapturer.tsx#L78) how it is all defined as `Record<string, (() => void) | boolean | React.ReactChild>`).

In the process of fixing this last bit, I also took the opportunity to remove the dynamic typing of these crucial props (`onXYZ`, `propagateXYZ`). It is all a bit more verbose now but only on the TypeScript side. On the pure JS side I think there's even a bit less code.

### Questions around the propagateXYZ props

I wanted to propose to deprecate them. Can't this be something the event handlers take care of themselves?

Also, there's a bit of a discrepancy with how we offer this API. Currently there's nothing impeding us to use it in this way:

```jsx
<KeyCapturer
  onEnter={() => { /* ... */ }}
  propagateArrowDown
/>
```

That is, we can pass an event handler for one key type but a propagate flag for another that does not even has a handler. Overall I think this API is a bit unintuitive and redundant on what an event handler is already able to do.

## PR Checklist

Still pending to take care of some of these stuff below.

-   [x] Added tests for bugs / new features (technically only fixed tests after changes, as there are no new features)
-   [ ] Updated docs (storybooks, readme) (N/A?)


henningmu:
> I wanted to propose to deprecate them. Can't this be something the event handlers take care of themselves?

I introduced this mainly due to laziness on the consuming side. Most users of the keycapturer would need to manually prevent default / stop propagation (likely due to bad design in general) so I added it as a default.
I'm okay with forcing this more explicitly onto the consumers though  👍

gnapse:
Ok, I will probably remove them in a subsequent PR if not in this one. Can I get a review here and now if this is mergeable as is? @henningmu @frankieyan @vishnugopal

PR #249, opened 2020-07-08
Proxi/typescript eslint plugin

proxi:
## Short description

Update `typescript-eslint` plugin to 3.x.

## PR Checklist

It does not change runtime behaviour.


src/components/Button.tsx
```diff
@@ -54,7 +54,6 @@ class Button extends React.Component<
             danger,
             disabled,
             name,
-            //eslint-disable-next-line @typescript-eslint/camelcase
```
gnapse, line 57: Why is this rule not needed anymore?

proxi: This rule does not exist anymore: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/camelcase.md

PR #250, opened 2020-07-08
More accurate type of onChange callback.

proxi:
## Short description

`Select.onChange` callback is currently typed with `string | number | undefined` as return value, but value type will always be `string`.

## PR Checklist

It only changes types. The new type narrower, so the change is non-breaking.

### DIFF
--- a/src/components/Select.tsx
+++ b/src/components/Select.tsx
@@ -20,7 +20,7 @@ type Props = {
     /** Currently selected value. */
     value?: string | number
     /** Callback for the change event. Will be called with the next value (not the full event). */
-    onChange?: (value: Props['value']) => void
+    onChange?: (value: string) => void
     /** Options that are rendered in the select. */
     options?: Option[]
 }


src/components/Select.tsx
```diff
@@ -20,7 +20,7 @@ type Props = {
     /** Currently selected value. */
     value?: string | number
     /** Callback for the change event. Will be called with the next value (not the full event). */
-    onChange?: (value: Props['value']) => void
+    onChange?: (value: string) => void
```
gnapse, line 23: I understand the part about not using `Props['value']` directly because it can be `undefined`. But shouldn't this here be `string | number`? I say it based on the fact that the `value` can be `string` or `number`.

proxi: This is exactly the issue here: as far as I understand, returned value will always be a `string`. There's no way to return a number, even if number was passed in, without explicitly converting back.

Having a type-aware select would be very nice, but not sure how hard to implement.

gnapse: I'm still not entirely satisfied with having a component that states that the value can be a number or a string, but then the even handler called to change that value only accepts string. Maybe it's that the `value` needs not be `number | string` and can be typed `string` only?

I can dig in later, but I'll probably won't include this in the next release yet. There will be more releases soon.

proxi: > I'm still not entirely satisfied with having a component that states that the value can be a number or a string, but then the even handler called to change that value only accepts string

In my view callback doesn't "accept" string, it "returns" it. Behaviour is exactly the same as HTML input element:

Zrzut ekranu 2020-07-15 o 19 06 16

IMO best API would be to make Select generic: `Select<string>` but it's more work on both sides.

gnapse: I'm still not super convinced overall. The thing with the `input` in the screenshot above is due to a type coercion. But we certainly don't want our components to rely on this.

But anyway, I'll merge this and include it in the next release. I'm overall not super happy about this `Select` component, that probably needs to be revisited beyond this.

proxi: I see what you mean, but it is exact same type coercion that works on callback.

IMO ideally Select would be a generic component like `Select<OptionType>` (is it even possible)? The implementation would have to be more complex, since options are coerced to string a separate map would be needed. Having a component like this would greatly simplify client code in many places, you could put objects (e.g. Workspace) in the list directly instead of workspace ids.

gnapse: It would be possible, but I also think it would overcomplicate things. In these cases I tend to either accept `number | string` or just a `string` and those who want numbers have to deal with it. That's how the native select works anyway.

The bad thing about accepting an entire `OptionType` object is that it introduces ambiguity. Imagine you are showing the options `[{ id: '1', label: 'First' }, { id : '2', label: 'Second' }]` but then your `onChange` callback accepts an entire full-fledged option instead of just the id is that there's nothing preventing it to happen that the callback receives `{ id: '1', label: 'Different label' }`. That's why I'd rather stick for the `onChange` to just handle an ID.

gnapse: BTW we also have a `Dropdown.Select` component in Todoist, now also copied over to Twist (for some reason in the context of the new composer Pedro did not use the Reactist `Select` and preferred to copy that one). I have to dig deeper into both components to unify them in a single one in Reactist for both apps to use.

I'll take the opportunity to mention here @pedroalves0 and @henningmu who should have more upfront knowledge on the differences and possible shortcomings of the Reactist `Select`. This is yet another thing we need to unify as we're doing with the button, tooltip, etc.

proxi: @gnapse I don't understand your explanation, what I meant is that Select would take a list of model entities (e.g Workspaces) and `onChange` would just return entire entity (`onChange(value: Workspace) => void`). It seems to be the most natural API for a select, doesn't it?

gnapse: What I meant is that passing the entire entity is not needed, because what you usually need and keep as "what the selected value is" is the id of the entity selected.

Also, with such an approach, how would this hypothetical `Select` component know how to interpret the attributes of that generic type to be able to know which attribute acts as `id` and which attribute acts as some sort of visual name or label identifying the object? The latter part (the label) is less of a problem because such a component will usually defer the rendering of each option in the list to the user of the component. But the id is key. I can see ways of achieving this, but again, it would overcomplicate things.

I'd rather do something where the entire component is totally agnostic of the data structure, you just give it a list of options and what's the id of each one. Something like what this `Dropdown.Select` from todoist-web already does:

https://github.com/Doist/todoist-web/blob/88ecaf40eeb6801d94e92daed66d05e2db849607/react_components/color_dropdown_select/ColorDropdownSelect.js#L42-L57

proxi: IMO there's big downside to the API you described in strictly-typed languages: if you only get back `id`, you have to make a lookup `universe[id]` and you get back an optional. Every time you use Select you either have a forced cast to get rid of the null, or an unnecessary check.

> how would this hypothetical Select component know how to interpret the attributes of that generic type to be able to know which attribute acts as id

Either by convention (`OptionType extends RequiredOptionType, RequiredOptionType implements { id, name}`) or by callback or field accessor (`Select<ModelType, IdField, NameField>`).

gnapse: > IMO there's big downside to the API you described in strictly-typed languages: if you only get back id, you have to make a lookup universe[id] and you get back an optional. Every time you use Select you either have a forced cast to get rid of the null, or an unnecessary check.

Now it is I who do not understand your explanation. Can you point out in the linked code from Todoist where is that this downsides show up?

gnapse: To further back my claims I'd point to a couple implementations of equivalent or very similar components in popular and modern component libraries. At a quick glance they all are driven by the option IDs and abstract themselves away from having to know the data structure of the objects being given as options:

- [material-ui](https://material-ui.com/components/selects/).
- reach-ui's [combo box](https://reach.tech/combobox) and [list box](https://reach.tech/listbox).

PR #237, opened 2020-07-06
Refactor build system to support CSS modules

vishnugopal:
## Short description

This builds infrastructure for a single rollup build that supports:

* CSS modules
* Code splitting just [like how antd does it](https://unpkg.com/browse/antd@4.3.0/).
* Conversion of SVG to inline data urls.

This also removes the previous webpack builder as that is no longer necessary with this workflow.

## Test Plan
This adds infrastructure to support CSS modules. Nothing has yet been made CSS module yet, as that needs additional changes in Twist & Todoist, but the builder has been completely changed. To test:

1. Checkout reactist
2. npm run build
3. Checkout twist-web
4. npm i; npm run reactist:link; npm run start
5. Make sure everything works fine.
6. Do same steps for todoist-web

Note there are 2 PRs, [one Twist](https://github.com/Doist/twist-web/pull/2497), and [one Todoist](https://github.com/Doist/todoist-web/pull/1678) that have to be reviewed with this together.
 
## PR Checklist
-   <strike>[ ] Added tests for bugs / new features</strike>
-   [x] Updated docs (storybooks, readme)
-   [x] Described changes in `CHANGELOG.md`
-   [x] Executed `npm run lint` and made sure no errors / warnings were shown
-   [x] Executed `npm run test` and made sure all tests are passing
-   [x] Bumped version in `package.json`
-   [x] Updated all static build artifacts (`npm run build-all`)

PR #274, opened 2020-07-21
Convert Button to use CSS modules

vishnugopal:
This builds on top of [this reactist PR](https://github.com/Doist/reactist/pull/237) to convert Button to use CSS modules

## To Test

In `twist-web` do `npm run reactist:link` on this branch, test everything is working ok.

## Pending Work

- [x] Merge https://github.com/Doist/reactist/pull/237
- [ ] <strike>Bring in styles from Twist into Reactist</strike> Decided to defer this to later in a theming PR as that's the only way to bring in colours etc. into Reactist as CSS variables. We also might have to always have a `reactist_button` class because of [changes like this](https://github.com/Doist/twist-web/pull/2461/files#diff-2a77a1500ab87d4918863f5f4d065bafR337).

## PR Checklist

## Versioning

Bumped up to v6.1.0


src/components/Button/index.tsx
```diff
@@ -54,10 +54,10 @@ const Button = React.forwardRef<HTMLButtonElement, Props>(function Button(
     ref,
 ) {
     const className = classNames(
-        'reactist_button',
-        variant ? `reactist_button--${variant}` : null,
-        size !== 'default' ? `reactist_button--${size}` : null,
-        { 'reactist_button--loading': loading },
+        styles['reactist_button'],
```
gnapse, line 57: With a css modules approach in place I'd rename all this classes to get away from a BEM-style naming scheme to one more unconcerned with name clashes, since css modules ensure these are local and not global.

- `.reactist_button` can become `.button` or `.container`
- `.reactist_button--xyz` can become `.xyz` for any "xyz" that we may have.
- camelCase should be used for any class name that needs more than one word in it. The js code accessing styles should not need the square-brackets notation to access any of the attributes of `styles`.

vishnugopal: Can do this, but moving out of `reactist_button` classes might be harder than I thought coz of [styles like this](https://github.com/Doist/twist-web/pull/2461/files#diff-2a77a1500ab87d4918863f5f4d065bafR337). 

Anyways can only land this after the theming PR drops as that's the only way to move such styles to within reactist.

src/components/Button/index.tsx
```diff
@@ -55,9 +55,11 @@ const Button = React.forwardRef<HTMLButtonElement, Props>(function Button(
 ) {
     const className = classNames(
         'reactist_button',
```
vishnugopal, line 57: Note that we add `reactist_button` and `reactist_button--primary` classes etc. also in addition to CSS module classes to support Twist backwards-compatibility for now.

tsdx.config.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
         // Also inline all assets in CSS files using base64 encoding & data URLs.
         config.plugins = [
             styles({
-                autoModules: /index/,
+                autoModules: true,
```
vishnugopal, line 9: Use the standard `.module.less` prefix instead as that's more widely known.

src/components/Button/index.tsx
gnapse: But I insist: the css module classes should not (need not) be named in the same way the global classes were named.

henningmu: I guess it's okay for now but shortly after shipping themes (or by shipping them?) we should bump a major and remove this 👍

src/components/__tests__/Button.test.tsx
```diff
@@ -53,7 +53,7 @@ describe('Button', () => {
               class="reactist_popover__wrapper reactist_tooltip__wrapper"
             >
               <button
-                class="reactist_button reactist_button--primary"
+                class="reactist_button reactist_button reactist_button--primary reactist_button--primary"
```
henningmu, line 56: What's happening here? Why do we add all classes twice?

gnapse: I think it is because we still add `reactist_button` for backward compatibility, but we also add `styles['reactist_button']` from the css module. It so happens that in tests css modules are most probably being mocked and it just returns the same name as the class name, instead of a dynamically generated one.

This is not the reason why asked for it, but it also surfaces the fact that there's no need to call the css module class names the way we did, as they're now local to the button. For instance, `reactist_button` can become `container` and `reactist_button--primary` can become `primary`.

henningmu: Sounds good, please do change the new class names to something else @vishnugopal 👍

vishnugopal:
Closing in favor of a more comprehensive theming solution.

PR #277, opened 2020-07-28
Typed Theming Experiment

vishnugopal:
This is an experiment to introduce typed theming into Reactist, and (eventually) port over all Twist & Todoist themes to Reactist.

Intended Development Workflow:

1. Work with the design team to define variables that change between themes. When possible, make these variables generically named, and similar to what the design team would call it. example: `primaryFillColor` instead of `markdownComponentFillColor`
2. Introduce these properly typed to the `Theme` interface in src/themes.
3. Make sure every theme in src/themes has concrete definitions for this property. This is easy to find now as the editor will complain if we miss out on a property definition. [This is one of the advantages of defining themes like this, and it also makes future refactoring easier]
4. Run `npm run storybook`. It should also spin up a babel instance to compile themes into a generated themes.less. 
5. Play around with the button component to change styles.

Screenshot 2020-07-28 at 10 47 34 AM

Tradeoffs that I've noticed:
1. Because theme switching is now via CSS variables and not LESS, we can't use things like the less `darken` method. All colours will have to explicitly defined. This imho is a good thing, but it's definitely less convenient.
2. This depends on CSS modules being defined, so first we'll have to port components over to that.
3. Naming requires some consensus with design. [antd for e.g. just gives up and names with numbers](https://ant.design/docs/react/customize-theme), but I feel we should do better. 

Please provide feedback for:
1. Development workflow: it's a bit slow still, but I can't think of easy ways to make this faster. It takes around 2-3s after a save to reflect in storybook.
2. Anything else!

Things that have not been implemented here:
1. `ThemeProvider`, but this is a standard component that'll just add a class to body (similar to how storybook works now).
2. A complete Button refactor out of Twist. This is something I'll do after I get more feedback to see if I'm going in the right direction!


src/themes/Theme.ts
```diff
@@ -0,0 +1,14 @@
+interface Theme {
+    primaryFillColor: string
```
henningmu, line 2: The variable names in Twist-Web can be considered the "gold standard" when it comes to naming and they're aligned with Design. We should use the same ones here (e.g. `--content-primary`, `--chromatic-fill-blue`, ...)

tsdx.config.js
```diff
@@ -6,9 +6,12 @@ module.exports = {
         // Also inline all assets in CSS files using base64 encoding & data URLs.
         config.plugins = [
             styles({
-                autoModules: /index/,
+                autoModules: true,
                 mode: 'extract',
                 url: { inline: true },
+                less: {
+                    globalVars: { primary_color2: '#3f82ef' },
+                },
```
henningmu, line 14: For what do we need this?

src/components/Button/index.tsx
```diff
@@ -40,16 +44,20 @@ const Button = React.forwardRef<HTMLButtonElement, Props>(function Button(
         disabled = false,
         tooltip,
         onClick,
+        theme,
```
gnapse, line 47: Possibly dumb question given that I just started to look at this diff and may lack context: Shouldn't the theme be global and maybe provided via a context? It does not seem scalable to be having to pass it to every component.

.storybook/config.js
```diff
@@ -1,6 +1,14 @@
 import { configure, setAddon } from '@storybook/react'
 import { setOptions } from '@storybook/addon-options'
 import chaptersAddon from 'react-storybook-addon-chapters'
+import { addParameters } from '@storybook/react'
+
+addParameters({
```
stkao05, line 6: Could we use [@storybook/addon-knobs](https://www.npmjs.com/package/@storybook/addon-knobs) for configuration for consistency with the existing setup?

src/components/Button/index.tsx
henningmu: I think the idea here is: we read the `theme` from a context but can override on a per component-instance basis by providing a different theme in props

vishnugopal: Yup that's it! Even though we have a global theme via a ThemeProvider (not yet in this PR), we might need a selective override for a single component.

tsdx.config.js
vishnugopal: This is a relic of a previous implementation, will remove!

.storybook/config.js
vishnugopal: Knobs are component specific. What we have here is a global theme definition, that's why the switcher is on top if that makes sense.

vishnugopal:
@stkao05 Having it in Typescript does have few advantages:

1. It's typed, so you'll never miss a theme configuration for a theme. Granted, right now it's not a big advantage, but going forward if we have multiple properties, we can change this structure as we see fit.
2. It provides sharing the theme variables between CSS and JS in the future in case we need it. In a recent Todoist PR of mine for e.g. you mentioned to share some variables (menu resize widths) to avoid duplication. We get that for free here.
3. We can (in the future if needed) port this over to a CSS-in-JS solution pretty easily.

There are some tradeoffs as well, as in:

1. You cannot use LESS macros (like darken) because those depends on LESS variables, and we use CSS variables here. But this'll be true of any solution that uses CSS variables.
2. The developer experience is slower.

If you feel strongly about this, I can refactor to make the Typescript bits go away.

> How to the user of Reactist custom the theme setting here? Would be great if you can demonstrate it via apply it to Todoist.

It works by implementing a ThemeProvider. I'll start a Twist PR tomorrow to demo this.

vishnugopal:
Hi @gnapse @henningmu @stkao05 I have implemented a ThemeProvider, and an accompanying [Twist PR here](https://github.com/Doist/twist-web/pull/2562).

I have also:
1. Changed all theme variable names to conform to Twist.
2. Removed the `reactist-button--primary/secondary` custom class as that's now part of the theme.

src/components/Button/Button.module.less
```diff
@@ -34,11 +34,11 @@
     }
 
     &--primary {
-        background-color: @primary_color;
-        color: white;
+        background-color: var(--reactist-button-primary-fill);
+        color: var(--reactist-non-adaptive-white);
```
henningmu, line 38: Seeing this: let's go off of Twist's variables and introduce `--reactist-button-primary-content` (maybe we don't even need the `reactist-` prefix everywhere?

Sorry for being so flaky on my reviews 😞

stkao05:
@vishnugopal 
I am still not sure if these benefits are worth the extra layer of indirection we are adding here. It isn't really about the slowness but rather an extra layer of architecture complexity. 

It may be hard to say they are useful or not but I would suggest we go with the simpler approach now and adopt a more complex approach if needed.

src/themes/ThemeProvider.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * Usage:
+ *
+ *  <ThemeProvider theme="twist-light" />
+ *
+ * somewhere in your React component tree.
+ *
+ * Note: Only one instance of ThemeProvider must be rendered. To override themes for a single component,
+ * use the `theme` prop for that component.
+ */
+export default function ThemeProvider({ theme }: ThemeProviderProps) {
```
henningmu, line 21: Probably this is enough for the initial use case but looking both at Todoist and Twist I think we want to turn this into a real context provider that exposes the theme and a function to change the theme. Should we add this immediately or keep it simple for as long as possible?

henningmu: Looking at the Twist PR I think we should keep it simple :D

src/components/Button/Button.module.less
vishnugopal: Hi @henningmu without the reactist prefix, these CSS variables collide with the twist ones, and that leads to general confusion in terms of which variables effect the change, so I'm keeping that in.

Change reg: `--reactist-button-primary-content` done!

src/components/Button/Button.module.less
```diff
@@ -34,30 +34,30 @@
     }
 
     &--primary {
-        background-color: @primary_color;
-        color: white;
+        background-color: var(--reactist-button-primary-fill);
+        color: var(--reactist-button-content-primary);
```
gnapse, line 38: Although I'm a fan, switching to CSS custom properties will have the consequence of Todoist not being able to adopt this, or having to style them fully instead of using the theme feature. This because Todoist still supports IE11 which does not have support for CSS custom properties.

src/components/Button/index.tsx
```diff
@@ -1,36 +1,45 @@
...
-     */
-    size?: 'default' | 'small' | 'large'
-    /**
-     * Tooltip that is displayed on hover.
-     *
-     * This replaces `title` which is not supported for these buttons to avoid confusion.
-     */
-    tooltip?: React.ReactNode
+type ThemeProps = {
+    theme?: 'twistLight' | 'twistDark'
```
gnapse, line 17: I know this PR is an experiment, so this is probably temporary. But it strikes as odd that there's a theme definition inside the button module. A theme should be something more global. Also the names having "twist" seems odd.

src/components/Button/index.tsx
```diff
@@ -1,36 +1,45 @@
...
 }
 
+export type ButtonProps = Omit<NativeButtonProps, 'title' | 'ref'> &
+    ThemeProps & {
```
gnapse, line 21: I raised this before and I'll raise it again. I am not a fan of us being able to override the theme on a per component basis.

I would expect the components to read the theme inside them.

src/themes/themes.less
```diff
@@ -0,0 +1,26 @@
+//This is an auto-generated themes definition file. Do not edit manually!
+//To make changes during development, run `npm run start` or
+//`npm run storybook`
```
gnapse, line 3: I'm not a fan of having auto-generated files be part of git. Not saying it should be removed, but wanted your opinion on this.

src/utils/underscore.ts
```diff
@@ -0,0 +1,12 @@
+export default function underscore(camelCasedWord: string) {
```
gnapse, line 1: I wouldn't mind having lodash as a dependency as long as we make sure we only import the couple of functions we need from it.

src/components/Button/index.tsx
vishnugopal: I've found that it is useful to override themes in certain scenarios. One example that comes to mind is in case of a lightbox component that behaves as if it's in a dark theme even inside light-theme contexts. Otherwise (as in the case of twist right now) we'll need to add lightbox-specific variables to the theme. 

I borrowed this pattern from [styled-components](https://styled-components.com/docs/advanced#the-theme-prop). This provides us additional flexibility, and doesn't take anything away. ThemeProvider is still available, but because of how it works, it cannot be nested within each other.

src/themes/themes.less
vishnugopal: I've seen this argued both ways, with some folks even committing node_modules to the repo. In this context though, I can remove it from the repo pretty easily. Will do!

src/components/Button/Button.module.less
vishnugopal: Hmm, I'm not sure what to do here. Without CSS variables, styling becomes much harder! Do we have a timeline on when we can retire IE 11 support? cc @henningmu

src/utils/underscore.ts
vishnugopal: We seem to be [moving away from lodash](https://twist.com/a/1585/msg/756020/m/30577974), do we want to do this?

src/components/Button/index.tsx
vishnugopal: After discussion with Steven and Henning, we might need to switch to a CSS class-based theme switcher, and where the styles LESS file has duplicate definitions for each CSS theme-class. This means that we'll have extra "un-needed" (for themes that are not the current theme) CSS loaded on every page. But I suppose that's the only way to make this work.

cc @gnapse @henningmu @stkao05

src/components/Button/index.tsx
frankieyan: Adding to this, if we use a real context provider we can use additional nested providers rather than using a prop. Example: https://evergreen.segment.com/get-started/theming

PR #278, opened 2020-07-28
Remove old MenuButton, replaced with a new set of Menu components

gnapse:
## Short description

This introduces a new richer and accessibility-compliant dropdown menu.

It has support for

- sub-menus
- disabled menu items
- optionally preventing closing the menu on certain item activation
- adding titled groups of menu items
- menu separators (simple `<hr />` elements)
- full accessibility support thanks to reakit (arrow keys nav, semantic elements recognized as menu and menu-items by screen readers, type-ahead some characters of an item's label to quickly move the focus to it).
- individual elements are fully customizable. They can have applied their own styles, class names, or any supported DOM attribute. They can even be changed into being other kind of element (e.g. allowing the `MenuButton` be a reactist `Button` and receive its props, or a menu item can be a link, thus having a `href` and triggering navigation).

~~The menu items have a stricter format where you provide a label, icon (optional) and keyboard shortcut (optional) instead of free-form `children`. Though this is something I may change because after the fact I realize it may be too restrictive. I'll probably revert the menu items to allow more freedom in their content. The main Twist app menu is an example of the need for this.~~ **Update:** menu items are now un-opinionated about their content. Much better that way.

I originally planned to not remove the existing `MenuButton`, merely deprecate it, and make this a minor release. But the name clash was what tipped it towards making this a replacement (can't export two different things named `MenuButton`), which of course means this is a breaking change.

There's a PR in the works that makes Twist adopt this new menu, replacing all the existing ones.

## PR Checklist

## Versioning

Should become part of a new major release (I'm exploring the possibility to make a single new major release for both this and #276 depending on the ability to synchronize having both merged around the same time).

## Demo

CleanShot 2020-07-28 at 19 44 26

## For the future

I have validated the possibility to implement some features in this menu that are needed in Todoist but not in Twist (e.g. using it as a context menu; allowing to control the open/closed state from the outside). I had some difficulties that should be solvable but I was already devoting too much time and the primary use case here is to be able to use this in Twist, where those features are not needed.

It would also be a nice to have to be able to have more semantic menu items such as checkbox menu item and a group of radio-button menu items. See reakit which has support for all this. We could adopt this for menu items such as Twist's "toggle dark/light theme" option in the main app menu.


src/components/Menu.tsx
```diff
@@ -0,0 +1,351 @@
...
+                        <svg viewBox="0 0 43.3 50">
+                            <polygon points="43.3 25 0 0 0 50 43.3 25"></polygon>
+                        </svg>
+                    </MenuButton>
+                    {children}
+                </Menu>
+            )
+        }),
+        [children, label, icon, handleSubItemSelect],
+    )
```
diegohaz, line 300: > it does not allow us to own the rendering of the menu items that open a sub-menu

I'm not sure if I understand what you mean here. But you can also use render props so as to avoid creating a new component in the render phase:

```jsx
<Reakit.MenuItem {...state} {...props} ref={ref}>
  {(itemHTMLProps) => ( // itemHTMLProps already contains ref
    <Menu>
      <MenuButton {...itemHTMLProps}>...</MenuButton>
      {children}
    </Menu>
  )}
</Reakit.MenuItem>
```

Something I think may cause bugs in your code is that `children`, `label`, `icon` and `handleSubItemSelect` should be memoized. Otherwise it's going to mount a new `SubMenuList` component on every render, losing internal state.

For example, in the example usage, `children` will be always different since `React.createElement()` always returns a different object.

```jsx
<SubMenu label="More options">
  <MenuList>
    <MenuItem label="Preferences" />
    <MenuItem label="Sign out" />
  </MenuList>
</SubMenu>
```

diegohaz: Alternatively, you can also create `SubMenuList` in the module scope as usual and pass props using `Reakit.Menuitem`.

```jsx
<Reakit.MenuItem
  {...props}
  {...state}
  as={SubMenuList}
  ref={ref}
  label={label}
  icon={icon}
  handleSubItemSelect={handleSubItemSelect}
>
  {children}
</Reakit.MenuItem>
```

Those additional props will be passed to `SubMenuList` as they aren't consumed by `Reakit.MenuItem`.

diegohaz: Another simpler option that may work:
```jsx
// inside SubMenu component
return (
  <Menu onItemSelect={handleSubItemSelect}>
    <Reakit.MenuItem
      {...state}
      {...props}
      ref={ref}
      as={MenuButton}
      label={label}
      icon={icon}
    />
    {children}
  </Menu>
);
```

gnapse: Yup, this was already not something I was convinced with. Additionally I may ditch the restrictive way of providing content to menu items and let it be a free form children prop.

Anyway, thanks a lot for your insights 🎉. I’ll take a look today and combined with the above it will probably not be needed.

gnapse: I went with the first option above, but alongside allowing the user to provide the content of the menu button (because I also had decided yesterday after pushing this that it would be best to leave it free and let users of the lib to add the icon or any other structure inside it themselves).

src/components/styles/menu.less
```diff
@@ -0,0 +1,81 @@
+@import './constants.less';
+
+.reactist_menulist[role='menu'] {
+    display: block;
+    white-space: nowrap;
+    background: hsla(0, 100%, 100%, 0.99);
+    outline: none;
+    font-size: 85%;
```
henningmu, line 8: Shouldn't we rely on our font size constants (e.g. `@smaller_font_size`) here and let users override if necessary?

PR #327, opened 2020-09-17
Removes max-width set for tooltip content

craigcarlyle:
## Short description

We should let the `max-width` of the Tooltip's content be set by the implementors rather than setting a blanket rule. As-is, it prevents us from using the new Tooltip component in Todoist because the content is cut off:

<table>
<thead>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
</thead>
<tbody>
  <tr>
<td>
Screen Shot 2020-09-17 at 9 01 04 AM
</td>
    <td>
Screen Shot 2020-09-17 at 9 02 00 AM
</td>
  </tr>
</tbody>
</table>

## PR Checklist

## Versioning

A new version will have to be created after this is merged to bring into Todoist.


gnapse (approved):
Looks good.

A couple of notes thinking out loud:

- we should look into how this may affect tooltips in Twist when this new upcoming version is out. Just saying, not something you need to act on right now (it doesn't even have to be you doing it).
- I wonder if you can hold up on merging and releasing this until you make sure no new changes to the tooltip are needed (in order to batch all the needed changes in the same Reactist release). I am assuming this change stems from you already having the Reactist tooltip being used in Todoist locally in your devenv, so you may know how far along we are on this.

craigcarlyle:
@gnapse, agreed to both. I'll hold off on merging until I'm done my initial implementation.

PR #329, opened 2020-09-18
Adds z-index for Tooltip

craigcarlyle:
## Short description

We currently use the Tooltip in Todoist in our Popper, which has a `z-index` of `1000`. We need to increase it from `2` so that it's actually visible.

## PR Checklist

## Versioning

This should be included with https://github.com/Doist/reactist/pull/327 in the next release.


src/components/Tooltip.tsx
```diff
@@ -62,6 +62,7 @@ function Tooltip({
                     {...tooltip}
                     {...props}
                     className={classNames('reactist_tooltip', className)}
+                    style={{ zIndex: 100000 }}
```
henningmu, line 65: Can / should we make this configurable? Feels strange to ship something with z-index 100000 just because we need in Todoist 😬

gnapse: I'd even go further and say that we do not need to change reactist for this, not even for making it configurable. It should be configurable already. Can't we apply styles to the `.reactist_tooltip` selector from inside a todoist stylesheet?

craigcarlyle: Oops, this was left over from my investigation, this should not be here 😅

craigcarlyle: @gnapse I think we should try to move towards having a consistent `z-index` strategy in Reactist. My worry is that if we don't have a centralized z-index order it tends to get very messy due to certain implementations. How strongly do you feel about keeping this in Todoist?

gnapse: Totally. Ideally we shouldn't ever need to tinker with z-index in app code, and our component library would take care of all that transparently. But until we're in a state where our apps are mostly using basic building blocks from the component library, and only working at a higher level, we may need to still adjust things from time to time.

gnapse (approved):
Can approve after the update without the hardcoded z-index.

Long term we should move more and more ownership of the various z-index levels over to Reactist components and have the app deal less and less with those details. This seems a good first step towards that end.

PR #336, opened 2020-09-25
Make tooltip decide not to show under certain unwanted conditions

gnapse:
## Short description

There's an issue I noticed with the new tooltip in combination with being applied to buttons that open a menu, dialog or other kind of popover. When the thing that the button opens is closed, the focus is moved to the button that triggered it. This part is expected and desired, a key part of proper keyboard accessibility. However, it seemed a bit out of place that, if the button also has a tooltip, this focusing back to the button would trigger the tooltip to appear.

This has been secretly preventing me from getting more aggressive in migrating more menus in twist to use this new menu component. This PR unblocks me.

For more details check [my raising of this issue in the reakit repo](https://github.com/reakit/reakit/discussions/749). All the credit for the solution goes to reakit's creator and maintainer @diegohaz, as you can see there.

## Test plan

Fire up storybook and go check the menu component story. Try to reproduce the issue as seen in this screen recording:

CleanShot 2020-09-23 at 23 15 07

Make sure you can't reproduce. That is, the tooltip should not show if you close the menu in any way (not even by pressing escape)

Also, make sure the tooltip is shown in all the usual circumstances (tabbing to the button, hovering, etc.)

## PR Checklist

## Versioning

This should generate a new patch release, as outwardly nothing has changed.

## Extra info

This PR also introduces the use of [testing-library/user-event](https://github.com/testing-library/user-event) to replace `fireEvent`. Check that repo for the details of what the differences are, or feel free to ask me why I think we should invest on it more heavily.


gnapse:
Hey @craigcarlyle, in addition to having added you as a reviewer, I wanted you to take care to seeing that it gets merged and released after being approved. In case I don't get to do that before I sign off today for an entire week (which is what seems more likely right now).

src/components/Tooltip.tsx
```diff
@@ -51,10 +51,33 @@ function Tooltip({
...
+     * Prevents the tooltip from automatically firing on focus all the time. This is to prevent
+     * tooltips from showing when the trigger element is focused back after a popover or dialog that
+     * it opened was closed. See link below for more details.
+     * @see https://github.com/reakit/reakit/discussions/749
+     */
+    function handleFocus(event: React.FocusEvent) {
+        // If focus is not followed by a key up event, does it mean that it's not
+        // an intentional keyboard focus? Not sure but it seems to work.
+        function handleKeyUp(e: Event) {
+            if ((e as KeyboardEvent).key !== 'Escape') tooltip.show()
```
craigcarlyle, line 64: This will cover the use case with the `Escape` key, but if a menu item is activated with an `Enter` or `Space` key the same problem will persist. 

I was thinking we could probably show/hide the tooltip if the focus event was caused by a user or programatically, but it looks like React's `FocusEvent` does contain `originalEvent`. Is there a way to check events like this in React?

gnapse: > This will cover the use case with the Escape key, but if a menu item is activated with an Enter or Space key the same problem will persist.

Yup. I realized this after the fact and left a comment in the linked conversation in the reakit repo. I figured it is acceptable as is because after all it affects keyboard-centric users but unblocks us from the main worrying group: mouse-centric users complaining about this.

However, if you have an idea and can take a shot at it, go for it. Just don't let this take too much of your time if you don't get lucky quickly. I can resume this after my time off if needed.

PR #352, opened 2020-10-05
Menu backdrop/overlay

gnapse:
## Short description

I took as my housekeeping task this week adding support for the Reactist menu to be usable as a context menu. While I wasn't able to achieve that yet, I realized that our menu does not have a backdrop/overlay, and since we recently decided that we want this as a default, I added it built into the menu component. So at least I got something done. Will resume the effort on context menus later in the week if I have time.

In the process I also upgraded the reakit version, as I noticed there were some new ones. I also set it fixed to a specific version as I believe this is a very sensitive dependency and we can't take changes with minor changes having some consequences. We'll take care of manually upgrading it from time to time (I certainly will).

## PR Checklist

-   [ ] Added tests for bugs / new features (could not do it, as elements do not position in jsdom as they do in the browser, so I cannot achieve the following: fire a click in the button and have the tests fail that click because an overlay is blocking it)
-   [ ] Updated docs (storybooks, readme) (N/A?)

## Versioning

This should be a patch release, or we can keep it in queue until new changes prompt us to make a release. I think there's no rush in having this released on its own, as it is a very minor change.


craigcarlyle (approved):
Looks good! 🚀

craigcarlyle:
This also closes https://github.com/Doist/reactist/pull/349.

PR #359, opened 2020-10-14
Fix @typescript-eslint/unbound-method by using class property arrow.

proxi:
## Short description

Fix `@typescript-eslint/unbound-method` rule violations and re-enable the rule in preparation to adopt shared [eslint-config](https://github.com/Doist/eslint-config). For context, see discussion in [here](https://github.com/Doist/twist-web/pull/2783).

The fix uses class property syntax. I know we've had some problems with transpiling it in other projects, so please double-check it's ok. I've run the tests and stories and checked output in es directory.

## PR Checklist

## Versioning

It only affects internals and does not require release of its own.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,8 +35,7 @@
         "react/no-did-update-set-state": "error",
         "react/no-find-dom-node": "off",
         "@typescript-eslint/explicit-module-boundary-types": "off",
-        "@typescript-eslint/ban-ts-comment": "off",
-        "@typescript-eslint/unbound-method": "off"
+        "@typescript-eslint/ban-ts-comment": "off"
     },
     "overrides": [
         {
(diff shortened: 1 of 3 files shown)


gnapse:
> we've had some problems with transpiling it in other projects, so please double-check it's ok.

For this reason I request that we do not merge it right now. I'm about to perform a release soon and would like more time to check this change locally against both Todoist and Twist.

proxi:
@gnapse Makes sense, please merge it when you find it convenient.

proxi:
We could also consider ignoring existing occurrences, but given it's only few and that it sometimes produces real issues (we have one or two in twist-web) I think cleanup makes sense.

proxi:
@gnapse Can it be merged now?

gnapse (approved):

PR #360, opened 2020-10-14
Re-enable no-unsafe-* rules in main codebase

proxi:
## Short description

no-unsafe-* lint rules are important, because violations have potential to produce a runtime crash. This PR re-enables all no-unsafe rules in main codebase, but allow exception in storybook and tests (hard to fix there due to older storybook and complicated mock typing).

Popover and Tooltip are not 100% strict and a couple of violations are ignored there. I never figured out how to retrict `children` type(s).

## Versioning

Only affects linting / types, does not need a release.

### DIFF
--- a/src/components/Tooltip.tsx
+++ b/src/components/Tooltip.tsx
@@ -71,6 +71,7 @@ function Tooltip({
         event.currentTarget.addEventListener('keyup', handleKeyUp, { once: true })
         // Prevent tooltip.show from being called by TooltipReference
         event.preventDefault()
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
         if (typeof child.props.onFocus === 'function') child.props.onFocus(event)
     }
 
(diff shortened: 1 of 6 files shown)


src/components/Tooltip.tsx
```diff
@@ -71,14 +71,15 @@ function Tooltip({
         event.currentTarget.addEventListener('keyup', handleKeyUp, { once: true })
         // Prevent tooltip.show from being called by TooltipReference
         event.preventDefault()
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
         if (typeof child.props.onFocus === 'function') child.props.onFocus(event)
```
gnapse, line 75: I'm not sure I understand why this call is unsafe.

gnapse (approved):
I'm sure this is all for the best and for some reason, but I'm not sure why what I commented on is considered unsafe. Would love to clarify that, because it seems legit to me.

src/components/Tooltip.tsx
proxi: I tried answering your question and I couldn't without digging in :| I'm afraid the reasons are varied and not obvious, but  the top issue is that `child.props` is `any`, and `any` could be `null` or `undefined`. The check would be valid if child.props were something like `Record<string, unknown>` but I don't know if it's even doable.

proxi: And if we got `any` out of the way, the answer probably boils down to https://github.com/microsoft/TypeScript/issues/21732

PR #374, opened 2020-10-17
Re-enable expect-expect lint rule.

proxi:
## Short description

Re-enable expect-expect lint rule. There doesn't seem to be good way to avoid it triggering in case of delegating test to another function, but it's better to address on per-case basis than disable everywhere.

https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/expect-expect.md

Only affects lint.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -30,7 +30,6 @@
         "webpack.config.*"
     ],
     "rules": {
-        "jest/expect-expect": "off",
         "prettier/prettier": "error",
         "react/no-did-mount-set-state": "error",
         "react/no-did-update-set-state": "error",
--- a/src/components/__tests__/Tooltip.test.tsx
+++ b/src/components/__tests__/Tooltip.test.tsx
@@ -61,6 +61,8 @@ describe('Tooltip', () => {
         })
     })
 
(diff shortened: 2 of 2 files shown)


henningmu (approved):

gnapse:
I've developed recently a feeling of not being a fan of this rule. I don't see it being a big help, and it is most of the time a bit of an annoyance. Especially while writing tests, having the entire test block highlighted by the editor as problematic, all the way until you end up writing an `expect`, it's a bit too much.

The good that it brings is objectionable. Tests without `expect` can be perfectly valid, and for the cases where they aren't, I feel this is something much better judged by humans than by an automated rule. The potential downside of having some legitimate case slip through does not outweigh the downside of enabling the rule.

proxi:
@gnapse Please open an issue or PR in https://github.com/Doist/eslint-config This was just about cleaning up this repo, so we don't have to override the rule. We could disable it everywhere. Though IMO it's not worth complicating config it it's only a problem in a few places.

gnapse:
Sure, makes sense. Done https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/expect-expect.md

PR #390, opened 2020-11-13
Moves button types to separate types

scottlovegrove:
## Short description

This creates separate types for `ButtonVariant` and `ButtonSize`. This makes things easier for consumers of the package to have tidier code when trying to set the value for these dynamically

For example, instead of a method signature like this: 
```
function getVariant(style?: ActionStyle): 'link' | 'primary' | 'secondary' | 'danger' | undefined {}
```
you would now be able to have 
```
function getVariant(style?: ActionStyle): ButtonVariant {}
```

## PR Checklist

## Versioning

This is just a patch release, no breaking changes.


gnapse (approved):
I considered `ButtonProps['variant']` to fill this need, but I realize that strictly speaking it also allows `undefined`, so this makes sense.

PR #399, opened 2020-11-27
Adds a TodoistQuickAdd component

amix:
## Short description

We want to make it easy to add the Todoist Quick Add in other React apps.

Note, this is a draft right now as it requires the deployment of to function properly:

* https://github.com/Doist/todoist-web/pull/2320

## PR Checklist

## Versioning

It's a new feature.


gnapse:
I feel like this should be a library on its own, and not part of Reactist. It does not even use much of Reactist (if at all), and for consumers of this it would make more sense to have something that would give them only this (e.g. proper tree-shaking usually depends on the consumer of this having things set up properly in their build system, so they might be bringing in all of Reactist just for this component).

I can help set up a skeleton lib package (using https://tsdx.io) where this could fit in without much extra work.

henningmu:
I thought it wouldn't hurt much, but maybe you're right 👍

gnapse:
I'm not sure, maybe I am overthinking it. It would be good to know what others in our team think cc @Doist/frontend-product 

Additional part of my thinking: Reactist, in my mind, is a reusable set of components that implement Doist's design system. It should have app-agnostic components and utilities that would then allow to build app-specific components and utilities.

amix:
@gnapse, this could make sense. Do you have an example of state-of-art small lib I could get inspired by?

gnapse:
> Do you have an example of state-of-art small lib I could get inspired by?

Not exactly. But I'd suggest we use https://tsdx.io which is a tool to generate a lib skeleton that comes ready out of the box with all the boring tooling already set up, support for TypeScript, etc. I can help if needed, but probably following their quick start guide should be enough. A quick read shows that you need to run `npx tsdx create todoist-quick-add-react` and you'd get a skeleton with a `src/` folder where you can place this component in, and let it be its main export on the `index.tsx` file and that's it.

src/components/TodoistQuickAdd.tsx
```diff
@@ -0,0 +1,134 @@
...
+    onLoadingSuccess,
+    onLoadingError,
+}: Props): null {
+    const [isLoaded, setIsLoaded] = useState<boolean>(false)
+    const loadingTimerToClear = useRef<number | undefined>(undefined)
+
+    // Set up post message listener
+    useEffect(() => {
+        function handleMessageReceived(event: CustomMessageEvent) {
+            const data = event.data
```
proxi, line 60: I'm no expert at this, but I believe for security reasons it is important that `origin` of message is verified.

src/components/TodoistQuickAdd.tsx
```diff
@@ -0,0 +1,134 @@
...
+            if (data && typeof data === 'string' && data.indexOf('Todoist Quick Add SDK')) {
```
proxi, line 61: There is a problem with such comparison pattern: any string message containing `Todoist Quick Add SDK` will pass the test. In bigger app like Twist or 3rd party app there are lots of messages floating around, and if someone e.g. types `Todoist Quick Add SDK` into some input box, it could trigger a message that would pass.

Also see my comment here about possibly sending an object: https://github.com/Doist/todoist-web/pull/2320/files#r531756753

src/components/TodoistQuickAdd.tsx
```diff
@@ -0,0 +1,134 @@
...
+                const json = JSON.parse(data) as MessageJSON
```
proxi, line 62: It doesn't seem very exploitable now, but for something as critical as public library meant to be used by 3rd party apps, I think this needs proper type check. Otherwise it's easy to inject poisoned message, and as library is later modified it could lead to security hole.

proxi:
@amix Great idea. I left two comments about potential security issues - I'm no expert, but I think we need to be paying much more attention to this, especially in code meant to be used outside.

+1 on keeping it separate from Reactist, it doesn't seem to belong here + it introduces dependency on Todoist.

Also, it isn't really quite working for me yet - not at all in Safari (always failing to load) and not much in Chrome (failing to load after showing quick add). Does it need for outside scripts to be deployed first?

src/components/TodoistQuickAdd.tsx
amix: @proxi this is already fixed in the latest version 👍

src/components/TodoistQuickAdd.tsx
amix: Good point. I'll make the origin check explicit.

src/components/TodoistQuickAdd.tsx
amix: Fully agree, @proxi 👍

amix:
Thanks for the review, @proxi. I am doing another jab at this where it's an independent typescript based lib.

> Also, it isn't really quite working for me yet - not at all in Safari (always failing to load) and not much in Chrome (failing to load after showing quick add). Does it need for outside scripts to be deployed first?

It needs to use this todoist-web extension: https://github.com/Doist/todoist-web/pull/2320

amix:
Closing this down as I'll create a standalone script for the Todoist QuickAdd 👍

PR #406, opened 2020-12-09
Organize styles during watch mode

henningmu:
## Short description

In recent product PRs we've fixed our local reactist setup. This went well for code changes but didn't work for style changes. Also it required to run `npm run build` before running the watch mode with `npm start`. With this PR this gets fully fixed, we no longer need to run build before the watch mode and we can change styles and see them reflected in the apps.

I tried to find a solution that is not based on bash scripts but integrates with tsdx / rollup but couldn't make it work with the copy and delete plugins (the order of these plugins cannot be specified in the way we need it because the hooks are not granular enough).

You can test this PR by running `npm run clean`, then `npm start` and then start up todoist or twist web and see your changes reflected immediately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "scripts": {
         "setup": "npm install && npm run validate",
         "validate": "npm run lint && npm run type-check && npm run test",
-        "start": "tsdx watch",
+        "start": "tsdx watch --onSuccess \"./scripts/organize-styles.sh\"",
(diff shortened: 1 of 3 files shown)


rfgamaral (approved):
Everything's working as expected, nothing to add 👍

frankieyan (approved):
Oops, at least the PR on the Twist side is still open. We'll need to update the docs again on the TD side 😃

PR #410, opened 2020-12-11
feat: Add notification component

frankieyan:
## Short description

This adds a new notification component.

[Spec](https://paper.dropbox.com/doc/Todoist-Twist-Web-Improving-version-update-adoption-rate-Design--BBFIHgL2tAcgaLLDi0XLfT1lAg-g0TjJJG5HQHivQYCGK4fT)
[Mocks](https://app.zeplin.io/project/5fb56b8992ad8e77a04e7804/dashboard?sid=5fb56bfe6820ababe6732504)

To see this in action, checkout https://github.com/Doist/twist-web/pull/3023

## PR Checklist

## Versioning

This will bump a minor version to v7.2.0


frankieyan:
@Doist/frontend-product this is still WIP, but I wanted to get your opinion on whether this: https://github.com/Doist/reactist/blob/47c7233ab33ea38ab83ed6b7dd1906c5acf56d2f/src/components/notification/notification.tsx#L15-L23 is a pattern we should move forward with, at least until we have a stable design system in place.

henningmu:
Do we really need that much flexibility for styling? Less options probably force a better behaviour :D 

What we've done in the past is allowing one `className` applied to the outermost element. Since reactist class names aren't minified during publish, consumers (twist / todoist) can then override the reactist names to their liking in their styling (see e.g. `styles/reactist.less` in Twist). Why not stick with that for now?

frankieyan:
While the result is more or less the same since we're still overriding styles from Reactist, I was thinking we'd start treating class names as internal to components, as we'll likely introduce CSS modules for Reactist and making these classes local. 

Maybe we can limit the number of elements we expose, I'm still not sure what I'll need to change in each application yet, but I think this can create a clearer boundary between Reactist and its consuming applications.

henningmu:
>  I was thinking we'd start treating class names as internal to components, as we'll likely introduce CSS modules for Reactist and making these classes local.

Personal opinion ahead: let's don't overthink this here and now and stick with what we have. The final goal of reactist customization is not to allow customizing everything but only placement and theming. Once we figure out theming (maybe by exposing multiple classes like you propose here) we should change it in one swoop.
I can see your point though and if you insist I think we can go ahead with this approach – thanks to TS generating documentation for this should be "easy"

frankieyan:
Yup, no not insisting at all which was why I wanted to open it up for discussion before the PR is ready. Makes sense to not deviate too much for now when our goal is to minimize overrides in the long run 👍

src/components/notification/notification.tsx
```diff
@@ -0,0 +1,101 @@
...
+    onClose,
+    closeAltText = 'Close',
+    className,
+    ...rest
+}: NotificationProps) {
+    const titleId = title ? `${id}-title` : null
+    const subtitleId = subtitle ? `${id}-subtitle` : null
+    const contentId = children ? `${id}-content` : null
+    const ariaLabelledBy =
+        children && contentId
```
henningmu, line 36: If `children` is true-ish `contentId` is as well, so only checking for one should be enough
```suggestion
        contentId
```

src/components/notification/notification.tsx
```diff
@@ -0,0 +1,101 @@
...
+            ? { 'aria-labelledby': contentId }
+            : titleId
+            ? { 'aria-labelledby': titleId }
+            : null
+    const ariaDescribedBy = subtitleId && !children ? { 'aria-describedby': subtitleId } : null
+
+    const notificationContent = (
+        <div id={contentId ?? ''} className="reactist-notification__content">
```
henningmu, line 44: Instead of using nullish coalescing here we could just fallback to empty string instead of `null` above? Same for the other ids.

After writing this I asked myself should we add empty ids to the DOM at all?

henningmu (approved):
It's sooo beautiful 🤩

src/components/notification/notification.tsx
frankieyan: Yeah I guess we should spread it like the aria attributes

src/components/notification/notification.tsx
frankieyan: Think I originally just checked for children here which made TypeScript angry 😄, but yeah checking for `contentId` is enough 👍

PR #414, opened 2020-12-18
chore: Colocate everything

henningmu:
## Short description

As discussed internally, we should colocate all things that belong together (components, tests, styles, (soon stories), ..). Also we decided to switch to an all-lower-case camel-case notation. This PR changes both things.

Originally, I wanted to upgrade Storybook but given our current situation that will require more drastic changes. That's why I'm introducing colocation first and then move all the stories over to the new story format and docs.

To make reviewing more enjoyable here is a meme 😂

colocate

## PR Checklist

## Versioning

_none, should be internal only 🤞_


henningmu:
@Doist/frontend-product would be nice to get another pair on eyes on this. Theoretically, it only moves files around but I want to make sure to not break anyones setup 👍

gnapse:
No need for a meme. You had me at "this PR changes both things". Especially the lower case dasherized file naming scheme. Will review right away.

gnapse (approved):
Looks good the way the changes are made. As you say, it looks like it should not affect outward usage.

Checked this while running Todoist locally against the local version of Reactist in this branch, and it all worked. The couple of Reactist components I recall are being used in Todoist.

I tried checking this on twist but ran into some difficulties to get twist-web running properly in the browser. But nothing related to this PR. Maybe some issues after so much time without running twist locally. Will figure it out tomorrow.

PR #423, opened 2021-01-08
feat: Adds Date and Time Pickers

scottlovegrove:
Fixes: https://github.com/Doist/twist-zoom/issues/54
Fixes: https://github.com/Doist/twist-zoom/issues/53

## Short description

Adds custom date and time pickers

## PR Checklist

## Versioning

This will bump a minor version to v7.3.0


gnapse:
I did not get to review this today, but will do on Monday.

package.json
```diff
@@ -123,6 +124,7 @@
     },
     "dependencies": {
         "dayjs": "^1.8.10",
+        "pikaday": "^1.8.2",
```
henningmu, line 127: I know this is copied from Twist, but reactist should set higher standards. [pikaday is no longer being maintained](https://github.com/Pikaday/Pikaday/issues/884) with over 200 open issues – I don't think we should continue using it and rather research alternatives.

src/components/DatePicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+    /* minimum date permitted */
+    min?: string
+    /* maximum date permitted */
+    max?: string
+    defaultValue?: string
+    /* Additional css class that is applied to DatePicker */
+    className?: string
+}
+
+class DatePicker extends React.Component<Props> {
```
henningmu, line 27: New reactist components shouldn't be class components

src/components/DatePicker.tsx
```diff
@@ -0,0 +1,68 @@
...
+    componentDidMount(): void {
+        const { onDateChanged, min, max } = this.props
+
+        new Pikaday({
+            theme: 'pika-twist-composer-input',
```
henningmu, line 32: reactist shouldn't contain explicit twist references

henningmu (changes requested):

package.json
scottlovegrove: Ugh, ok. Were there any investigations into this already done for Twist?

henningmu: Not that I'm aware of. In Todoist we use a custom implementation because the calendar needs to be a bit more involved. Not sure what's best to use maybe @gnapse knows a good implementation of a simple date picker?
Maybe we could use the native calendar? – looks a bit shabby but could work for these one-off selection scenarios 🤷‍♂️

scottlovegrove: > Maybe we could use the native calendar?

@henningmu The problem with the native date picker is that it can't be styled. So if you are in the dark theme in twist, the calendar always appears with a white background. It's one of the main reasons I opted to try and do the date picker like this (I was anticipating design asking for it to be styled).

scottlovegrove: @gnapse Any thoughts on this? Any recommendations for datepickers you have?

gnapse: Not off the top of my head. Can I suggest, before we make a rush decision here, to leave out this from Reactist, and for the time being, add it to the context where you need it in the short term?

scottlovegrove: Ok, will remove it for now.

src/components/DatePicker.tsx
scottlovegrove: As this component has been removed for now, this is now resolved.

scottlovegrove:
I've removed the DatePicker component for now and tidied up the commit history. @henningmu @gnapse

src/components/TimePicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+    min,
+    max,
+    defaultValue,
+    className,
+    optionsClassName,
+}: Props): JSX.Element {
+    const times = [
+        {
+            value: '',
+            text: '--:--',
```
gnapse, line 37: we could stick to track the value only, and generate the text from the value

src/components/TimePicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+    onTimeChanged?: (value: string) => void
+    min?: string
+    max?: string
+    defaultValue?: string
+    className?: string
+    optionsClassName?: string
+}
+
+export default function TimePicker({
+    minutesInterval,
```
gnapse, line 26: you can set up the fallback value if absent right here when de-structuring:
```suggestion
    minutesInterval = MINUTES_INTERVAL,
```

src/components/TimePicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+    onTimeChanged,
+    min,
+    max,
+    defaultValue,
```
gnapse, line 30: Why this name? Is this a [controlled](https://reactjs.org/docs/forms.html#controlled-components) or an [uncontrolled](https://reactjs.org/docs/uncontrolled-components.html) component? If it's a controlled component (as it seems to me it is at first sight) then this should be called `value` to follow the convention.

src/components/TimePicker.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+function zeroPad(i: number): string {
+    return `0${i}`.substr(-2)
+}
+
+function breakUpTime(value?: string): Dayjs | undefined {
+    if (!value) return
+
+    const time = dayjs(value, 'HH:mm')
+    return time
```
gnapse, line 80:
```suggestion
    return value ? dayjs(value, 'HH:mm') : undefined
```

src/components/__tests__/TimePicker.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+    const timePickerTestId = 'time-picker'
+
+    describe('no min/max entered', () => {
+        it('displays the correct time ranges with default time span', () => {
+            render(<TimePicker />)
+
+            const select = screen.getByTestId(timePickerTestId) as HTMLSelectElement
+
+            expect(select).toBeInTheDocument()
+
```
gnapse, line 15:
```suggestion
```

Not needed. By querying with any `screen.get*` function, it already guarantees that what you queried for was found.

src/components/__tests__/TimePicker.test.tsx
```diff
@@ -0,0 +1,150 @@
...
+            const options = extractOptions(select.options)
```
gnapse, line 16: I suspect you do not need this `extractOptions` helper. You should get the same (or almost the same thing) with:

```suggestion
            const options = screen.getAllByRole('option')
```

src/components/icons/ClockIcon.svg.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react'
+
+const ClockIcon = () => (
```
gnapse, line 3:
```suggestion
function ClockIcon() (
```

src/components/select/select.tsx
```diff
@@ -23,23 +23,35 @@ type Props = {
...
 }
 
-function Select({ value, options = [], onChange, disabled = true, className = '' }: Props) {
+function Select({
+    value,
+    options = [],
+    onChange,
+    disabled = true,
+    className = '',
+    optionsClassName,
```
gnapse, line 36: I was going to object to adding this prop in the time picker, but I did not do so because I saw it being passed to the Select, and thought it was already available there.

But now I see it is being added here too.

My objection is not super strong, because I still think all these components have a lack of composability in how their APIs are designed. But I still think it may not be needed even in the short term. You can always apply a class name to the parent select, and in css you can style the options by using a selector such as `.selectClassName option`.

gnapse (commented):
When I suggested not adding these components here now, I was referring to both. Although I do not object to adding the time picker, I think in general, for the future, we're better off incubating new components for a while in their place of origin, and then see what's the best way to extract them into Reactist.

My main concern is that we may hastily design a public interface for a component, only to realize later on that it could be designed in a different, more convenient and more reusable way.

Anyway, don't take this as an objection in this case, because for this particular component I think the risks are low.

(In general the risks with Reactist are low, because I assume that Doist is probably the only place where Reactist is used (or one of a few) but I'd still like to set a higher standard of publishing public APIs once they are sound enough).

src/components/select/select.tsx
scottlovegrove: > and in css you can style the options by using a selector such as .selectClassName option

I was already doing this when the component was in twist, but once I moved it to reactist, the css I had for this stopped working and the only way I could get it styled was to do it like this 🤷🏻‍♂️

src/components/__tests__/TimePicker.test.tsx
scottlovegrove: When I originally had this in twist-web, I was able to just do `const options = [...select.options]`, but that wasn't available to me in reactist, something about es6 (I think).

scottlovegrove:
> When I suggested not adding these components here now, I was referring to both. Although I do not object to adding the time picker, I think in general, for the future, we're better off incubating new components for a while in their place of origin, and then see what's the best way to extract them into Reactist.

I'm happy to do this, I just know that it was @henningmu's suggestion to put these in reactist, so would like to get his thoughts on this.

src/components/TimePicker.tsx
scottlovegrove: I think this comes from the default input element doesn't it? 
image

gnapse: Indeed, but the default input element has both `defaultValue` and also a `value` prop. You can use it in one of those two modes: controlled or uncontrolled. When you use it as a controlled input, you use `value`, and `defaultValue` is for initialization purposes only. That is, `defaultValue` only provides the initial value with which it is rendered, and subsequent changes to its value is not managed by React (that is, there's no need for an `onChange` handler).

However, in this case it seems to me that this component only supports the controlled mode. That is, you provide its value as a prop at all times, and the context where this component is rendered is supposed to manage its value as state, and listen to attempts to change it via the `onChange` prop. If this is so, the customary thing to do is to have a `value`/`onChange` pair of props. No need to have a `defaultValue` when it does not support the uncontrolled mode.

src/components/__tests__/TimePicker.test.tsx
gnapse: Oh ok, something to look into in how reactist is setup. But still, the alternative I shared should work regardless.

src/components/select/select.tsx
gnapse: Ok, not sure what was the problem in that case, but we should not add a prop to a public API in a library for a short term fix. Once you add it, it becomes a breaking change to take it back. We need to figure out why it does not work.

gnapse:
Sure, I also think they eventually need to end in Reactist. I'd just prefer that we incubate them a bit on their first use, before we bring them here.

For example, I'd like to research on other time and date picker alternatives out there to at least draw inspiration on how they are designed (the API side of their design, not their visual design). Once you commit to a public interface in a public library, many changes that we only realize after the fact become drastic breaking changes that are harder to pull off once the library is being used in different projects.

scottlovegrove:
> Sure, I also think they eventually need to end in Reactist. I'd just prefer that we incubate them a bit on their first use, before we bring them here.

That makes sense. @henningmu are you ok with that approach?

scottlovegrove:
Ok, so I'm going to make an executive decision here and close this PR for now. I want to speed things along a little bit with the integrations I'm working on, so I'm quite happy for these components to live, and mature, in Twist first. I will open a PR on twist to bring these components in with the feedback above applied.

src/components/TimePicker.tsx
scottlovegrove: By doing it this way, it meant I could just drop the times into the `options` property on the `Select` without having to do any additional work.

PR #426, opened 2021-01-20
Fix type definition for MenuItem when no as attribute

### LINKED ISSUE #421: Type definitions for components with dynamically typed elements does not work when using the default element type
Originally discovered by @rfgamaral in https://github.com/Doist/twist-web/pull/3049

## Describe the issue

When using one of the more recently developed components, those with the ability to change the base element type they render with (e.g. `<MenuItem as="a" />`) if we omit the `as` prop, the props type is not inferred correctly.

### How to reproduce

Try adding obviously invalid props to a `MenuItem` in a project using Reactist:

```jsx
// foobar is not flagged as an error
<MenuItem foobar onSelect={handleSelect}>Click me</MenuItem>
```

If you do it in a `MenuItem` without an explicit `as` prop, you should notice that TypeScript does not flag any error.

However, if you do it in a `MenuItem` with an explicit `as` prop you'll get an error:

```jsx
// foobar is not flagged as an error in both cases
<MenuItem as="a" foobar href="/">Click me</MenuItem>
<MenuItem as={Link} to="/" foobar>Click me</MenuItem>
```

See in the screen recording below how the error underline appears under `foobar` only when the `as="button"` is present.

CleanShot 2021-01-06 at 19 16 50

clempat:
Closes #421 

## Short description

Not having 'as' attribute defined, results into TT not being defined and probably allow any props as result. I added as default type for TT 'ComponentType', which looks making duplicates but working well.

I deleted the second line which then become useless. 

Here some examples:
photo_2021-01-20_09-45-13

## PR Checklist

-   [x] Executed `npm run validate
` and made sure no errors / warnings were shown

### DIFF
--- a/src/components/menu/type-helpers.ts
+++ b/src/components/menu/type-helpers.ts
@@ -74,8 +74,9 @@ interface ForwardRefFunction<ComponentType extends As, ComponentProps = {}> {
 }
 
 export interface ForwardRefComponentWithAs<ComponentType extends As, ComponentProps> {
-    <TT extends As>(props: PropsWithAs<TT, ComponentProps>): React.ReactElement | null
-    (props: PropsWithAs<ComponentType, ComponentProps>): React.ReactElement | null
+    <TT extends As = ComponentType>(
+        props: PropsWithAs<TT, ComponentProps>,
+    ): React.ReactElement | null
     readonly $$typeof: symbol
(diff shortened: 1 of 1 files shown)


rfgamaral (approved):
Looks good to me :) Thank you for your contribution 👍

gnapse (approved):

PR #433, opened 2021-02-05
Adds defaultValue and otherProps to Select

scottlovegrove:
## Short description

Adds a prop for `defaultValue` as this would be useful for instances where you can't rely on `useState` in order to update the `value` field, but still want the `Select` to have a value by default.

I've also added an `otherProps` field collection to propagate any additional props into the select (so that things like `test-data-id` can be set against the Select itself.

## PR Checklist

## Versioning

patch, see changelog entry.


henningmu (approved):
LGTM 🙌

gnapse (approved):

PR #434, opened 2021-02-08
feat: Add Box, Columns, Inline, and Stack design system components

gnapse:
## Short description

Introduces a set of new layout components and the start of a new themeable design system.

## PR Checklist


.storybook/config.js
```diff
@@ -8,5 +9,13 @@ setOptions({
...
 })
 
-const loadStories = () => require('../stories')
+addDecorator(withKnobs)
+
+const req = require.context('../src', true, /.stories.tsx$/)
+
+function loadStories() {
+    require('../stories')
+    req.keys().forEach((file) => req(file))
```
frankieyan, line 18: If we want to keep using the `new-components` folder, it'd be nice to sort the component path here so the components are still sorted alphabetically in storybook.

src/new-components/columns/columns.module.css
```diff
@@ -0,0 +1,142 @@
+/* Default horizontal spacing */
+.space-xsmall {
+    margin-left: calc(var(--reactist-spacing-xsmall) * -1);
```
frankieyan, line 3: Will we go back to .less here to keep IE11 support for now?

frankieyan (commented):
@gnapse I pushed a couple of commits that allow the css-loader rule to be defined separately. One caveat is that by using tsdx, our dist build is now built with rollup and so is configured separately from storybook, so tsdx.config.js has to be updated as well.

I've tested it in Twist and the styles are picked up correctly now:
image

image

src/new-components/columns/columns.module.css
gnapse: I moved away from `.less` on purpose. Like I mentioned at some point, Todoist will have to see how to deal with this while we cannot drop IE support. Probably by not using the components off of this repo for a while. I know it sucks, but I did not want our implementation here to be hindered by having to support IE11. Also, CSS custom properties allow apps using these components to customize the variables at the root and the design system will pick it up.

frankieyan: No objections from my side, but this will probably mean extra work for you to keep reactist updated and to avoid the two implementations from diverging. Would you be ok with that?

gnapse: Yes. It would mean extra work either way (I already did the work here and taking it all back to .less would mean more work, plus having to import a bunch of .less variables all over the place is a big overhead as well. Not to mention that we'd need to write the CSS for dark mode twice.

.storybook/config.js
gnapse: I sorta get what you mean, but not quite exactly. Do you mind contributing that change? Seems small and you have the context because it is your idea. (unless what you mean is a simple `.sort()` on this array before calling `forEach()`).

gnapse:
@frankieyan, there were no big changes to bring (only a new prop in the `Inline` component). This because I have only brought the layout components (and `Text` and `Heading`) and not the other ones (the form components).

I made other improvements not prompted by changes in the Todoist components.

This is ready as far as I'm concerned. (to be mergeable into a beta/alpha version of reactist, we still have a lot to do, like write tests, bring in more components, etc.)

frankieyan:
@gnapse I was able to see the errors locally and they are fixed with the following commits:

* 837916f
  The `classnames/types` module exists, but eslint wasn't happy with it. Since the type package's organization are internal implementation details, I've updated it to extract the type from the top-level classnames export instead: `type ClassValue = Parameters<typeof classNames>[number]`
 
* 67eadbd
  `HTMLHeadingElement` doesn't seem to be the correct type if we wanted to extend props from a JSX element. This has been changed to `JSX.IntrinsicElements['h1' | 'h2' | etc]` instead

I'll create a beta branch, get this merged, and then have a beta version released 👍

frankieyan (approved):

PR #441, opened 2021-02-18
chore: Fix npm/node to v6 and latest LTS

frankieyan:
## Short description

This uses the `engines` field to enforce npm to be within the v6 range when running `npm i`, as using a newer version will change the format of the lock file. Note that `npm ci` doesn't respect this field (yet).

While we should move to npm v7 in the future, we need this to be coordinated within the team so that the lock file doesn't get drastically changed between PR's. 

## Test Plan
- [x] When using the wrong version of node or npm, `npm i` is blocked
- [x] Storybook runs successfully
- [x] Assets can be built successfully

## PR Checklist

## Versioning

This should be treated as a breaking change, as projects that are on older versions of node/npm will run into warnings when installing Reactist as a dependency, and cases where `engine-strict` is specified will result in an error.


henningmu (approved):
Awesome. I'm using nvm instead fnm and it worked perfectly 👏

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+## 9.0.0
+- [Build] The project now requires node v14.5.5+ and npm v6.14.11+ to install and run.
+- We're skipping right over v8.0.0 as its tag has already been created previously
```
henningmu, line 7: Could use some prettier here

package.json
```diff
@@ -2,7 +2,7 @@
     "name": "@doist/reactist",
     "description": "Open source React components by Doist",
     "author": "Henning Muszynski <henning@doist.com> (http://doist.com)",
-    "version": "7.2.1",
+    "version": "9.0.0",
```
henningmu, line 5: Why not `8.0.0`?

frankieyan: The [v8.0.0](https://github.com/Doist/reactist/tree/v8.0.0) tag already exists upstream, and I wasn't sure if there'd be any issues if I force pushed a tag, or if it's even allowed from the releases page, so I figured it's easier to just skip a version.

(There's a note on this in the changelog document)

CHANGELOG.md
frankieyan: We should make this a part of lintstaged 😉

PR #467, opened 2021-03-10
fix: Move patch-package to dependencies

frankieyan:
## Short description

With #464 I added patch-package in order to patch rollup-plugin-style, however when being installed as a package with `npm ci`, patch-package will not be available and the install will exit.

This moves patch-package to our dependencies so that it'd be available. When installing as a package, it will omit patching dev dependencies, so other than the extra package being installed, there should be no additional consequences: [ref](https://docs.npmjs.com/cli/v6/using-npm/scripts#prepare-and-prepublish)

The alternative is to run patch-package with the [`prepare` script](https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts), which sounds like it covers all the scenarios that we would want it to run in, however, patch-package used to recommend the prepare script and have since changed it to postinstall: https://github.com/ds300/patch-package/commit/f9fb13b2a25c60991ff88f0efe0399f6f5f6f623.

## Test Plan

While not exactly the same as installing off of the registry, we can try installing this off of git and seeing the postinstall script run:

- In another repo, run `npm install git://github.com/doist/reactist#frankie/fix-patch-package`
- Run `rm -rf node_modules`
- Run `npm ci`
- [ ] You should see the following output:
```
> @doist/reactist@9.1.0-beta.1 postinstall /Users/frankie/Documents/twist-web/node_modules/@doist/reactist
> patch-package

patch-package 6.4.6
Applying patches...
No patch files found
```

Whereas if you try the steps above with v9.1.0-beta.1, it should fail

## PR Checklist

## Versioning

v9.1.0-beta.2

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8083,8 +8083,7 @@
         "@yarnpkg/lockfile": {
             "version": "1.1.0",
             "resolved": "https://registry.npmjs.org/@yarnpkg/lockfile/-/lockfile-1.1.0.tgz",
-            "integrity": "sha512-GpSwvyXOcOOlV70vbnzjj4fW5xW/FdUF6nQEt1ENy7m4ZCczi1+/buVUPAqmGfqznsORNFzUMjctTIp8a9tuCQ==",
-            "dev": true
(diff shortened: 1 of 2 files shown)


henningmu (approved):
Works as advertised 👏

PR #472, opened 2021-03-17
feat: Add aria-label prop to Loading component

frankieyan:
## Short description

Adds an optional `aria-label` prop to `<Loading>`

## PR Checklist

## Versioning

v9.1.0


craigcarlyle (commented):
Nice accessibility initiative @frankieyan! 

In addition to `aria-label`, what do you think about including an `aria-live="assertive"` attributes along with `role="alert"`?

As it's currently implemented, a customer would have to navigate to the loading indicator to understand that something is loading. Using the attributes above would make the screen reader announce the change without requiring navigation to occur.

Here are some links for references:
* https://codeburst.io/how-to-create-a-simple-css-loading-spinner-make-it-accessible-e5c83c2e464c
* https://www.carbondesignsystem.com/components/loading/accessibility

stories/components/LoadingStory.tsx
```diff
@@ -17,7 +17,7 @@ const LoadingPropTypesChapter = {
 
 const LoadingStory = () => (
     <section className="story loading">
-        <Loading />
+        <Loading aria-label="Loading" />
```
craigcarlyle, line 20: Could we set "Loading" as the default `aria-label` if one is not provided?

frankieyan: 💯

gnapse:
I think that more than an aria label, the loading component should be a role="alert" element. Thoughts?

PR #473, opened 2021-03-17
fix: Allow stylesheets to be bundled in storybook production build

frankieyan:
## Short description

Fixes Storybook's production build so that stylesheets are included again.

This one has actually been lingering around since v6.0.0 - because we declared our package as side-effect-free (and later having declared CSS files being the only side-effects), Webpack drops our *.less imports from the production build as we simply import them, rather than passing their reference around as we do with CSS modules.

See https://webpack.js.org/configuration/module/#rulesideeffects and https://webpack.js.org/guides/tree-shaking for more details.

Since we don't expose any *.less files in the tsdx bundles, we should leave package.json as is and only declare this with Storybook's Webpack config.

## Test plan
- [ ] Open up the /docs folder in a browser (or use something like VSCode's [Live Server](https://marketplace.visualstudio.com/items?itemName=ritwickdey.LiveServer). Verify that styles are being loaded.

## PR Checklist

## Versioning

No need for a release here


henningmu (approved):
Great going for the root cause – works like a charm 👏

PR #477, opened 2021-03-26
fix: Call MenuItem onClick if provided

gnapse:
## Short description

A `MenuItem` element should support receiving a `onClick` event handler. A good motivation to do it is to catch the click event early, and possibly call `event.preventDefault()`, which should prevent the menu item's selection action to happen.

## PR Checklist

## Versioning

I consider this a bug fix, since the `onClick` handler is already supported in the prop types, but we were not calling it internally. Hence this marks a new patch release.


src/components/menu/menu.tsx
```diff
@@ -209,20 +209,22 @@ type MenuItemProps = {
...
     const { hide } = state
 
     const handleClick = React.useCallback(
-        function handleClick() {
-            const onSelectResult: unknown = onSelect ? onSelect() : undefined
+        function handleClick(event: React.MouseEvent<HTMLButtonElement>) {
+            onClick?.(event)
+            const onSelectResult: unknown =
+                onSelect && !event.defaultPrevented ? onSelect() : undefined
             const shouldClose = onSelectResult !== false && hideOnSelect
```
frankieyan, line 223: While it works, it doesn't seem very explicitly that closing the menu is the intended behaviour.

Perhaps the next line can include the `defaultPrevented` condition like: `(onSelectResult !== false || event.defaultPrevented) && hideOnSelect`

frankieyan (approved):

src/components/menu/menu.tsx
gnapse: I'm not sure I agree.

The condition to decide if the menu should be closed has not changed. We only prevent closing it if the `onSelect` action returned `false` (leaving aside that other `hideOnSelect` flag). So if `onSelect` did not return `false` we close it. Then `onSelect` cannot return `false` if it was never closed. I'd rather decouple the `shouldClose` value from the `event.defaultPrevented` value.

frankieyan: My concern is that this isn't very obvious. It would be good if we can document `onClick` in the props type, something along the lines of "If the provided event's `preventDefault` is called, then onSelect would not be executed and therefore unable to prevent the default auto-close behaviour"

gnapse: I can do that.

Though I would say this is pretty obvious. I would always expect that a `onClick` event, if it prevents default, will prevent the consequences of that click to happen. Just as it happens when you do that in a link `onClick`, which will prevent the link to trigger navigation.

gnapse:
@frankieyan any idea why the new package version has not been published yet? It seems more than enough time since I merged this. And I do not see the publish action triggered.

frankieyan:
The action is run when a release a created 👍

gnapse:
Ah, I thought a push was enough. Thanks!

PR #485, opened 2021-04-06
Bring changes from Todoist made to existing components

gnapse:
## Short description

Brings in the rest of the design system components that we've been incubating in Todoist's code for its new settings. Now becoming available in Reactist so we can use them in Twist's new settings as well.

## PR Checklist

## Versioning

This will introduce a new minor beta release.


src/new-components/text-link/text-link.module.css
```diff
@@ -0,0 +1,21 @@
...
+    font-weight: inherit;
+    font-family: inherit;
+    text-decoration: none;
+}
+
+.container:hover {
+    text-decoration: underline;
+}
+
+.container svg {
```
frankieyan, line 12: I think the component can be more explicit in accepting an icon. What do you think about accepting an `icon` prop that expects a component with a className prop? This way, the component also guarantees the order in which the icon is displayed (ie. always on the left).

src/new-components/columns/columns.tsx
```diff
@@ -39,6 +39,9 @@ const Column = forwardRefWithAs<ColumnProps>(function Column(
                     ? getClassNames(styles, 'columnWidth', width.replace('/', '-'))
                     : null,
             ]}
+            minWidth={0}
```
frankieyan, line 42: This will fix an issue I had with it not shrinking below the content's width 👍

frankieyan (approved):
It's a bit hard to see these components in action without having stories in place, but I trust that they are what we're using now in Todoist.

I don't think any of the existing components changes will conflict with their existing usage. For some reason `npm start` stalls through ide tools when linking reactist today so I haven't been able to test this in action yet. I'll give it another go later today but please help verify that this hasn't broken anything when updating in Twist.

---

When you're ready to release a new beta, please do the following:

- Update changelog.md
- Merge dev into beta
- Build storybook dist - the gh-pages hosted storybook is now running off of the beta branch
- Merge this into beta
- Cut a release tagging v9.2.0-beta.2

src/new-components/text-link/text-link.module.css
gnapse: I am precisely not wanting to assume an icon in a text link will be always on the left.

src/new-components/button-link/button-link.module.css
```diff
@@ -0,0 +1,9 @@
+.container {
+    gap: var(--reactist-spacing-medium);
```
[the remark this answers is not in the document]
gnapse: It's a grid and a [flex property](https://caniuse.com/flexbox-gap). In flex it is not supported in IE11 (not a problem for Reactist and Twist) but it is not yet supported in Safari, although it will be supported soon (it is already supported in Safari Technology Preview). I acknowledge that this is a problem, but I wanted to worry about that later.

src/new-components/base-field/base-field.tsx
```diff
@@ -0,0 +1,82 @@
...
+//
+// BaseField
+//
+
+type HtmlInputProps<T extends HTMLElement> = React.DetailedHTMLProps<
+    React.InputHTMLAttributes<T>,
+    T
+>
+
+interface BaseFieldChildren {
```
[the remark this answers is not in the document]
gnapse: I can stick to `type`. It started this way because Braid uses interfaces mostly, and early on I got inspired by their code. I honestly do not see it as a big deal if we intermix styles in different modules. Most resources out there even lean on "use `interface` until you need a feature that you can achieve only with `type`". I'll consider opening a thread about it later on.

gnapse:
> It's a bit hard to see these components in action without having stories in place

Agree. I have to make time to work on this, but right now I did not have as much. Also, I am worried about Natalie already being working in the Storybook migration, and I wondered if I should invest time in writing stories for an older version of storybook given that we're migrating to a newer version.

frankieyan:
I pulled this into Twist and found two warnings, looks like their classes haven't been defined:

image
image

These two should be the only ones - they get called a lot and flooded my console so it's a bit hard to tell.

gnapse:
@frankieyan my plan is to start developing locally on Twist some of the new settings screens using these components, and I'll polish and fix things on the Reactist side before going forward with merging and releasing this. I'll take a look at what you mention here.

gnapse:
According to semver we should be doing `9.2.0-beta2` instead of `9.2.0-beta.2`. See https://semver.org/spec/v1.0.0-beta.html#semantic-versioning-specification-semver

Also, I wonder why you have sometimes incremented the minor version when introducing new beta releases recently, instead of incrementing the number after the word "beta".

gnapse:
> my plan is to start developing locally on Twist some of the new settings screens using these components, and I'll polish and fix things on the Reactist side before going forward with merging and releasing this. I'll take a look at what you mention here.

**Update:** I am backtracking on this. Working locally with Reactist linked is not as smooth as I thought. Types are not picked up correctly from the linked reactist, and changes to stylesheets require restarting the `ide twist-web run` command. So I will release this as is. I followed the steps you provided.

frankieyan:
> According to semver we should be doing 9.2.0-beta2 instead of 9.2.0-beta.2

Ah feel free to omit the period in the next release 👍

> Also, I wonder why you have sometimes incremented the minor version when introducing new beta releases recently, instead of incrementing the number after the word "beta".

It's to keep the beta always at 1 minor version above the latest regular release. Do you think this is confusing and we should just keep it at one static version until we merge back to dev?

gnapse:
> It's to keep the beta always at 1 minor version above the latest regular release. Do you think this is confusing and we should just keep it at one static version until we merge back to dev?

It was confusing in the sense that it was not apparent to me what the reason was. But it makes sense in a way.

I wonder though if we should target v10 non beta as the one where we'll move all that is new out of beta and into the regular releases.

PR #497, opened 2021-04-15
fix: Simplify markup generated by Stack and Inline, fix ButtonLink styles

gnapse:
## Short description

1. Removes the extra element generated by both `Stack` and `Inline` around each child component it receives.
    - Originally brought up here: https://github.com/Doist/twist-web/pull/3367#discussion_r612986379
    - Relevant discussion: https://github.com/seek-oss/braid-design-system/issues/386
2. fix the ButtonLink styles
    - it was showing the text in the color of regular links and not the expected color according to its variant
    - it was also expanding all the way to the right, filling the available width

## PR Checklist

## Versioning

A new beta pre-release, most probably v9.2.0-beta.5


src/new-components/stack/stack.module.css
```diff
@@ -3,42 +3,42 @@
     margin-top: calc(var(--reactist-spacing-xsmall) * -1);
```
frankieyan, line 3: Just a question - why do we use a negative margin on the parent and then top margins/paddings in the children? Wouldn't it be easier to use bottom margins on all but the last child?

frankieyan (commented):
I pulled this into Twist and things look mostly good with the thread list/view where I've used `Stack`. 

Couple of places where we'll have to ensure `<li>`'s are used now when pulling this in, and some older components, such as the thread loading skeleton, sets/resets its own margins (even with just a single class name), which fails to have its margins overridden by the Stack, though I think that's to be expected and these cases should go away as we use more of the design system components.

The nesting on the settings field headers has reduced by 3 divs. It's still wrapped around 5 but it's a good improvement already.

frankieyan (approved):

frankieyan:
Another thought - should we start documenting with `README.md`'s in the components' folders as drafts before we kick off storybook demos? Either that or in the paper document - but I feel we should start jotting down design decisions / documenting use-cases for each specific component as we go rather than try to write up everything later on.

src/new-components/stack/stack.module.css
gnapse: We probably can. Also, it can still be top margin on all but the first child.

I went this way because on my early prototyping I did not manage something along those lines at first, and this worked. But it was probably a mishap of me not trying things in the right way, and since this worked first, I left it as is with the thought that this is internal implementation and that we could always revisit and improve it as long as we keep the behaviour.

gnapse:
> should we start documenting with README.md's in the components' folders as drafts before we kick off storybook demos?

Absolutely. The issue is time, at least for me. Having worked on these a bit "on the side" while working in all the settings UIs has on the one side saved me time because once I managed to have the components in place they've been good for not having to be writing a lot of custom repetitive CSS, but the time saved so far barely accounts for compensating for the time spent on them. But I still haven't had time for the rest (writing tests, updating storybook, etc.)

I've already discussed with Henning that we need dedicated time soon to give some love to all this, and we're on the same page that it will happen (in addition to my personal DO, which will be likely about it as well).

gnapse:
> It's still wrapped around 5 but it's a good improvement already.

This is a bit of an unfair comparison, fixating for some reason on the `form` as your point of reference. The `form` lies up above, even containing the upper header of the view (the header that does not scroll with the content of the right-hand area of the modal).

Under this `form` there are two divs that are needed: one that's the scrollable container with a constrained height, and another one that's immediately inside it that expands its height to fit its content, and is the basis of what scrolls inside the outer one. Those outer two divs have barely anything to do with the design system and a regular layout with custom css would've probably have the same structure.

There are really three other divs that are arguably what you are concerned about and that we could see if we can reduce that number. These are precisely the source of the 3-divs reduction we achieved here: 3 nested stacks that now have one less div each.

gnapse:
@frankieyan I needed to fix the ButtonLink (was showing the text in the color of regular links and not the expected color according to its variant, plus it was also expanding all the way to the right, filling the available width).

So I got the opportunity that I had not merge this to put those changes in the same release. Can you check it out? The new relevant changes are in the second commit (6be3538), and the changelog text in the 3rd commit.

frankieyan (approved):

PR #498, opened 2021-04-16
chore: Upgrade Storybook to v6.2.1

nats12:
## Short description

Upgraded Storybook to version 6.2.1 - there are [significant changes](https://storybook.js.org/blog/storybook-6-migration-guide/) with this update. I've kept to the current implementation as much as possible, focussing solely on upgrading and getting stories to work with the new version, however due to it being a major version release:

- stories have been rewritten to ensure they are all exported and their relevant prop types are displayed. What was was known as 'chapters' are now individual stories
- some stories have been converted into `.mdx` files due to their structure
- the config files are totally different
- certain addons are soon to be deprecated, namely `addon-knobs` which has been replaced in the solution by `addon-controls`. This now works in the same way as [Marketist's](https://doist.github.io/marketist/) controls
- storybook 6's out of the box design has changed from the previous version, so once again, it looks closer to what Marketist's Storybook looks like

## PR Checklist

## Versioning

As previously mentioned there has been some changes to the structure and design of the stories (not Reactist's components itself), which may require publishing a new version. I'm not entirely sure, however, so I'm keen to hear the reviewer's thoughts on this taking into account what has been done previously in this situation.


nats12:
@pedroalves0 unsure if you've seen that build error before but from when I was investigating the error locally, all solutions pointed to adding the `@babel/plugin-proposal-private-methods` plugin to the `.babelrc` file (which doesn't seem to work when building) or setting `@babel/preset-env` and `@babel/proposal-class-properties` to `loose` mode. Doing the latter could cause some [issues](https://developit.github.io/babel-legacy-docs//docs/advanced/loose/) down the line so not a fan of that approach. Let me know if you've come across this before and what may have been the cause, I'll continue to have a look

stories/components/Menu.stories.tsx
```diff
@@ -146,34 +144,22 @@ const items = [
...
-                as allowing to control the visible state of the menu from outside, and allow it to
-                be used as a context menu).
-            </p>
-            <ul>
-                {items.map((item, index) => (
-                    <Item key={index} {...item} />
-                ))}
-            </ul>
-        </section>
+        <>
```
pedroalves0, line 149: This fragment doesn't seem to be necessary.

pedroalves0 (commented):
Nice work 👍 The changes look good and from a quick look into the latest storybook docs they seem to comply with them. The stories run as expected locally.

- Even with https://github.com/Doist/reactist/pull/498/commits/cdf4fc0d6c82bd2dd878720fd9c394ffd2434c90 in place I (and apparently the CI too) still get an error when building:

  image


- Previously there was a single section with all stories, now there's a separate `mdx` section:

   image

   Can we opt-out of this?

>  I'm not entirely sure, however, so I'm keen to hear the reviewer's thoughts on this taking into account what has been done previously in this situation.

>  Let me know if you've come across this before and what may have been the cause, I'll continue to have a look

For both of these I'm pinging @gnapse and @frankieyan, as they've been working on reactist more closely than I have in recent times.

nats12:
> * Even with [cdf4fc0](https://github.com/Doist/reactist/commit/cdf4fc0d6c82bd2dd878720fd9c394ffd2434c90) in place I (and apparently the CI too) still get an error when building:
>   image

Setting their modes explicitly to false seems to have solved the problem. Still not entirely sure why it requires the plugin, it might be an idea to upgrade the babel plugins to a more recent version to see if that fixes it, although I'll investigate and create a new separate task for that as this one is quite large already.

nats12:
> * Previously there was a single section with all stories, now there's a separate `mdx` section:
>   image
>   Can we opt-out of this?

I spent some time thinking about how to structure this one seeing as it was quite different to the other stories. Notice [here](http://doist.github.io/reactist/?path=/story/modal--component-documentation) how it has a Component Documentation tab within the side bar with custom markdown content (when I say custom I mean content that isn't the components from the stories). In Storybook 6, there's a Docs tab which is made up of all of the story components associated within a story file. I couldn't find a way to keep all of the story components in this tab _as well as_ the markdown documentation from the existing site. The story `.tsx` files were no longer parsing markdown in the same way. Given it required both markdown and `.jsx`, I thought using it's MDX pages would work nicely, however it does separate it from the stories section. 

Should I look into this further to see if there's a way to do this? Perhaps by using a markdown parser and having the documentation content as its own story?

pedroalves0:
> Should I look into this further to see if there's a way to do this? Perhaps by using a markdown parser and having the documentation content as its own story?

My question was more about if there's some option to disable sectioning in storybook, if there isn't let's not make things too complex. We should stick with the most future-proof approach. Is it `mdx`? Will new components use it?

Btw, it got to my attention that stories are not co-located anymore, and file co-location is something that we've been striving for. Was this new story structure discussed within the team (perhaps I missed it)?

gnapse:
> Btw, it got to my attention that stories are not co-located anymore, and file co-location is something that we've been striving for. Was this new story structure discussed within the team (perhaps I missed it)?

I don't think they never were even before this PR. But we certainly want to move in this direction (the new components in the `beta` already use this convention).

@nats12 is it possible to have story modules co-located alongside the component, styles and test modules? I hope so.

nats12:
> 
> My question was more about if there's some option to disable sectioning in storybook, if there isn't let's not make things too complex. We should stick with the most future-proof approach. Is it `mdx`? Will new components use it?
> 

@pedroalves0 it turns out that this sectioning is all down to the `<Meta>` title you give your MDX file - I had mine under a `MDX/...` hierarchy hence the sectioning. I've now removed it and Modal appears alongside all other stories. Thanks for pointing this out!

As for the co-location, @gnapse that's right, they weren't co-located before this PR but it is possible. It's just a case of moving the files to sit alongside their components and then pointing to a different path in storybook's config file. As this PR was focussed on a version upgrade and already seemed a bit large, I thought co-locating could be tackled separately. But let me know if you feel it should also be done as a part of this same task 👍🏼

gnapse:
> As this PR was focussed on a version upgrade and already seemed a bit large, I thought co-locating could be tackled separately. But let me know if you feel it should also be done as a part of this same task 👍🏼

Not at all, I think it's fine to not tackle co-location since they were not co-located previously.

pedroalves0 (approved):

nats12:
@gnapse regarding versioning - should I bump the version to .beta-6 or keep as it is? I was unsure about this as, although it's an upgrade on Storybook and its stories and not Reactist's components, the stories canvasses etc have changed somewhat.

gnapse:
No. No beta. This is a change on top of the `dev` branch, so this go into a regular release. After it gets merged we'll need to bring beta up to speed (i.e. migrating the stories in that branch to the new storybook version). But that's another story.

gnapse:
TBH I'm a bit confused right now with having a prerelease versioning scheme vs. a regular one. Maybe @frankieyan can help us determine what's the next version number we should use in the regular releases numbering outside of beta.

frankieyan:
The last release off of dev was https://github.com/Doist/reactist/releases/tag/v9.1.2 so this could be either v9.1.3 or v9.2.0. Just a note that the [hosted docs](http://doist.github.io/reactist) won't be updated yet as we switched the target branch to beta. @nats12 I can take care of the changes in the beta branch after this one has been merged and released.

Having that said, we originally just wanted to maintain the beta branch for a couple of weeks but it's been quite a bit of extra work and the two parallel release numbers can cause confusion. I don't think it really gives us much value over working off of the main branch and properly documenting breaking changes (even if it may be frequent), especially when the webpack configs have diverged between the two branches. Ultimately, our goal here is just to work in the open rather than providing a good DX outside of Doist for now.

@gnapse if you don't have any objections, I can work to fold beta back into dev after the changes here are merged.

gnapse:
I agree it is a hassle, but in my mind we ideally should work on the new component's tests and/or at least their storybook. For instance, so many of the small fixes we've been doing recently ware because I ported newer components over but the ported versions were done in the dark and without ever running a page that used them, to even see them. This because we never wrote the new stories for them to not add to the workload that this piece of work here implied. As I've started to use new components in the new Twist settings, I've noticed mostly style issues that needed fix. And I anticipate more.

So I would vote for keeping the beta branch. My take is that we can freeze development for now in the non-beta branch. This piece of work here was an exception.

However, I'm open to not follow what I just said if you and/or others feel differently.

PR #500, opened 2021-04-20
Various fixes to some of the new design system components

gnapse:
## Short description

List of fixes:

- Add explicit link styles to `TextLink`
- `SwitchField` now properly animates the handle moving left or right.
- Style fixes to the `PasswordField`.
- Added ref forwarding to `TextField`, `PasswordField` and `Heading`.

## PR Checklist

## Versioning

New pre-release, probably v9.2.0-beta.6.


src/new-components/password-field/password-field.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react'
+import { useForkRef } from 'reakit-utils'
```
frankieyan, line 2: The implementation of `useForkRef` looks almost identical to `useCombinedRef` we have within TD/TW. The project's [readme](https://github.com/reakit/reakit/blob/master/packages/reakit-utils/README.md) also warns that there may be breaking changes in minor releases, so to better prevent breakage we can either:

- Explicitly pin the current version of `reakit-utils` to package.json. I noticed right now we're using whichever version comes with reakit
- Port over `useCombinedRefs` as a utility file here in Reactist (and to replace the copy in TD/TW)

frankieyan (approved):
Approved to not block you, but I think something should be done about `reakit-utils` 👍

src/new-components/password-field/password-field.tsx
gnapse: I checked and we're already pinning down the reakit version to a specific version, which should make the reakit-utils stable, right?

I do want to host some of our reusable hooks in Reactist, but that's out of the scope for now.

PR #501, opened 2021-04-21
feat: Forward ref from Heading component

frankieyan:
## Short description

Allows the top-level element's ref to be forwarded from the `Heading` component

## PR Checklist

## Test plan

Pull this into Twist with `npm link` (ide-tools won't symlink locally AFAIK) and run the test suite. There should be no test failures. 

## Versioning

v9.2.0-beta.6

### DIFF
--- a/src/new-components/heading/heading.tsx
+++ b/src/new-components/heading/heading.tsx
@@ -1,8 +1,10 @@
 import * as React from 'react'
 import { getClassNames } from '../responsive-props'
 import { Box } from '../box'
-import styles from './heading.module.css'
+import { forwardRefWithAs } from '../type-helpers'
 import type { Tone } from '../common-types'
+
+import styles from './heading.module.css'
 
 type HeadingLevel = 1 | 2 | 3 | 4 | 5 | 6 | '1' | '2' | '3' | '4' | '5' | '6'
 type HeadingElement = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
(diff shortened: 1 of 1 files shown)


src/new-components/heading/heading.tsx
```diff
@@ -21,15 +23,10 @@ type HeadingProps = SupportedHeadingElementProps & {
...
-function Heading({
-    level,
-    weight = 'regular',
-    size,
-    tone = 'normal',
-    children,
-    lineClamp,
-    ...props
-}: HeadingProps) {
+const Heading = forwardRefWithAs<HeadingProps>(function Heading(
```
gnapse, line 26: We can and should do it with a regular `React.forwardRef` call, as this component does not support allowing the caller to pass in a `component` prop and change the element it renders with.

gnapse (changes requested):
Looks good, but I think we can simplify it a bit. See my comment below.

gnapse:
@frankieyan I have #500 open, and I was wondering if we can make a single new release where we include this alongside those changes as well. Let me know if/how we can coordinate this.

src/new-components/heading/heading.tsx
frankieyan: Makes sense 👍

frankieyan:
👍 Let's see if we can wrap up the two PRs around the same time. I'm seeing a bit of issues with ide-tools hanging when used with `--link-reactist` so definitely would like to get it in quickly. I should be able to get the changes here over to you before you start your day tomorrow.

gnapse:
@frankieyan it turns out I needed to do the same for `TextField` and `PasswordField`, so I added those changes in #500, and I included the change for the `Heading` component as well. I set you as reviewer, so if you approve that PR it supersedes this one.

Also, if you approve and you really need this, feel free to do the release in that branch. Otherwise I can work on it when I start my work day this Thursday.

frankieyan:
Closed in favour of #500

PR #509, opened 2021-04-27
Make disabled button component more accessible through aria-disabled

nats12:
## Short description

Add `aria-disabled` to the `Button` component to ensure its state is read by screen readers.

## Motivation / longer description 
The current implementation using `disabled` is ignored by screen readers and so I wanted to experiment a bit with this component to make it more accessible. By adding `aria-disabled`, simply changing the CSS selector which already applied opacity anyway, and conditionally blocking `onClick` events, the button is now picked up by screen readers whilst it maintains its disabled state look. 

The change requires no changes to any components making use of the button since it still expects a `disabled` prop. The only difference now is that the prop is only used to determine whether `onClick` should fire or `aria-disabled` should be true or false.

## PR Checklist

- Run storybook 
-   [ ] Inspect any of the buttons within the `Disabled Buttons Story` and make sure that `aria-disabled` is applied and `disabled` is no longer present
-   [ ] Ensure that the button is styled the same as the current disabled one in production and that pointer events are disabled
- To check for onClick events:
    - The best and probably easiest way is to create a dummy function in `button.tsx` where you console.log something out
    - Pass that function to the button's `onClick` attribute and make sure nothing is printed out to the console when you click on the disabled button
    - Set `disabled` to `!disabled` in the `onClick` - ensure something is printed to the console (this is a check to ensure events are fired when disabled is false)

As I mentioned, this is something I wanted to experiment with and I'm open to any feedback  or changes to it!

### DIFF
--- a/src/components/button/button.less
+++ b/src/components/button/button.less
@@ -8,7 +8,7 @@
     background-color: transparent;
     padding: 0;
 
-    &:disabled {
+    &[aria-disabled='true'] {
         opacity: 0.4;
         cursor: not-allowed;
     }
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -64,8 +64,8 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(function Button(
             ref={ref}
             type={type}
             className={className}
-            disabled={disabled || loading}
-            onClick={onClick}
+            aria-disabled={disabled || loading}
+            onClick={disabled || loading ? undefined : onClick}
         >
             {children}
         </button>
--- a/src/components/dropdown/__snapshots__/dropdown.test.tsx.snap
+++ b/src/components/dropdown/__snapshots__/dropdown.test.tsx.snap
@@ -102,8 +102,8 @@ exports[`Dropdown Dropdown.Box renders the Body component first when top prop is
         size="default"
       >
         <button
+          aria-disabled={false}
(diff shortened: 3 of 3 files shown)


gnapse:
Wow. Thanks. I did not know that screen readers do not acknowledge the `disable` attribute. Do you have a source for that? (not because I doubt it, but to know more)

As for the change: I'd suggest we add both attributes instead of trying to mimic in js what the `disabled` one does. Unless you did not go down that path for a reason that I could be missing.

nats12:
I was reading up a bit on the topic and found: 

> By using the tabulator key and navigating within a form the screen readers will ignore the existence of any disabled button.

Taken from [here](https://a11y-101.com/development/aria-disabled). The problem is that as soon as you add the `disabled` attribute, it's no longer focusable and so it gets completely ignored by the reader. I tried this out with a screen reader and saw this was the case when tabbing through. Whereas when trying with `aria-disabled`, the reader will read out `button disabled` at the end.

> As for the change: I'd suggest we add both attributes instead of trying to mimic in js what the disabled one does. Unless you did not go down that path for a reason that I could be missing.

Just tried this to see what would happen with both and it seems like `disabled` overrides `aria-disabled` as it goes back to skipping the entire section in which the button is found in.

gnapse (approved):
Oh well, right. I agree with you that not fully disabling it is better. Thanks for the push back.

PR #514, opened 2021-05-06
fix: Make Columns component grow in a flex container

frankieyan:
## Short description

As a follow up to https://github.com/Doist/twist-web/pull/3422#discussion_r623995188, this makes the `Columns` component stretch when used inside of a flex container:

image

There is currently no way to override this from the Columns component's API. Do we see any use cases where we wouldn't want this to happen?

## PR Checklist

## Versioning

v9.2.0-beta.10


frankieyan:
cc. @gnapse

nats12 (approved):
Code looks good and have tested manually 👍🏼

> Do we see any use cases where we wouldn't want this to happen?

I can't think of any

gnapse:
I originally proposed to do this, and I think it makes sense. I see no reason for it not to be this way. But of course, actual use will be a better judge of this. Let's see how it goes.

gnapse:
@frankieyan (cc @nats12 as you were also involved in the conversation): I found an instance of this change here not being desirable. This change assumes that the `Columns` component is inside a flex container that is set to `flex-direction: row`. In these circumstances, we do want the columns to stretch, because they stretch horizontally. But if it is inside a `flex-direction: column` container, we do not want the columns to stretch vertically.

I'm reverting this change, and to allow us to have control over it, I've added the `flexGrow` and `flexShrink` props to all layout components so they can forward it to the `Box` they render. I think it makes sense to give users of these components to instruct them how to behave when the parent is a flex container.

Let me know what you think.

frankieyan:
@gnapse Yup exposing these as props sounds good to me 👍

PR #516, opened 2021-05-06
chore: Update @doist/eslint-config to 6.x

proxi:
Closes #...

## Short description

Update @doist/eslint-config to 6.x. Only very minor internal changes.

## PR Checklist


src/components/keyboard-shortcut/keyboard-shortcut.tsx
```diff
@@ -118,7 +118,7 @@ function KeyboardShortcut({
     children,
     className,
     translateKey = globalTranslateKey,
-    isMac = navigator.platform != null && navigator.platform.toUpperCase().includes('MAC'),
+    isMac = navigator.platform?.toUpperCase().includes('MAC'),
```
gnapse, line 121: This expression is not guaranteed to be boolean. It could be `undefined`. It's a bit of a nitpick because functionally it will all still work as expected, but I'd rather do this:

```suggestion
    isMac = Boolean(navigator.platform?.toUpperCase().includes('MAC')),
```

gnapse (approved):

src/components/keyboard-shortcut/keyboard-shortcut.tsx
gnapse: > `isMac?: boolean` so the expression does not have to be Boolean, IMO version without conversion is cleaner.

IMO optional boolean props are only optional for the sake of making it possible to consumers of the component to omit passing it explicitly. But the component should be explicit about what boolean value it gives to it when omitted. But anyway, this is a nitpick of mine.

> Can navigator.platform even be undefined? TypeScript does not think so.

Probably not, but in looking at the documentation about it, it does not seem safe that we're relying on it:

> This feature is no longer recommended. Though some browsers might still support it, it may have already been removed from the relevant web standards, may be in the process of being dropped, or may only be kept for compatibility purposes. Avoid using it, and update existing code if possible

From https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform

proxi: How do you like `?? false` instead?

gnapse: That works too.

PR #517, opened 2021-05-07
A couple of style fixes

gnapse:
Works towards closing https://github.com/Doist/Issues/issues/2709 (it does not actually close it until we upgrade Twist to use the new version of Reactist that we'll release with these changes).

## Short description

I messed up in #497 and added some `!important` rules that right now I'm not even sure why they were needed because removing them does not affect anything in the new settings in Twist (presumably the reason I did this was related to the new settings UI).

Anyway, here I'm reverting that. I also sneaked in an unrelated style fix so that the new `SelectField` components have the correct text color in dark mode in Twist.

## Test plan

- Set Reactist locally to this branch, run `npm start` in the terminal in Reactis's folder, and then run Twist locally while linked to use the local Reactist instead of the one from npm.
    - [ ] Then make sure that buttons all around Twist in dark mode appear with a light text color on a dark background, instead of dark text on dark background.
- With the new twist and new settings feature flags on, visit the new settings while in dark mode.
    - [ ] Make sure that the `select` elements text color is also light-colored text on dark background.

## PR Checklist

-   [x] N/A ~~Added tests for bugs / new features~~
-   [x] N/A ~~Updated docs (storybooks, readme)~~

## Versioning

A new beta release.

### DIFF
--- a/src/new-components/select-field/select-field.module.css
+++ b/src/new-components/select-field/select-field.module.css
@@ -33,6 +33,7 @@
     appearance: none;
     box-sizing: border-box;
     width: 100%;
+    color: var(--reactis-content-primary);
     background: none;
     border-radius: var(--reactist-border-radius-small);
     border: 1px solid var(--reactist-framework-border);
--- a/src/components/button/button.less
+++ b/src/components/button/button.less
@@ -35,7 +35,7 @@
 
     &--primary {
         background-color: @primary_color;
-        color: white !important;
+        color: white;
 
         &:not([disabled]):hover {
             background-color: @primary_color_darker;
@@ -44,7 +44,7 @@
 
     &--danger {
         background-color: @color_red;
-        color: white !important;
+        color: white;
 
         &:not([disabled]):hover {
             background-color: darken(@color_red, 10%);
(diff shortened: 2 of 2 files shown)


src/new-components/select-field/select-field.module.css
```diff
@@ -33,6 +33,7 @@
     appearance: none;
     box-sizing: border-box;
     width: 100%;
+    color: var(--reactis-content-primary);
```
rfgamaral, line 36:
```suggestion
    color: var(--reactist-content-primary);
```

rfgamaral (changes requested):
I took a quick look at this one, and the buttons and select seem to be fixed, however, I noticed this:

image

frankieyan:
Thanks for taking a look at this one @rfgamaral 🙌

src/new-components/select-field/select-field.module.css
gnapse: This reminds me of re-considering using https://github.com/seek-oss/vanilla-extract internally. With it you get static analysis of css variables, builds fail if you use invalid css variable names, just as it happens for JS today.

I probably will open a thread about it.

gnapse:
@rfgamaral I noticed that too, later in the day, after having pushed this. For some reason `option` elements seem to inherit `color` but not `background` or `background-color`. This is unfortunate. Anyway, I added not an explicit background colour to them, and it should work as expected now, both in light and in dark mode.

rfgamaral (approved):
Looks great now, ship it 🚀

PR #520, opened 2021-05-11
refactor: Migrate modal story to pure MDX story format

nats12:
## Short description

Converted the modal story into pure MDX format as we start to move towards MDX documented components.

Ref: https://twist.com/a/1585/ch/190200/t/2455247/

This involved extracting all functions from the helper file and placing them into the MDX file alongside the stories. I've tried to reuse as much code as possible here, defining templates where possible, but please feel free to suggest any other ways in which I could make these more reusable.

## PR Checklist

- [ ] Make sure the code makes sense
- [ ] Make sure that all stories under Modal look and work the same way as before when clicking on their trigger buttons [see deployed version](http://doist.github.io/reactist/?path=/story/components-modal--modal-header-only-story) for reference 
- [ ] Make sure all the controls under the Modal Playground Story work as before

## Versioning

v9.1.5 will be released to include this story


frankieyan (commented):
This is looking really good 👍

Random thoughts, but they don't need to be a part of this PR:
- I wonder if there is value in having a separate template .mdx file, because the modal here is a bit of a different use case where it's made up of several sub-components, so our ArgsTable is currently a bit different. It'd be good to have an example where our props are inferred from the components themselves
- Also if we have a separate template, should we define some placeholders for information we should fill out? ie. component description, use cases, anything else we can think of? For reference, this is the fullstack team's storybook: https://doist.github.io/marketist/?path=/docs/typography-text--standard
- How do we add the description and default fields to the args table? Is it through jsdocs, or a `defaultProps` property on the component?

stories/components/Modal.stories.mdx
```diff
@@ -1,22 +1,36 @@
...
 
 # Component Documentation
 ### How to use
 ---
 
-The **Modal** component consist of four subcomponents *Box, Header, Body, Actions* which are exported as keys of the Modal object. Hence, to use a Modal your code should look like this:
+The **Modal** component consists of four subcomponents *Box, Header, Body, Actions* which are exported as keys of the Modal object. Hence, to use a Modal your code should look like this:
 ```
 import { Modal } from 'reactist';
 ...
```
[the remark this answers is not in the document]
nats12: oops! fixed in latest commit 👍🏼

stories/components/Modal.stories.mdx
```diff
@@ -1,22 +1,36 @@
...
-    ModalScrollableBodyStory,
-    PlainMediumModalStory,
-    ModalPlaygroundStory } from './ModalStoryHelper'
+import Button from '../../src/components/button'
+import './styles/modal_story.less'
+
 
-<Meta title="Modal" component={Modal} />
+<Meta 
+  title="Modal"
```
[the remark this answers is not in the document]
nats12: Good point, also done in latest commit

nats12:
_**TL;DR**: **a)** I think you're right and Modal might not be the best story to use as reference and **b)** getting prop types inferred for this component is proving to be tricky even without the MDX implementation._


> I wonder if there is value in having a separate template .mdx file, because the modal here is a bit of a different use case where it's made up of several sub-components, so our ArgsTable is currently a bit different. It'd be good to have an example where our props are inferred from the components themselves

So I've been trying out different implementations to get prop types inferred with `Modal` but can't get it to work. Removing the MDX file and going for the standard DocsPage implementation doesn't seem to generate the table either (I've tried applying the `defaultProps` property as well). So I think you may be right in that this is a special case which may not be the best to use as a reference, but we also need to keep in mind that we may not be able to show a props table for this with either approach.

> Also if we have a separate template, should we define some placeholders for information we should fill out? ie. component description, use cases, anything else we can think of? For reference, this is the fullstack team's storybook: doist.github.io/marketist/?path=/docs/typography-text--standard

If this is regarding `Modal` this would be a good idea although it might overcrowd the template. I've added different sections for helpers, template binds etc - did you mean extra comments for things like where descriptions could go within each template bind?

> How do we add the description and default fields to the args table? Is it through jsdocs, or a defaultProps property on the component?

The descriptions get added to the table automatically when the props are inferred from the components, which sadly isn't the case in `Modal`'s case. We can define them manually, however the [docs](https://storybook.js.org/docs/react/writing-docs/doc-blocks#customizing) specify that this customisation is only available using the `<ArgsTable story=<name of story> />` format instead of the `<ArgsTable of={<component>}> />` format we are after. So the prop would have to go on each story looking similar to what you tried before:

```
<ArgsTable story={"ModalHeaderOnlyStory"} />
.... 

<Story
    ....
    argTypes={{
        title: {
             description: 'some description'
         }
     }}
....
</Story>
```
The problem with this is that we would have to have a table for each story and the same props often show up on each story.

stories/components/Modal.stories.mdx
```diff
@@ -63,54 +77,230 @@ The **Modal.Actions** displays an action bar on the bottom of the Modal. It work
...
-<Canvas>
-  <Story name="ModalScrollableBodyStory">
-    <ModalScrollableBodyStory />
+<!-- Prop type tables -->
+<!-- 
+Modal story is made up of several components, each with their own arguments,
+so <ArgsTable of={Modal} /> wouldn't work in this case. Instead, we specify the
+props of one story i.e. Playground
+-->
+<ArgsTable story="ModalPlaygroundStory" />
```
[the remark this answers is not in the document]
nats12: I've tried doing this as well thinking each of Modal's subcomponent types would get picked up but they don't. I've even tried importing `Box` directly and setting that as a subcomponent as well as ArgsTable's `of` value

frankieyan:
In that case the changes here for the modal looks good to me 👍 we can investigate issues with the props table separately - it would be awesome if descriptions and default values can be inferred, or if at least there is an easy way to define them.

For now, what do you think about bringing the playground story closer to the props table (as the first story)?

Another thing is that it looks like the stories themselves can have a name defined, rather than using the exports' names: https://storybook.js.org/docs/react/writing-stories/introduction#rename-stories. Should we do this here (with just the modal to start?)

frankieyan (approved):

nats12:
> Another thing is that it looks like the stories themselves can have a name defined, rather than using the exports' names: storybook.js.org/docs/react/writing-stories/introduction#rename-stories. Should we do this here (with just the modal to start?)

Yes this is a good idea, we are already setting the story's name property anyway so why not give them better names.

> For now, what do you think about bringing the playground story closer to the props table (as the first story)?

@frankieyan This would make sense as it could get confusing toggling the controls and not seeing the changes being reflected in any of the docs tab stories. Something to keep in mind with this though is that it would mean Playground would appear first under Modal within the sidebar 

image

stories/components/Modal.stories.mdx
```diff
@@ -49,68 +69,261 @@ Nevertheless, the **Modal.Header** will look the same whether you go for option
...
+        }}
+    >
+        {ModalTemplate.bind({})}
+    </Story>
+</Preview>
+
+<Preview withToolbar>
+    <Story
+        parameters={{ docs: { source: { type: 'code' } } }}
+        name="Modal Header And Body"
```
frankieyan, line 277: Super nitpicker, but I wonder if sentence case would look better in the sidebar

stories/components/Modal.stories.mdx
```diff
@@ -49,68 +69,261 @@ Nevertheless, the **Modal.Header** will look the same whether you go for option
...
+</Preview>
+
+<Preview withToolbar>
+    <Story
+        parameters={{ docs: { source: { type: 'code' } } }}
+        name="Modal Header"
+        args={{
+            title: 'Header of Modal',
+            buttonTitle: 'Header Only',
+            buttonText: 'Click me to launch a Modal with Header',
```
frankieyan, line 267: I don't think it was like this the first time I tested the stories, but the non-playground examples will now create a modal in the playground area:

image

stories/components/Modal.stories.mdx
```diff
@@ -49,68 +69,261 @@ Nevertheless, the **Modal.Header** will look the same whether you go for option
...
+
+<Preview withToolbar>
+    <Story
+        parameters={{ docs: { source: { type: 'code' } } }}
+        name="Plain Medium Modal"
+        args={{
+            body: PlainMediumText,
+            title: 'Header of Modal',
+            buttonTitle: 'Header, Body and Actions',
+            buttonText: 'Click me to launch a medium-sized Modal with Header and plain Body',
```
frankieyan, line 322: Not a knock on the work here, and not something we need to fix right now, but it's a bit weird that the "medium" sized modal at 680px can be larger than a "large"-sized one (60%) (the top one is medium):

image

stories/components/Modal.stories.mdx
nats12: Great spot - this was due to the playground story template including an additional div with id `modal_box` which is what the helper uses to render stories. Will be fixed in next commit

stories/components/Modal.stories.mdx
nats12: Tried this out and it does look nicer, it might even make more sense given they're not components just variations of one

stories/components/Modal.stories.mdx
nats12: This is odd - I had a look and the media breakpoint here seems to kick in.

image

When applying `width: calc(~'100% - 80px');` as the width in `&.large` instead of 60%, it works as expected. As this seems to be an issue directly relating to the component as a whole,  I'll raise an issue so that it's tackled separately

nats12: Issue raised here: https://github.com/Doist/reactist/issues/521

PR #525, opened 2021-06-04
chore: Change all `dev` branch references to `main`

rfgamaral:
### Overview

An attempt to change all `dev` references I could find to `main`. The more critical ones are scripts and CI related files (GitHub Workflows / Azure Pipelines). The less critical ones, documentation files.

Please review carefully, and let me know if you think I've missed some references.

### Test plan

If all pipelines and statuses checks are green, it should be good to go. But the real test will be after this is merged.

### DIFF
--- a/.github/ISSUE_TEMPLATE/Feature.md
+++ b/.github/ISSUE_TEMPLATE/Feature.md
@@ -21,4 +21,4 @@ Provide an example for how this feature would be used.
 
 ### Possible implementations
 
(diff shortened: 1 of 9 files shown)


frankieyan (approved):
The changes look good to me 👍

There is one reference in Reactist.stories.mdx (the first story that appears in storybook) that points to a `development` branch to a dist folder that is no longer kept in source control:

https://github.com/Doist/reactist/blob/7205605364b74b32a1f9428f0399a5945d24f506/stories/reactist/Reactist.stories.mdx#L11

We can change this to static references to 
`/node_modules/@doist/reactist/dist/reactist.cjs.production.min.js`
`/node_modules/@doist/reactist/styles/reactist.css`

PR #528, opened 2021-06-10
fix: Prevent emojis from being cut off in Text component when lineclamp is enabled

frankieyan:
## Short description

When `lineClamp` is enabled for the `<Text>` component, then certain emojis can be cut off.

Not sure if this is the best solution as it can cause an extra character to be truncated, but this should at least prevent emojis from being cut off.

## Test plan

- In the `Text` story, change the `lineClamp` demos to use a text string that ends with an emoji, such as `Here we go again 😊`
- [ ] Shrink the viewport. The emojis should no longer be cut off and should instead be truncated right away if it doesn't fit

|Before|After|
|-|-|
|image|image|

## PR Checklist


src/new-components/text/text.tsx
```diff
@@ -43,6 +43,7 @@ const Text = forwardRefWithAs<TextProps>(function Text(
                 lineClampMultipleLines ? styles.lineClamp : null,
                 lineClamp ? getClassNames(styles, 'line-clamp', lineClamp.toString()) : null,
             ]}
+            paddingRight={lineClamp ? 'xsmall' : undefined}
```
gnapse, line 46: I would add a comment above this line with a link to this issue to make it clear to future readers why this prop  is added:

```suggestion
            // see https://github.com/Doist/reactist/pull/528
            paddingRight={lineClamp ? 'xsmall' : undefined}
```

gnapse (approved):

PR #530, opened 2021-07-14
feat: Add `lineHeight` prop to `<Text>`

frankieyan:
## Short description

This adds a "default" line-height to the Text component's various sizes. While there is only one value available for each size, it's not applied by default unless the `lineHeight="paragraph"` prop is provided so as to not introduce a breaking change and affect existing layouts. 

Line height values are from [Figma](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product---Web?node-id=239%3A26&viewport=579%2C485%2C1) and also as requested by Alex at https://github.com/Doist/todoist-web/pull/3213

A similar change is added to Todoist's implementation without introducing the new `size` values at [`2729a83` (#3213)](https://github.com/Doist/todoist-web/pull/3213/commits/2729a83f4c74756e505b6ec3f99e70aad2f3506a)

## PR Checklist

## Versioning

Next beta version

### DIFF
--- a/src/new-components/default-styles.less
+++ b/src/new-components/default-styles.less
@@ -36,6 +36,12 @@
     --reactist-font-weight-medium: 500;
     --reactist-font-weight-strong: 700;
 
+    /* line heights */
+    --reactist-line-height-caption: 16px;
+    --reactist-line-height-copy: 18px;
+    --reactist-line-height-body: 21px;
+    --reactist-line-height-subtitle: 20px;
+
     /* border and separator colors */
     --reactist-framework-separator: rgb(230, 230, 230);
     --reactist-framework-border: rgb(214, 214, 214);
--- a/src/new-components/text/text.module.css
+++ b/src/new-components/text/text.module.css
@@ -31,6 +31,22 @@
     color: var(--reactist-chromatic-content-red);
 }
 
+.lineHeight-caption-paragraph {
+    line-height: var(--reactist-line-height-caption);
+}
+
+.lineHeight-copy-paragraph {
+    line-height: var(--reactist-line-height-copy);
+}
+
+.lineHeight-body-paragraph {
+    line-height: var(--reactist-line-height-body);
+}
+
+.lineHeight-subtitle-paragraph {
+    line-height: var(--reactist-line-height-subtitle);
+}
+
 /* truncated text */
 
 .line-clamp-1 {
--- a/src/new-components/text/text.stories.tsx
+++ b/src/new-components/text/text.stories.tsx
@@ -24,7 +24,7 @@ export default {
 export function TextStory() {
     return (
         <section className="story">
-            <Stack space="medium">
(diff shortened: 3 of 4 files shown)


frankieyan:
Tagging @gnapse as well so you're aware 👍

gnapse:
> While there is only one value available for each size, it's not applied by default unless the lineHeight="paragraph" prop is provided so as to not introduce a breaking change and affect existing layouts.

But wait, should all texts have these line heights? We should not introduce props whose only purpose is to not introduce breaking changes. Props and variations of our components should reflect what the design system needs and expects. Do we expect to need both instances of text elements? That is, instances that should have these new line heights and instances that should not have these new line heights?

frankieyan:
You may be right. While I wanted to distinguish between single-line text and paragraph text and not introduce changes everywhere, they are more subtle than I thought. They do still shift things down a pixel or two here and there, and is especially noticeable for single-line texts, but it feels like things are less cramped overall:

out
(Changes are from [`2729a83` (#3213)](https://github.com/Doist/todoist-web/pull/3213/commits/2729a83f4c74756e505b6ec3f99e70aad2f3506a))

In Twist it's a bit different. The modal cascades down a 1.7 line-height and this actually tightens up the lines more:

twist

In the Twist thread views, you can notice a bigger gap between the user names and the comment content, and the thread titles vs snippets. The nav items also look like they're a little bit off even though they are flex center aligned. Most of these previously had the browser default of `line-height: normal` which is [around 1.2](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values):

twist-thread-list

twist-thread-view

Anyway, I imagine we'll need to run through this again in the future once the new text stack is finalized. @alexmuench can we have your opinion here as well? Do you think the line heights from the [Product library](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product---Web?node-id=239%3A26&viewport=579%2C485%2C1) can be added to all Text components app-wide, or would it make sense to divide them into single-line vs paragraph use cases?

alexmuench:
@frankieyan I asked in the [Design thread about text styles](https://twist.com/a/1585/ch/402739/t/2592276/c/68915033). I can't answer it at this moment, but we'll get back to you.

gnapse:
Frankie, those before/after animations are really cool!

pedroalves0:
@frankieyan, I see that you folks are still discussion some details, let me know if/when I should review.

alexmuench:
Hi, it looks like our current typography styles in the library are outdated and we will need some time to port the new ones over from this [Paper doc](https://paper.dropbox.com/doc/New-Text-Styles-Proposal--BOkh6PcsxxfYVAyqwvON4heGAg-GDsNYe0uMcQSEw2U4CUi2#:uid=152297400404227920892190&h2=Text-Styles) and do more internal testing inside the design team. For now, can we postpone this activity?

frankieyan:
Ok I'll close this and apply an override for the education modal instead 👍

PR #531, opened 2021-07-26
fix: Allows classname to be passed through on ButtonLink

scottlovegrove:
## Short description

`ButtonLink` allows you to set your own `className` but then does nothing with it. It gets "used" as part of the `...props` usage, but then gets overridden a couple of lines later.

## PR Checklist

## Versioning

This is a small bug fix.


src/new-components/button-link/button-link.tsx
```diff
@@ -14,6 +14,7 @@ const ButtonLink = forwardRefWithAs<ButtonLinkProps>(function ButtonLink(
     { component = 'a', variant, size = 'default', openInNewTab = false, ...props },
     ref,
 ) {
+    const { className } = props
```
henningmu, line 17: Could be extracted right in the function signature:
```suggestion
    { component = 'a', variant, size = 'default', openInNewTab = false, className, ...props },
    ref,
) {
```

_not sure if this passes prettier_

henningmu (approved):
Left a stylistic comment, feel free to merge and release afterwards 👍

src/new-components/button-link/button-link.tsx
scottlovegrove: Yep, you're right 🤦🏻‍♂️ Have updated this now.

gnapse:
Actually, this was sort-of intended. My vision for the design system is that we should not need to pass class names to our components. If we do, it is a sign that there's something missing from the design system.

I have been thinking of allowing an escape hatch, and allowing passing a class name via an unconventionally named prop (e.g. `customClassName="…"`) more or less in the same vein as `dangerouslySetInnerHTML`. This would allow the occasional escape mechanism for when the component does **not yet** conform to what's expected of it from the design system, while at the same time make it easier for us to find in the codebase all the cases that are needing to use this escape hatch and why (stress in "not yet" because the idea is that the design system components should fulfill all their possible visual alternatives without custom css).

@scottlovegrove, can you share here the use case for which you need to customize the styles of the `ButtonLink`?

scottlovegrove:
@gnapse In my case, I needed to be able to anchor the button to the bottom of its parent container:
image

I could see no way of doing this with the current implementation of the components (please correct me if i'm wrong).

gnapse:
Hmmm, that's a layout need, which stresses my instinct that we should not customize the button styles. I thought you needed to customize the button look and feel (e.g. its color, border, text styling, etc.)

It looks like what you needed was to have a flex container with a `flex-direction: column`, and have the first child grow to fill all the available height. Can you point me via DM to the code where this layout is present? I can take a look and see if we can get rid of the need for the class name prop for this case.

scottlovegrove:
Will do 👍🏻

PR #532, opened 2021-08-04
fix: Add prop table for modal story

nats12:
## Short description

During the upgrade to Storybook 6 and the shift to MDX stories, we noticed that the prop types for the `Modal` story weren't inferred due to its structure (a component with subcomponents with their own props and types as its type). 
After a lot of researching online and creating a discussion in [Storybook's repo](https://github.com/storybookjs/storybook/discussions/15573), we have yet to find a way to automatically infer prop types of this component. Two workarounds were: **a)** inferring prop types from one of the actual stories i.e. Playground and not the component or **b)** manually set these props and their descriptions (the former wouldn't handle this). In this PR, I've attempted option **b.** 

Granted it may not be the best solution, as any changes made to the `Modal` component will mean inconsistent prop types between itself and the story. But I'm struggling to see how we can get these inferred without completely changing the structure of our `Modal` component so that it's no longer a component made up of [types](https://github.com/Doist/reactist/blob/main/src/components/modal/modal.tsx#L228).

This PR can serve as a bit of an exploratory area if we're not happy with this approach. Feel free to suggest alternative ways to achieve this or even contribute to this branch.

## PR Checklist

- Run storybook 
- Visit the modal story and click on it's Docs tab
	- [ ] Make sure you see a prop table with all the same Modal props we see under the Control tab
	- [ ] Make sure all props under this table have descriptions
	- [ ] Make sure that changing the prop's controls changes the Playground story.

### DIFF
--- a/stories/components/Modal.stories.mdx
+++ b/stories/components/Modal.stories.mdx
@@ -151,47 +151,67 @@ The **Modal.Actions** displays an action bar on the bottom of the Modal. It work
 
 <!-- Helpers -->
 
-export const renderModal = (modal) => {
+export function renderModal(modal) {
     ReactDOM.render(modal, document.getElementById('modal_box'))
 }
-export const getStory = (text, modal, title) => (
-    <section className="story">
-        <p>{title}</p>
-        <div id="modal_box" />
-        <Button variant="primary" onClick={() => renderModal(modal)}>
-            {text}
-        </Button>
-    </section>
-)
+
+export function getStory(text, modal, title) {
+    return (
+        <section className="story">
+            <p>{title}</p>
+            <div id="modal_box" />
+            <Button variant="primary" onClick={() => renderModal(modal)}>
+                {text}
+            </Button>
+        </section>
+    )
+}
 
 <!-- Templates -->
 
-export const WithPlaygroundTemplate = ({ ...args }) => (
-    <section>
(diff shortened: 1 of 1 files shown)


stories/components/Modal.stories.mdx
```diff
@@ -117,25 +179,16 @@ export const WithPlaygroundTemplate = ({ ...args }) => (
...
-    body = '',
-    title,
-    subtitle = '',
-    buttonTitle,
-    buttonText,
-    medium = false,
-    plain = false,
-    large = false,
-}) => {
+export const ModalTemplate = (args) => {
```
frankieyan, line 182: Can we destructure this and group each components' props together, rather than passing everything down into each one?

frankieyan: nit: Can we use function declarations for these since it's the convention in our projects?

frankieyan (approved):
Using the `Canvas` instead of `Preview` seems to allow the right source code to be displayed in each example as well. Awesome!

stories/components/Modal.stories.mdx
nats12: @frankieyan I've destructured props in the latest commit but not quite sure what you meant by grouping them together?

frankieyan: The change LGTM @nats12 👍

PR #533, opened 2021-08-11
Polymorphism types refactoring and className prop constraints

gnapse:
## Short description

This can be summarized in two changes (one commit for each):

- Rename the `component` prop to `as`.
- Replace in all the design system components the prop `className` with one named `exceptionallySetClassName`.

See [a comment in Twist where I explain a bit about this](https://twist.com/a/1585/ch/190200/t/2693346/c/69452729). Also, check out https://github.com/Doist/reactist/pull/531#issuecomment-891425295.

## PR Checklist

## Versioning

Since we're still releasing this in the beta branch, this marks a new regular beta release. Most likely, it will be released as `v9.2.0-beta.16`.


henningmu:
If you want to discourage it even further the prop could accept an object like `dangerouslySetInnerHTML` 😁
```jsx
<Stack exceptionallySetClassName={{ __className: "oops" }} />
```

henningmu (approved):

frankieyan (commented):
I tried this out but I'm not sure if it's working as expected or if I'm misunderstanding it. For example, the `<Stack>` has the default mode of `obfuscateClassName`, so it should disallow `className` as a prop and instead should accept `exceptionallySetClassName`, correct?

In the Stack story, I changed it to:
image

But it ends up with the `bar` class applied:
image

gnapse:
@frankieyan this is because in `Stack` the `{...props}` spread comes after the explicit `className` that stack passes to the `Box` it renders. I did not think about this because passing `className` to the `Stack` is a type error, but indeed, it may be wise to prevent this and make sure our internal class names override the one from `props`. Will fix that.

gnapse:
I'm pushing a fix for the class names override issue. I'll make sure to write a test for this to not regress in the future. But I'll leave tests for a future PR where I'll be adding tests all around.

gnapse:
> If you want to discourage it even further the prop could accept an object like `dangerouslySetInnerHTML` 😁
> 
> ```js
> <Stack exceptionallySetClassName={{ __className: "oops" }} />
> ```

@henningmu you got me thinking with this. I may do it.

gnapse:
Nah, I did not do it in the end. This is good enough to remind folks about this being exceptional, plus allowing us to search for the cases where it is being used to identify potential points of improvement.

I'll merge this onto the branch I'll be using to gather my work during this month.

PR #534, opened 2021-08-12
Add tests for the layout-related components

gnapse:
## Short description

Adds the first new set of tests for the new design system components. I left it short, focusing on the layout components mostly, to make it easier to digest. And also to validate the approach.

Granted, most of these tests are not giving us the full confidence we'd expect. It all works by validating that certain class names are being added to the components, but we could still break things when touching the CSS files. My aim is to eventually add either visual regression tests, or Cypress component tests. I experimented with the latter, but I was not able to make it work in the short term. These tests are still valuable as they validate the logic inside JS code that decides what class names to apply.

## PR Checklist

## Versioning

This is not going to be released yet. This is going to be merged in a PR where I'm gathering lots of tests.

### DIFF
--- a/src/new-components/box/box.test.tsx
+++ b/src/new-components/box/box.test.tsx
@@ -1,0 +1,151 @@
+import * as React from 'react'
+import { render, screen } from '@testing-library/react'
+import { Box, BoxProps } from './'
+
+describe('Box', () => {
+    it('renders its children as its content', () => {
+        render(
+            <Box data-testid="box">
+                Hello <strong>world</strong>!
+            </Box>,
+        )
+        expect(screen.getByTestId('box').innerHTML).toMatchInlineSnapshot(
+            `"Hello <strong>world</strong>!"`,
+        )
+    })
+
+    it('can be rendered as any HTML element', () => {
+        render(<Box data-testid="box" as="section" />)
+        expect(screen.getByTestId('box').tagName).toBe('SECTION')
+    })
+
+    it('applies any extra className given to it', () => {
+        render(<Box data-testid="box" className="custom" />)
+        expect(screen.getByTestId('box')).toHaveClass('custom')
+    })
+
+    it('applies some extra class names corresponding to other styling props', () => {
+        render(
+            <Box
+                data-testid="box"
+                display="inlineBlock"
+                maxWidth="xlarge"
+                minWidth="xsmall"
+                border="standard"
+                borderRadius="full"
+                background="aside"
+                overflow="hidden"
+            />,
+        )
+        expect(screen.getByTestId('box')).toHaveClass(
+            'box',
+            'display-inlineBlock',
+            'minWidth-xsmall',
+            'maxWidth-xlarge',
+            'bg-aside',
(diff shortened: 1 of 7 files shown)


gnapse:
ping @jvalente

jvalente:
Sorry @gnapse don't have my Github notifications coming through in a proper way. 
Will review today.

jvalente (approved):

gnapse:
> since there's multiple components using `exceptionallySetClassName`, would it make sense to have such behaviour extracted into an HOC and tested only once?

The implementation of that feature is already done once via the `polymorphicComponent` helper (which is not really a HOC, but a wrapper around `React.forwardRef`). Regarding the repeated tests, that's a good point. Will think about it.
 
> the "can be rendered as any HTML element" tests are in fact testing only one specific HTML element. Would it make sense to test that `element.tagName === component.props.as`?

Not sure that I follow your point here. The way the tests are written already ensure that whatever you're passing in the `as` prop is what you get as the `tagName`. Take this test, for instance:

```js
    it('can be rendered as any HTML element', () => {
        render(<Box data-testid="box" as="section" />)
        expect(screen.getByTestId('box').tagName).toBe('SECTION')
    })
```

Isn't it clear in that test that the tag name came from the `as` prop? Maybe you refer to testing this via setting the `as` prop more than once to different values, and check that when it changes, the `tagName` changes? If so, I'm not sure that we need all that complexity. The test is there mostly to make sure we do not inadvertently break this feature. But let me know if it's that or not, and what you think about what I said.

gnapse:
@jvalente to move a bit faster, I'll take your approval and will merge this, as I have a couple more PRs on top of this one, waiting to be published. I can take into account both your feedback and anything from @nats12 in PRs in the recent future, as I'll be making quite a few these days.

jvalente:
@gnapse for sure! Please don't be blocked. My comment were merely curiosities so that I can speed up my understanding on how we approach the test development and component organization. The PR is 👍👍

> The implementation of that feature is already done once via the polymorphicComponent helper.

Gotcha, will take a look!

> Maybe you refer to testing this via setting the as prop more than once to different values, and check that when it changes, the tagName changes? If so, I'm not sure that we need all that complexity.

Agreed, it's a simple thing. My point was more to try to understand if it is a common practice to access the rendered component props on the test.

gnapse:
> if it is a common practice to access the rendered component props on the test.

Actually, we cannot. I was also wondering what you meant with `component.props.as`?

I suspect you may be referring to something about how enzyme tests are written. Bear in mind that we do not use enzyme anymore. We now use React testing library (`@testing-library/react`). The main difference is that this library actually makes components render to a real DOM (headless browser) and then we are only able to check against the HTML that these components render. That is, aside from the moment we render a component, from then on in the test the concept of `props`, `state` and `component` no longer exist. The HTML is all there is. This ensures that tests do not check up on implementation details (something that was common in enzyme tests).

jvalente:
This is exactly what I was looking for @gnapse. This way of testing is way better. Thank you very much for the explanation! 🙌

PR #536, opened 2021-08-16
Box component improvements

gnapse:
## Short description

Adds a few new features or improvements to the `Box` component:

1. Refactor padding styles, plus new storybook example about padding.
2. Introduce support for applying semantically spaced margin values to boxes (including negative margin).
3. Add a new value `full` to `Box`'s `maxWidth` prop (which sets the box's `max-width: 100%`).

## PR Checklist

## Versioning

This is being merged into a temporary base branch to be released in about a week or so alongside other changes.

## Demo

https://user-images.githubusercontent.com/15199/129632848-dd4431c3-1cfa-4e2d-aa75-5d11c23fb0e3.mp4


src/new-components/box/margin.module.css
```diff
@@ -0,0 +1,507 @@
...
+.marginTop--xlarge {
+    margin-top: calc(var(--reactist-spacing-xlarge) * -1);
+}
+.marginTop--xxlarge {
+    margin-top: calc(var(--reactist-spacing-xxlarge) * -1);
+}
+
+/* margin-top for tablet */
+@media (min-width: 768px /* --reactist-breakpoint-tablet */) {
+    .tablet-marginTop-xsmall {
```
pauloslund, line 46: I dug into how these classes are getting applied based on the values provided to the props, it's really cool!

src/new-components/storybook-helper.tsx
```diff
@@ -63,7 +63,7 @@ function times(count: number): number[] {
 
 function reusableBoxProps(): Partial<Record<keyof BoxProps, ReturnType<typeof selectWithNone>>> {
     return {
-        maxWidth: selectWithNone(['xsmall', 'small', 'medium', 'large', 'xlarge']),
+        maxWidth: selectWithNone(['xsmall', 'small', 'medium', 'large', 'xlarge', 'full']),
```
pauloslund, line 66: What are your thoughts on adding the margins to this helper as well? I only ask because the `Box/Interactive Props Story` now includes padding values, but not margins which feels a little strange. Maybe it would make more sense to just add the margin value directly to the Interactive Props Story so it doesn't interfere with other ones in potentially unknown ways.

pauloslund (approved):
Looks good and works as expected! The new stories are definitely helpful 👍

src/new-components/storybook-helper.tsx
gnapse: Margin, unlike padding, is not added to what we call the "reusable box props". They are reusable in the sense that other higher-level layout components are allowed to have these props set on them directly as well. And it is not added on purpose: applying margin to Stacks, Columns, etc. may easily break them. TBH, even padding could break them, but not as easily. So, if at all, I may remove the ability to apply padding to anything other than a `Box`, instead of the other way around that you suggest.

Part of this is related to what I covered in [this video](https://www.dropbox.com/s/4cbsu5bum5ewgpl/2021-08%20design%20system%20and%20exceptionallySetClassName.mov?dl=0) that I shared a while back. Specifically the part where I talked towards the end about nesting layout components.

PR #540, opened 2021-08-19
feat: New Tabs component based on Reakit/Tab

frankieyan:
## Short description

This adds a new `Tabs` component based on top of https://reakit.io/docs/tab, deprecating the old `Tabs` that hasn't been used too much within our apps.

Also added the `usePrevious` hook from our apps and a quick story for it as well.

image

## Test plan
- Check out the `Docs` tab at http://localhost:6006/?path=/docs/components-tabs
- [ ] Do the set of props being exposed and how they are named make sense?
- [ ] Go to the `Canvas` tab and play around with the controls. Is the component being updated accordingly when props change?

## PR Checklist

## Versioning

Next beta


src/new-components/tabs/tabs.tsx
```diff
@@ -0,0 +1,197 @@
...
+type TabPanelProps = {
+    /** The content to be rendered inside the tab */
+    children: React.ReactNode
+
+    /** The tabPanel's identifier. This must match its corresponding `<Tab>`'s id */
+    id: string
+
+    /**
+     * By default, the tab panel's content is not rendered unless its tab is active. This behaviour can be overridden
+     * by setting `lazyload={false}`. Note that the children component tree will still execute, including any side
```
frankieyan, line 172: This doesn't prevent the components from executing, but the alternative is to probably render content via a render prop, so we canexpose the tabState or simply whether the tab is active. Is there another way?

frankieyan:
@gnapse I haven't based this off of your branch as I haven't made use of too many layout components, other than `Inline`, which I don't suppose have its interface changed. Let me know if that's not the case 👍

src/new-components/tabs/tabs.test.tsx
```diff
@@ -0,0 +1,129 @@
+import * as React from 'react'
+import { screen, render } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { Tabs, Tab, TabList, TabPanel } from './'
+
+describe('Tabs', () => {
+    it("allows each TabPanel's visibility controlled by each tab", () => {
```
gnapse, line 7:
```suggestion
    it("allows each TabPanel's visibility to be controlled by its corresponding tab", () => {
```

src/new-components/tabs/tabs.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+    it("setting each TabPanel's lazyload prop to false renders tabs even when they aren't active", () => {
+        render(
+            <Tabs>
+                <TabList aria-label="test-tabs">
+                    <Tab id="tab1">Tab 1</Tab>
+                    <Tab id="tab2">Tab 2</Tab>
+                    <Tab id="tab3">Tab 3</Tab>
+                </TabList>
+                <TabPanel id="tab1" lazyload={false}>
```
gnapse, line 47: I'd call it `lazy` instead of `lazyload`. And I'd make it false by default.

src/new-components/tabs/tabs.tsx
```diff
@@ -0,0 +1,197 @@
...
+type TabListProps = (
+    | {
+          /** Label the tab list for assistive technologies. This must be provided if `aria-labelledby` is omitted */
+          'aria-label': string
+      }
+    | {
+          /**
+           * One or more element ID's used to label the tab list for assistive technologies. Required if
+           * `aria-label` is omitted
+           * */
```
gnapse, line 125:
```suggestion
          /**
           * One or more element IDs used to label the tab list for assistive technologies. Required if
           * `aria-label` is omitted.
           */
```

src/new-components/tabs/tabs.tsx
```diff
@@ -0,0 +1,197 @@
...
+            {...tabState}
+        >
+            {children}
+        </BaseTab>
+    )
+}
+
+type TabListProps = (
+    | {
+          /** Label the tab list for assistive technologies. This must be provided if `aria-labelledby` is omitted */
```
gnapse, line 118:
```suggestion
          /** Labels the tab list for assistive technologies. This must be provided if `aria-labelledby` is omitted. */
```

src/new-components/tabs/tabs.tsx
gnapse: There's a third mode that could be useful: do not render inactive tabs at first, but once they're activated at least once, getting out of them and back in does not re-render them.

I realize this may be tricky to achieve in a generic way, but it was super useful in the members & groups UI in the new Twist settings. You see, I did not want both tab panels to render at first, for performance reasons. But once I visited the groups' panel, I did not want it to unmount and re-mount. It would mean I loose transient state in it, such as the scroll position on the groups list, for instance.

If this could be possible to support, maybe this prop should be an enum-like prop instead of boolean (e.g. `render: 'always' | 'active' | 'lazy'` where lazy would be the mode I described).

src/new-components/tabs/tabs.tsx
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Used to group components that compose a set of tabs. There can only be one active tab within the same `<Tabs>` group.
+ */
+function Tabs({
+    children,
+    selectedId,
+    color = 'primary',
+    variant = 'normal',
+}: TabsProps): React.ReactElement {
+    const tabState = useTabState({ selectedId })
```
gnapse, line 47: By fully owning the state management inside this component, we lose some control. I'm thinking, for instance, in the members & groups UI in the New Twist settings. In it, there was content outside the tab panels that was affected by what was the currently selected tab.

https://user-images.githubusercontent.com/15199/130153170-f0e983c1-5a93-4721-9f20-e09736a8f5be.mp4

See how, under this design, there's some UI in the unused area to the right of the tab list. This part of the UI is not rendered inside the tab panels, but the fact that I have access to the state (namely the currently active tab) I was able to conditionally render what's in that area.

I realize this is more convenient, but if a component like this is going to power all our tab-like UIs, it may need to think about this in its design, at least in the long run. It's ok if this gets this PR out of scope for your immediate needs, though.

gnapse (commented):
Some preliminary comments. I can try this soon and make a final full review plus QA.

src/new-components/tabs/tabs.tsx
frankieyan: > Also, I'm not sure that it is true that if the sub-tree does not render, side effects will still happen. Can you ellaborate on that? What'll happen is that react will still generate that sub-tree of React elements, but it will never mount them, hence effects inside those components will never execute.

I could have sworn both the function body and effects executed last night, but trying it again the un-rendered children don't have their component function bodies run anymore. False alarm!

src/new-components/tabs/tabs.tsx
frankieyan: Yes, that's a valid use case 👍 I've added a `<TabAwareSlot>` component to provide a render prop for this

gnapse: Wow, that's a really cool solution. Thanks!

src/new-components/tabs/tabs.tsx
```diff
@@ -0,0 +1,228 @@
...
+}
+
+/**
+ * Allows content to be rendered based on the current tab being selected while outside of the TabPanel
+ * component. Can be placed freely within the main `<Tabs>` component.
+ */
+function TabAwareSlot({ children }: TabAwareSlotProps): React.ReactElement | null {
+    const tabContextValue = React.useContext(TabsContext)
+
+    return tabContextValue ? children({ selectedId: tabContextValue.selectedId }) : null
```
gnapse, line 225: I wonder if there's anything interesting in the `tabContextValue` to be shared with consumers of this component. No need to look into it now, though. We can wait until the need arises, if ever.

gnapse (approved):
Cool, good work. It works as expected, and looks good.

One thing though: I'd say we can remove the old tabs component in this very PR. This is going to beta anyway, and we'll (hopefully soon) release beta into main and will make a breaking change release. The only possible counterargument I see is that we may be using the old tabs in twist-web, and migrating the current use cases is either hard, or not worth it because it's old Twist code that will be removed. Is that the case?

Also: can we put the `Tabs` story under the "design system" nav header?

frankieyan:
> One thing though: I'd say we can remove the old tabs component in this very PR. This is going to beta anyway, and we'll (hopefully soon) release beta into main and will make a breaking change release. The only possible counterargument I see is that we may be using the old tabs in twist-web, and migrating the current use cases is either hard, or not worth it because it's old Twist code that will be removed. Is that the case?

It looks like we have no references to the old tabs in both apps. Let's remove it 👍

> Also: can we put the Tabs story under the "design system" nav header?

I was under the assumption that the Design System grouping was more for layout/global components, and we could be grouping components separately similar to how Marketist organizes them (e.g. input components, buttons, forms etc), but I can move it over first and discuss later.

src/new-components/tabs/tabs.tsx
frankieyan: Here is the full list: https://github.com/reakit/reakit/blob/a211d94da9f3b683182568a56479b91afb1b85ae/packages/reakit/src/Tab/TabState.ts#L13

I guess because the tab state isn't documented, they feel a bit internal to Reakit tab's implementation, although I've made use of `select` already. That said, I don't think there's much else in there that needs to be exposed for now, let's wait for a use case as you said 👍

src/new-components/tabs/tabs.tsx
```diff
@@ -125,6 +125,12 @@ type TabListProps = (
            */
           'aria-labelledby': string
       }
+    | {
+          /**
+           * For cases where multiple instances of the tab list exists, the duplicates may be marked as aria-hidden
```
gnapse, line 130: What do these cases (where multiple instances of the tab list exists) look like?

frankieyan: I have a use case for [rendering the tab buttons on the thread list's sticky header](https://paper.dropbox.com/doc/Inbox-1.1-Done-tabview--BQ2nJwTOQGKh2v0k3L7_RpU9Ag-Lwc0Vs9AnWb3BA6uLy3VX#:h2=Scrolling)

What I'm thinking is that this `<TabList>` will sit inside of a `<TabPanel>`. I tested with a simple case in storybook and it seems to re-render correctly 🤞

gnapse:
One last thing I thought about: will this tab component allow to be controlled rather than uncontrolled? Meaning, you control the selected tab state yourself based on something external to the component itself.

This is important for router-driven tab UIs. For instance, both the members & groups UI of the new Twist settings, and the subtasks/comments/activity tabs in a Todoist task details view will need this.

cc @frankieyan

src/new-components/tabs/tabs.tsx
gnapse: Yeah, it's ok for now. If something comes up, we'll see what to do about it.

src/new-components/tabs/tabs.tsx
gnapse: Any comments on the third mode I mentioned? Even if you do not cover it now, I'd say you should make the prop be a union type rather than a boolean: `render: 'always' | 'active'`. Thoughts?

frankieyan: It's already implemented @gnapse:

https://github.com/Doist/reactist/blob/f40dc7a2e5b2b83eeb376c00abd15f7221ec5e82/src/new-components/tabs/tabs.test.tsx#L77

gnapse: Oh, I missed that. Nice, thanks!

Happy Parks And Rec GIF

frankieyan:
> One last thing I thought about: will this tab component allow to be controlled rather than uncontrolled? 

Yup:

https://github.com/Doist/reactist/blob/f40dc7a2e5b2b83eeb376c00abd15f7221ec5e82/src/new-components/tabs/tabs.test.tsx#L115

While Reakit allows you to set a default selectedId [ref](https://reakit.io/docs/tab/#default-selected-tab), changing the tab via prop changes only seemed to be achievable by calling `select` from the TabState off of an effect, as re-running `useTabState` with the new selectedId don't seem to re-render the tab components even when they receive a new TabState.

gnapse:
Cool, it seems you've got it all covered and I missed some of these details. This is super good to go.

PR #542, opened 2021-08-19
Text and Heading tests, new stories and improvements

gnapse:
## Short description

Most notably, this PR adds `textAlign` support for `Box`, and uses it to implement a `align` prop in both `Heading` and `Text`.

In addition to that, it adds tests for both the `Heading` and `Text` components, and new stories to showcase the `align` responsive feature. It also adds jsdoc comments to the props for each of these two components.

## PR Checklist

## Versioning

To be released in bulk with other features that have been accumulating in this PR's base branch.

## Demo

https://user-images.githubusercontent.com/15199/130112038-b67ad21f-e085-4bf1-bc0f-122a55820284.mp4

### DIFF
--- a/src/new-components/box/box.module.css
+++ b/src/new-components/box/box.module.css
@@ -328,3 +328,45 @@
 .border-standard {
     border: 1px solid var(--reactist-framework-border);
 }
+
+/* text-align */
+.textAlign-start {
+    text-align: start;
+}
+.textAlign-center {
+    text-align: center;
+}
+.textAlign-end {
+    text-align: end;
+}
+.textAlign-justify {
+    text-align: justify;
+}
+@media (min-width: 768px /* --reactist-breakpoint-tablet */) {
+    .tablet-textAlign-start {
+        text-align: start;
+    }
+    .tablet-textAlign-center {
+        text-align: center;
+    }
+    .tablet-textAlign-end {
+        text-align: end;
+    }
+    .tablet-textAlign-justify {
+        text-align: justify;
+    }
+}
+@media (min-width: 992px /* --reactist-breakpoint-desktop */) {
+    .desktop-textAlign-start {
+        text-align: start;
+    }
+    .desktop-textAlign-center {
+        text-align: center;
+    }
+    .desktop-textAlign-end {
+        text-align: end;
+    }
+    .desktop-textAlign-justify {
+        text-align: justify;
+    }
+}
--- a/src/new-components/box/box.test.tsx
+++ b/src/new-components/box/box.test.tsx
@@ -90,6 +90,35 @@ describe('Box', () => {
         )
     })
 
+    describe('textAlign="…"', () => {
+        it('adds the appropriate class names', () => {
+            const { rerender } = render(<Box data-testid="box" textAlign="start" />)
+            const boxElement = screen.getByTestId('box')
+            expect(boxElement).not.toHaveClass('textAlign-start')
+            expect(boxElement).not.toHaveClass('textAlign-center')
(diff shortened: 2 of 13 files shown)


gnapse:
ping @nats12

src/new-components/box/box.module.css
```diff
@@ -328,3 +328,45 @@
...
+    .desktop-textAlign-center {
+        text-align: center;
+    }
+    .desktop-textAlign-end {
+        text-align: end;
+    }
+    .desktop-textAlign-justify {
+        text-align: justify;
+    }
+}
```
nats12, line 372: Since everything ending in `textAlign-center` will always be aligned to the center and so forth, could we condense these styles by just having: 

```
[class*="textAlign-center"] {
  text-align: center;
}
```
..etc for center, end, justify and start. What do you think?

nats12: Out of curiosity, why do we need the breakpoints? Would the expectation not be that anything containing `textAlign-center` is always centred regardless of the breakpoint? Unless there are some other styles that these are overriding

nats12 (commented):
Hey @gnapse, just left some comments regarding some css styles. I didn't see any README/CHANGELOG changes amongst them as specified in the plan. Will you also be updating these?

src/new-components/box/box.module.css
gnapse: This answers both of your comments:

Nope, what you suggest will actually break the system. Here's how it works:

The prefix `tablet-` or `desktop-` means that we only want that style to kick in when on screen sizes within the sizes of tablets or desktops.

Imagine a component like this:

```jsx
<Box textAlign={{ mobile: 'end', tablet: 'center', desktop: 'start' }} />
```

In HTML, it will end up being a div like this:

```html
<div className="box textAlign-end tablet-textAlign-center desktop-textAlign-start" />
```

So, if we do the styles of those classes like this:

```css
[class*="textAlign-start"] {
  text-align: start;
}
[class*="textAlign-center"] {
  text-align: center;
}
[class*="textAlign-end"] {
  text-align: end;
}
```

That div will always have its text aligned to the start or end (I'm not sure how the CSS rules are resolved when there are conflicting ones). But the fact is that all those styles will apply to it.

But what we really want is for certain styles to apply only on certain screen sizes. That's why the class names with the `tablet-` and `desktop-` prefixes are only applied inside a media query. The style for "mobile" does not have a prefix or a media query because this is the default style (however odd it may sound). How it works is: we apply the default style, which, if there are no overriding for tablet or desktop, will be the one that applies all the time. But, if there are overrides for tablet, for instance, this will effectively make the default class without a tablet/desktop prefix apply only on mobile screen sizes, because the override for tablet will kick in on any device larger than a mobile screen size.

Hope I answered your questions, and I did not misinterpreted what you meant.

gnapse:
@nats12 regarding the changelog, I'm not making a release after this PR. I've gathered a few PRs in a branch, from which I'll make a release next week. A release that will include this and several other recent changes. The PR where I'll prepare the release will feature a changelog update.

gnapse:
@nats12 is there anything left to address before you approve this? Let me know.

src/new-components/box/box.module.css
nats12: Hey @gnapse, thanks for the explanation, seeing it like this `<div className="box textAlign-end tablet-textAlign-center desktop-textAlign-start" />` makes a lot of sense why it wouldn't work!

nats12 (approved):
> @nats12 is there anything left to address before you approve this? Let me know.

Looks good to me! Nothing left to address @gnapse, I now understand the reasoning behind the classes 👍🏼

PR #550, opened 2021-08-25
Improvements to LoadingSpinner

gnapse:
## Short description

The main motivation was that the loading spinner was colored red, and it was using a CSS variable that was guaranteed to be red according to its name, even if the actual red shade could change. I've now changed it to be colored based on a brand color name variable, which is for apps to define freely. This will solve being able to see this loading spinner colored blue in Twist, but red in Todoist.

And given that I was already working on this component, I took the time to write some documentation, stories/examples, and tests. I also took the time to rename it as `Loading` and have it replace the legacy `Loading` component that we had. This is a breaking change that I'll take care of updating in the app.

♿️ I also changed it from the accessibility POV to be a `role="progressbar"` with indeterminate progress, rather than being a `role="alert"`. cc @craigcarlyle in case you can comment on this direction I'm taking.

## PR Checklist

## Versioning

Will go out in an upcoming new beta release soon.

## Demo

CleanShot 2021-08-25 at 17 02 08


gnapse:
ping @Bloomca, this is now fully ready for a review.

src/new-components/loading/loading.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+
+const sizeMapping: Record<Size, number> = {
+    small: 24,
+    medium: 36,
+    large: 48,
+}
+
+function Loading({ size = 'small', label }: LoaderProps) {
+    const numericSize = sizeMapping[size]
```
Bloomca, line 26: question: Do you think it makes any sense to provide a default value? Like what would happen if we pass a wrong value?

Bloomca (approved):
Looks good to me, I left a small question regarding check that value is supported, but it is a minor thing.

src/new-components/loading/loading.tsx
gnapse: I was relying on TypeScript here, but good point. We should provide a fallback to the same value as "small" if an invalid value is passed in some code consuming reactist without type checking in place.

PR #552, opened 2021-08-26
docs: Use Docs as Storybook default tab

frankieyan:
## Short description

We're starting to have some awesome stories in place, which are much more useful for devs looking for information on how to use the actual components. Let's make these docs tabs default rather than the canvas playgrounds.

Ref: https://github.com/storybookjs/storybook/issues/12111#issuecomment-676094000

## Test plan
- [x] When clicking on the nav items, you should be navigated to the docs tab rather than the canvas tab of each component

## PR Checklist

## Versioning

Next beta


gnapse:
I'm rebasing this to make it green.

One thing I'm not sure about: in my experience, I cannot see how to use controls in live examples when in doc mode. I always have to switch to the canvas mode. Is it expected to be this way?

frankieyan:
>  I cannot see how to use controls in live examples when in doc mode. I always have to switch to the canvas mode. Is it expected to be this way?

I looked through the docs and don't see any mention of this either. Given that controls take up as much space as a prop table, it might not make much sense to have them rendered inline though even if we're able to.

After digging through, I did find that we could also customize the order of the tabs: https://github.com/storybookjs/storybook/blob/next/addons/docs/docs/recipes.md#reordering-docs-tab-first

gnapse:
I gotta say I'm not a fan of this change. I find the canvas view much more helpful (mostly because that's where I see the controls to play with a story). I'm wondering if I'm missing some advantage for this to make sense.

frankieyan:
@gnapse I feel that once we have a wider audience (ie. designers, integration devs) use Reactist more, the docspages will be more useful for them as introduction to the components. I'm not holding that stance that strongly though and since it's still mostly our tool for internal development let's optimize for that. Please feel free to change it back to the canvas tab 👍

gnapse:
There are two major reasons why dislike it:

- In the canvas tab I do not have access to the controls. We could solve that by making most stories less centric on using controls. We kinda do that already in a few cases, but I'd take it to the extreme.
- It changes you back to dock each time you navigate. That is, I'm ok with it being "Docs" by default at first, but if in a certain story on the sidebar I change to "Canvas", and then navigate to other story within the same section/component, it switches me back to "Docs", for every click I do on the sidebar.

I won't change it for now. We'll see.

frankieyan:
> It changes you back to dock each time you navigate

Ah yes, this is definitely annoying. Maybe there is a way to change the navgiation's links based on your current docs/canvas selection.

PR #554, opened 2021-09-01
chore: Add step to read node version from .nvmrc file

nats12:
## Short description

Adds a step to read the node version from `.nvmrc` and then uses the outputted `NODE_VERSION` across other workflow files. PRs have been failing as the variable was being used outright without having the step to read from the file first.

Note: I had a test error regarding an irrelevant `await` operator on `userEvent`, I've fixed this in this branch too since it was preventing me from pushing.

## PR Checklist

-   [ ] CI should be green, especially coverage reporting


pedroalves0 (commented):
There's a lint error related to that test change 🤔

gnapse (approved):

PR #560, opened 2021-09-07
doc: Add new stories with tips and tricks

gnapse:
## Short description

Things I found worth documenting after the work this last month.

## PR Checklist

## Versioning

Documentation only, so there's no need to release this. Just merge when approved.


stories/reactist/simplified-nested-markup.stories.mdx
```diff
@@ -0,0 +1,132 @@
...
+</div>
+```
+
+Originally, our components were like that of Braid, but this issue was brought up internally, and we decided to make it simpler, even if it adds a burden to the consumer of these components. A burden that we need to be aware of. You can find more information about this change in [the PR that introduced it](https://github.com/Doist/reactist/pull/497). The concern is not ours alone, as it [was raised in Braid's own repo as well](https://github.com/seek-oss/braid-design-system/issues/386).
+
+We explain below why this change made sense.
+
+## Why we made the change
+
+A good deal of the time, what we nest inside a `Stack` is already a block element. Also, we favour having components without outward margin in the out-most container. So **there's no concern that the margin that the `Stack` applies to its child elements will clash with margin that the child elements themselves may have**. So it seems a waste of resources to have a UI language that could potentially generate lots of useless extra containers around a relatively simple UI.
```
rfgamaral, line 48:
```suggestion
A good deal of the time, what we nest inside a `Stack` is already a block element. Also, we favour having components without outward margin in the out-most container. So **there's no concern that the margin that the `Stack` applies to its child elements will clash with the margin that the child elements themselves may have**. So it seems a waste of resources to have a UI language that could potentially generate lots of useless extra containers around a relatively simple UI.
```

src/new-components/box/box.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+## You don't need the `classNames` helper as often now
+
+The new `Box` component's `className` prop receives the same kind of values supported by this helper. So you can do things like this:
+
+```jsx
+// This…
+<Box className={classNames(styles.container, 'custom-class-name')} />
+
+// …can be written like this
+<Box className={[styles.container, 'custom-class-name']} />
```
pauloslund, line 16: 😍 This is so great!

rfgamaral (approved):
LGTM, just added very minor suggestions.

src/new-components/box/box.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+    { [styles.disabled]: isDisabled },
+  ]}
+/>
+```
+
+Granted, this is only supported in the `Box`. This is because the `Box` is the only design system component that officially supports custom styles. Remember that for the other components you can use `exceptionallySetClassName`, but this one is an escape hatch, that we should aim to not use as much as the design system matures. So this one does not support the feature.
+
+## The `paddingX` and `paddingY` shorthand props
+
+The `Box` component has props to set its padding in all four direction: `paddingTop`, `paddingRight`, `paddingBottom`, `paddingLeft`. You probably have noticed that we can set them all at once with `padding`.
```
pauloslund, line 32:
```suggestion
The `Box` component has props to set its padding in all four direction: `paddingTop`, `paddingRight`, `paddingBottom`, `paddingLeft`. You have probably noticed that we can set them all at once with `padding`.
```

src/new-components/box/box.stories.mdx
```diff
@@ -0,0 +1,41 @@
...
+
+But you probably missed the fact that we can set the padding in one of the two dimensions at once as well. This is achieved with `paddingX` (which sets `paddingLeft` and `paddingRight`) and `paddingY` (which sets `paddingTop` and `paddingBottom`).
```
pauloslund, line 34: Super minor note just to remove repeating `probably`
```suggestion
But you might have missed the fact that we can set the padding in one of the two dimensions at once as well. This is achieved with `paddingX` (which sets `paddingLeft` and `paddingRight`) and `paddingY` (which sets `paddingTop` and `paddingBottom`).
```

stories/reactist/nesting-vs-polymorphism.stories.mdx
```diff
@@ -0,0 +1,70 @@
...
+The main purpose is to be able to take advantage of the new components while at the same time not dropping the ball on using semantic HTML.
+
+However, it can also be used to render as another React component:
+
+```jsx
+// Link below can be thought of as the Link component in react-router
+<ButtonLink as={Link}>Click me</ButtonLink>
+```
+
+This is useful to be able to get the functionality of one component with the visuals and functionality of another component. In the example above, we render a link that has the visual look and feel of a `Button` from our design system, but in which the link functionality is provided by react-router, allowing this link to perform <abbr title="Single-Page Application">SPA</abbr>-like navigation on the client-side instead of a full server round trip.
```
pauloslund, line 26:
```suggestion
This is useful to be able to get the functionality of one component with the visuals and functionality of another component. In the example above, we render a link that has the visual look and feel of a `Button` from our design system; but the link functionality is provided by react-router, allowing this link to perform <abbr title="Single-Page Application">SPA</abbr>-like navigation on the client-side instead of a full server round trip.
```

stories/reactist/simplified-nested-markup.stories.mdx
```diff
@@ -0,0 +1,132 @@
...
+    </Stack>
+</Stack>
+```
+
+If we generated extra containers for each child, this would generate a markup similar to what's shown below:
+
+```html
+<div class="stack space-xlarge">
+  <div class="stack child-wrapper">
+    <div class="stack space-small>
```
pauloslund, line 72:
```suggestion
    <div class="stack space-small">
```

stories/reactist/simplified-nested-markup.stories.mdx
```diff
@@ -0,0 +1,132 @@
...
+      <div class="stack child-wrapper">
+        <label>Username</label>
+      </div>
+      <div class="stack child-wrapper">
+        <input name="username" />
+      </div>
+    </div>
+  </div>
+  <div class="stack child-wrapper">
+    <div class="stack space-small>
```
pauloslund, line 82:
```suggestion
    <div class="stack space-small">
```

stories/reactist/simplified-nested-markup.stories.mdx
```diff
@@ -0,0 +1,132 @@
...
+</div>
+```
+
+And just imagine if inside the inner stack we rendered a `Inline` element, which also did the same for its child elements.
+
+### How it looks like without the extra container element
+
+```html
+<div class="stack space-xlarge">
+    <div class="stack space-small>
```
pauloslund, line 100:
```suggestion
    <div class="stack space-small">
```

stories/reactist/simplified-nested-markup.stories.mdx
```diff
@@ -0,0 +1,132 @@
...
+        <label>Username</label>
+        <input name="username" />
+    </div>
+    <div class="stack space-small>
```
pauloslund, line 104:
```suggestion
    <div class="stack space-small">
```

pauloslund (approved):
I really appreciate this added documentation! Reading through it all it was really easy to follow, and the examples made a lot of sense.

There are few nits that I noticed that I left comments on, but nothing major.

stories/reactist/nesting-vs-polymorphism.stories.mdx
```diff
@@ -0,0 +1,70 @@
...
+    <Column as={Stack} space="small">
+        <Three />
+        <Four />
+    </Column>
+</Columns>
+```
+
+This is not guaranteed to work 100% of the time because styles could clash.
+
+In the above example, for instance, the `div` generated by each `Column` element will have margin applied to it to achieve the `space="medium"` effect. And the `div` generated by each `Stack` will have margin applied to it to achieve the `space="small"` effect of the `Stack`. You'd think is not the same margin in the same direction, because stacks add vertical spacing while columns add horizontal spacing. However, columns add vertical spacing also, for when they show in collapsed form due to the viewport getting narrower (i.e. responsive design).
```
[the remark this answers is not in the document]
gnapse: I'll reword this differently to see if you get it, instead of trying to explain it to you. It probably is not clear for future readers either.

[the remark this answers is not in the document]
gnapse:
```suggestion
In the above example, for instance, the `div` generated by each `Column` element will have a margin applied to it to achieve the `space="medium"` effect. While the `div` generated by each `Stack` will have margin applied to it to achieve the `space="small"` effect of the `Stack`.

At first, you might think there's no problem because the margin on the `Stack` child elements is applied in the vertical direction (top and bottom) to keep a vertical space between them. While the margin on the `Column` elements is applied in the horizontal direction (left and right) to keep columns separated. So, in theory, these margins should not clash. However, columns add vertical spacing also, for when they show in collapsed form due to the viewport getting narrower (i.e. responsive design).
```

src/new-components/box/box.stories.mdx
gnapse: Ricardo suggested another unrelated edit to this very line, so I'll apply your suggestion alongside Ricardo's suggestion.

src/new-components/box/box.stories.mdx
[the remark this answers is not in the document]
gnapse: I'm combining here your suggestion with another change suggested by Paul below:

```suggestion
The `Box` component has props to set its padding in all four direction: `paddingTop`, `paddingRight`, `paddingBottom`, `paddingLeft`. You have probably noticed that we can set them all at once with `padding`. But you probably missed the fact that we can set the padding in one of the two dimensions at once as well. This is achieved with `paddingX` (which sets `paddingLeft` and `paddingRight`) and `paddingY` (which sets `paddingTop` and `paddingBottom`).
```

PR #563, opened 2021-09-09
Button improvements

gnapse:
## Short description

A few improvements to the new button component

- Properly styles the secondary destructive button that was previously not fully finished design-wise in a way that could be customizable in dark mode. (see https://github.com/Doist/Polish/issues/342#issuecomment-915151574)
- Adds support for the loading state on the button (see https://twist.com/a/1585/ch/81455/t/2722260/c/70122683).
- Adds examples in the stories for how to customize buttons for dark mode.
- Improves button icons spacing and sizes. It also improves the ability to fully customize all sizing and spacing by apps via CSS variables.
- Use a `<Button variant="quaternary" />` to power the modal's close button. This allows this button to follow any app's styling customizations applied to buttons.

## PR Checklist

## Demo

https://user-images.githubusercontent.com/15199/132773357-6ee587cc-1df8-44a0-a86a-9b3ac9891dbe.mp4

## Versioning

New beta patch release.


src/new-components/button-link/button-link.test.tsx
```diff
@@ -258,4 +264,82 @@ describe('ButtonLink', () => {
             expect(buttonLink.textContent).toMatchInlineSnapshot(`"😄"`)
         })
     })
+
+    describe('when loading={true}', () => {
```
frankieyan, line 268: I wonder what the use case for this is. Being a link users can right-click and open the target link in a new tab, skipping its onClick prop. If we can't think of a use case then maybe we don't expose it as an option yet?

src/new-components/button-link/button-link.stories.mdx
```diff
@@ -475,79 +487,92 @@ export function Template({ label, variant, tone, size }) {
...
-                        size={size}
-                        aria-label={textLabel}
-                        tooltip={label}
-                        icon={<Icon />}
-                        disabled
-                    />
-                </Box>
-            </Stack>
-        </Inline>
+        <Stack space="large">
```
frankieyan, line 490: For some reason the props table isn't rendering for the ButtonLink:

image

src/new-components/modal/modal.module.css
```diff
@@ -130,21 +130,7 @@
...
-    border: none;
-    line-height: 0;
-    margin-right: -8px;
-}
-.closeButton:hover,
-.closeButton:focus {
-    background-color: var(--reactist-framework-fill-crest);
-    color: var(--reactist-content-primary);
+.closeButtonIcon {
+    margin-left: -4px;
```
frankieyan, line 134: Maybe we need to support a smaller size variant to replace Twist's ActionButton

src/new-components/modal/modal.tsx
```diff
@@ -142,13 +143,13 @@ function Modal({
 //
 
 type ModalCloseButtonProps = Omit<
-    JSX.IntrinsicElements['button'],
-    'type' | 'className' | 'onClick' | 'aria-label' | 'aria-labelledby' | 'children'
+    ButtonProps,
+    'type' | 'children' | 'variant' | 'icon' | 'startIcon' | 'endIcon' | 'disabled' | 'loading'
 > & {
     /**
      * The descriptive label of the button. Used as its aria-label attribute.
```
frankieyan, line 150: Should we remove the second part of the comment? It's a bit redundant

src/new-components/button-link/button-link.stories.mdx
frankieyan: I tried exporting the component inline like in Tabs but it only managed to fix the canvas tab but not on the docs tab. Worth a fix anyway if it at least enables it for canvas:

|Before|After|
|-|-|
|image|image|

src/new-components/button-link/button-link.test.tsx
frankieyan: Thinking on this some more, we may be generating a link that takes a while to resolve. What do you think about disallowing the onClick prop instead while allowing loading?

src/new-components/modal/modal.module.css
gnapse: Not sure what you mean, as this line where you're commenting on is irrelevant to the size of the button. Replacing the ActionButton is a future goal of mine. We can discuss this then.

src/new-components/button-link/button-link.stories.mdx
gnapse: I'll switch to inline export for the components affected in this PR.

src/new-components/button-link/button-link.test.tsx
gnapse: I thought about this probably making no sense for links, but I decided not to second guess what it would be useful for. Maybe I can take it off the table for links, and we only add it if it makes sense.

gnapse: Nah, I'll leave it. The Braid `ButtonLink` has it too. It won't hurt.

gnapse:
@frankieyan can you approve? I'm taking care of most of your feedback (will push it soon) and none of it seems blocking anyway. I may need this to push for new twist-web changes early today.

frankieyan (approved):

PR #565, opened 2021-09-10
Better adjust a modal's vertical position when in fitContent mode

gnapse:
## Short description

There are two modes by which modals handle vertical space (not introduced in this PR, is what we've already have):

1. Some modals expand to fill most of the viewport height (`height="expand"`).
2. Other modals fit their content's height (`height="fitContent"`).

> **Note:** in both cases the modal body's content is set to scroll if it becomes larger than the larger height they can fill.

### What's changing

This PR does not change what modes we have. It only changes how the mode `height="fitContent"` positions the modal vertically.

- **Before this PR**, these `fitContent` modals were vertically centred all the time. This caused a problem with modals that have content that causes them to change their height dynamically. These modals would then jump in both directions, shrinking and expanding as the content height changed.
- **With this PR**, the modal is kept at a fixed separation from the top of the viewport, and the bottom edge is the only one that will expand and shrink up and down as the content height changes.
    - In case the modal height expands too much, we also make sure that the bottom edge keeps separated from the bottom viewport edge at least the same amount of space we leave at the top.

Relevant thread: https://twist.com/a/1585/ch/81455/t/2755963/

### Demo

CleanShot 2021-09-10 at 16 48 56

## PR Checklist

## Versioning

I'll merge this, but will not release it on its own, as it is not urgent. It will get combined with future PRs into the next beta patch release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+## next
+
+-   [Fix] Modals that fit their content's height now expand their height only downwards, and are not
+    kept vertically centered, but slightly towards the top of the screen.
+
 ## v10.0.0-beta.6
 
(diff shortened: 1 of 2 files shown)


pedroalves0 (approved):
Works nicely 👏

PR #567, opened 2021-09-20
Fix modal button icon alignment, truncate button labels when too long

### LINKED ISSUE #566: Modal: Close button icon is misaligned
### Current behaviour

The close modal button seems to be misaligned when used in Twist:

Screen Shot 2021-09-20 at 11 09 39 AM

### Expected behaviour

The icon should be aligned

### Possible solutions

Remove the following from `modal.module.css`:

```css
.closeButtonIcon {
    margin-left: -4px;
    margin-top: -3px;
}
```

gnapse:
## Short description

Closes #566.

The reason has to do with the fact that I naively designed the button component with some variables to allow consumers of it to control the size of icons based on the size of the button. This has proven to cause more troubles and solve none. SVGs have proven to be more suitable to be left having the size they naturally have. So let's say that it is now the responsibility of consumers of the buttons to use SVGs that look ok inside the button on which they need an icon to appear.

In twist-web I had already realized this issue, and we changed these CSS variables to effectively set the icons' `width` and `height` to `auto`, which is the same as not setting them in the first place. This change now removes these "icon size" CSS variables altogether.

### Extra change: truncate button labels when too long

Here's a demo:

image

## PR Checklist

## Versioning

Will be released right after merging, as a new beta version.


craigcarlyle (approved):
@gnapse The changes here LGTM. One thing I think we should consider, however, is adding a `title` attribute to the `button` element with the value of the button's label so that in the rare cases where a button's text does truncate it is still fully readable for the user:

Screen Shot 2021-09-21 at 9 55 00 AM

What do you think?

gnapse:
> One thing I think we should consider, however, is adding a title attribute to the button element with the value of the button's label so that in the rare cases where a button's text does truncate it is still fully readable for the user:

But we'd have to add it all the time, preemptively. Because we cannot know when it will be needed or not. I wonder if this would be annoying and/or interfere with the tooltip. Speaking of which, consumers of the button can always add it themselves, or add a tooltip.

craigcarlyle:
That's fair. Let's roll with this as-is then 👍

PR #589, opened 2021-10-14
feat: Menu a11y improvements

craigcarlyle:
## Short description

This PR fixes a couple of accessibility issues with the `Menu` component:
* Shift focus to the button when the `Menu` component is closed
	* Adds a test case for this behaviour
* Adds automated a11y check with `axe` when `Menu` is initially rendered and when opened 
* Adds accessible badge to the component's story:
menu

## PR Checklist

## Versioning
This does not immediately require a new release.


gnapse:
> Shift focus to the button when the `Menu` component is closed

One would think that reakit took care of this. I wonder why that's not the case. I even wonder if this is a bug in reakit. Might be worth looking into raising the issue there, or see if someone has brought it up before.

PR #594, opened 2021-10-18
chore: Adds a11y badge, automated tests to numerous components

craigcarlyle:
## Short description

Adds the a11y badge and automated tests to:
- Box
- Button
- Columns
- Heading
- Hidden
- Inline
- PasswordField
- SelectField
- Stack
- Tabs
- Text
- TextField
- Tooltip

## PR Checklist

## Versioning
This does not immediately require a new release.


src/new-components/box/box.test.tsx
```diff
@@ -252,4 +253,13 @@ describe('Box', () => {
...
     })
+
+    describe('a11y', () => {
+        it('renders with no a11y violations', async () => {
+            const { container } = render(<Box>Hello world!</Box>)
+            const results = await axe(container)
+
+            expect(results).toHaveNoViolations()
+        })
+    })
```
gnapse, line 264: Is this a formality? For completeness? What's the value of testing that raw `div` renders without a11y violations?

src/new-components/button/button.test.tsx
```diff
@@ -346,4 +347,22 @@ describe('Button', () => {
...
+                    <Button variant="primary" tone="destructive">
+                        Destructive
+                    </Button>
+                </>,
+            )
+            const results = await axe(container)
+
+            expect(results).toHaveNoViolations()
+        })
+    })
```
gnapse, line 367: There's value in doing more tests for the button in other states: while disabled and while loading? These do affect some a11y related attributes. Maybe you can add those buttons to be rendered in this very test without adding more test cases.

I also think that icon-only buttons deserve their own test case, mirroring this test case, but rendering all buttons as icon-only buttons.

src/new-components/password-field/password-field.test.tsx
```diff
@@ -260,4 +261,15 @@ describe('PasswordField', () => {
         expect(inputElement).toHaveValue('password value')
         expect(screen.getByTestId('value')).toHaveTextContent('password value')
     })
+
+    describe('a11y', () => {
+        it('renders with no a11y violations', async () => {
+            const { container } = render(
+                <PasswordField data-testid="password-field" id="custom-id" label="New Password" />,
```
gnapse, line 268:
```suggestion
                <PasswordField label="New Password" />,
```

src/new-components/password-field/password-field.test.tsx
```diff
@@ -260,4 +261,15 @@ describe('PasswordField', () => {
...
+            )
+            const results = await axe(container)
+
+            expect(results).toHaveNoViolations()
+        })
```
gnapse, line 273: I think this could be expanded to also check for a11y violations in password fields that have a `hint` (because this is supposed to add a `aria-describedby` attribute), and maybe test password fields that are disabled.

This feedback here applies to most of the other field components (`SelectField`, `TextField`). I won't repeat the same comment in those components, but whatever you decide to take from this suggestion here should be replicated in those other too.

src/new-components/select-field/select-field.test.tsx
```diff
@@ -193,4 +194,15 @@ describe('SelectField', () => {
         expect(selectElement).toHaveDisplayValue('Light theme')
         expect(screen.getByTestId('container')).toHaveAttribute('data-theme', 'light')
     })
+
+    describe('a11y', () => {
+        it('renders with no a11y violations', async () => {
+            const { container } = render(
+                <SelectField data-testid="select-field" id="custom-id" label="Province" />,
```
gnapse, line 201:
```suggestion
                <SelectField label="Province" />,
```

src/new-components/text-field/text-field.test.tsx
```diff
@@ -182,4 +183,15 @@ describe('TextField', () => {
         expect(inputElement).toHaveValue('Software developer')
         expect(screen.getByTestId('value')).toHaveTextContent('Software developer')
     })
+
+    describe('a11y', () => {
+        it('renders with no a11y violations', async () => {
+            const { container } = render(
+                <TextField label="Whatʼs your name?" hint="We need it for billing purposes" />,
```
gnapse, line 190: Test it without hint, to be consistent with other field tests, and with a hint as well.

stories/components/Dropdown.stories.tsx
```diff
@@ -6,6 +6,9 @@ import Button from '../../src/components/deprecated-button'
 export default {
     title: 'Components/Dropdown',
     component: Dropdown,
+    parameters: {
+        badges: ['notAccessible'],
```
gnapse, line 10: I would deprecate this one as well. I'm pretty certain that we're going to introduce a different component for this purpose.

stories/components/Time.stories.tsx
```diff
@@ -10,6 +10,9 @@ import Time from '../../src/components/time'
 export default {
     title: 'Components/Time',
     component: Time,
+    parameters: {
+        badges: ['accessible'],
+    },
```
gnapse, line 15: Does this one have jest-axe tests already?

stories/components/Tooltip.stories.tsx
```diff
@@ -10,6 +10,9 @@ import type { ButtonProps } from '../../src/components/deprecated-button'
 
 export default {
     title: 'Components/Tooltip',
+    parameters: {
+        badges: ['accessible'],
```
gnapse, line 14: Does this one have jest-axe tests already?

gnapse (approved):
I'm approving, although I think that some of my comments need to be acted on in one way or another. But I'll leave it to you to decide, and no need to wait for a second review after that.

src/new-components/box/box.test.tsx
craigcarlyle: > Is this a formality? For completeness?

Correct.

stories/components/Time.stories.tsx
craigcarlyle: Nope, added.

stories/components/Tooltip.stories.tsx
craigcarlyle: Nope, added.

PR #611, opened 2021-11-16
feat: Updates Avatar to support responsive patterns

### LINKED ISSUE #605: Avatar: Allow responsive sizes
### Motivation

One of the great things about the new components is the responsive nature, being able to set a value for mobile/desktop/tablet. I realise that Avatar is one of the original components, but I think the `Avatar` component could definitely do with it.

### Example

This is currently what I have where `isMobile` is something from [`react-device-detect`](https://www.npmjs.com/package/react-device-detect), but it means it's not truly responsive.
```
<Avatar
    size={isMobile ? 'l' : 'xl'}
    avatarUrl={authenticatedUser.avatarUrl}
    user={{
        name: authenticatedUser.name,
        email: authenticatedUser.emailAddress,
    }}
/>
```
What I'd like to be able to do is
```
<Avatar
    size={{
        mobile: 'l',
        desktop: 'xl',
    }}
    avatarUrl={authenticatedUser.avatarUrl}
    user={{
        name: authenticatedUser.name,
        email: authenticatedUser.emailAddress,
    }}
/>
```

scottlovegrove:
Closes #605 

## Short description

Updates the Avatar component to use the newer way of doing reactist components, using css variables etc. It also adds the abaility to use the responsive elements that have been introduced with v11 so that an avatar's size can dynamically change based on the screen size.

One thing to note, I have intentionally left the size values as is otherwise this would have resulted in a breaking change. 

## PR Checklist

## Versioning

Left versioning alone, it can be released when one is next done.


src/new-components/avatar/avatar.tsx
```diff
@@ -28,20 +30,17 @@ const AVATAR_COLORS = [
...
 type Props = {
     className?: string
     colorList?: string[]
-    size?: AvatarSize
+    size?: ResponsiveProp<AvatarSize>
     avatarUrl?: string
     user: { name?: string; email: string }
 }
 
 function Avatar({ user, avatarUrl, size = 'l', className, colorList = AVATAR_COLORS }: Props) {
```
gnapse, line 41: I'd take this opportunity to refactor it even further towards how the new design system components work or expose their options. Specifically, I'd recommend replacing the `className` prop with one called `exceptionallySetClassName`. This allows to give the signal that the components are meant to be self-styled, and that overriding their styles is a thing to only do exceptionally.

gnapse: I'd also even remove the `colorList` prop, as it goes against styles being self-owned. I checked and it does not seem to be used in Twist or Todoist.

scottlovegrove: > I'd recommend replacing the className prop with one called exceptionallySetClassName.

I did think about that, but that would have also been a breaking change.

scottlovegrove: I'll add a deprecated flag on classname so it's not such a breaking change.

scottlovegrove: > I'd also even remove the `colorList` prop, as it goes against styles being self-owned. I checked and it does not seem to be used in Twist or Todoist.

What should the background colour be then if these are removed? Or do you just mean remove the param but keep the internal list?

gnapse: Deprecating the `className` is fine too. I assume that would be alongside adding the new `exceptionallySetClassName` as the recommended alternative to the deprecated prop.

As for the `colorList`, I was proposing to get rid of the prop only. The internal list stays.

scottlovegrove: > I assume that would be alongside adding the new `exceptionallySetClassName` as the recommended alternative to the deprecated prop.

Yes, and I've added a note to the depcated tag to use `exceptionallySetClassName`.

> The internal list stays.

Okey doke 👍🏻

gnapse: Which, BTW, this is a change that would also breaking. So if you're trying to avoid all breaking changes (which is a good strategy after all), then you might as well not remove the prop, and simply set it as deprecated as well.

scottlovegrove: Good point 👍🏻

src/new-components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+        expect(avatar).toMatchSnapshot()
+    })
+
+    it('renders initials of user name when avatarUrl is not supplied', () => {
+        render(getAvatar())
+
+        const avatar = screen.getByTestId('avatar')
+
+        expect(avatar).toMatchSnapshot()
```
gnapse, line 20: What's the value of these calls to `toMatchSnapshot`? You are already validating via another assertion below what the test case description says (renders initials).

src/new-components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        expect(avatar.textContent).toBe('HM')
```
gnapse, line 21: Can you try with this custom matcher and see if it works?

```suggestion
        expect(avatar).toHaveTextContent('HM')
```

src/new-components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+    })
+
+    it('renders initials of user email when avatarUrl is not supplied', () => {
+        render(getAvatar({ user: { email: 'henning@doist.com' } }))
+
+        const avatar = screen.getByTestId('avatar')
+
+        expect(avatar).toMatchSnapshot()
+        expect(avatar.textContent).toBe('H')
```
gnapse, line 39: Same observations about there being no need for the snapshot, and the use of `toHaveTextContent`.

src/new-components/avatar/avatar.test.tsx
```diff
@@ -0,0 +1,65 @@
...
+        expect(avatar.textContent).toBe('HM')
+    })
+
+    it('renders initials on custom color', () => {
+        render(getAvatar({ colorList: ['red', 'green', '#0000FF'] }))
+
+        const avatar = screen.getByTestId('avatar')
+
+        expect(avatar).toMatchSnapshot()
+        expect(avatar.textContent).toBe('HM')
```
gnapse, line 30: The snapshot already validates the initials, plus this test is not about the initials, but about the colors. So I'd remove the `textContent` assertion.

src/new-components/avatar/avatar.tsx
```diff
@@ -52,10 +60,16 @@ function Avatar({ user, avatarUrl, size = 'l', className, colorList = AVATAR_COL
...
           }
 
+    const sizeClassName = getClassNames(styles, 'size', avatarSize)
+
     return (
-        <div className={avatarClass} style={style}>
+        <Box
+            className={[className, styles.avatar, sizeClassName, exceptionallySetClassName]}
+            style={style}
+            data-testid="avatar"
```
gnapse, line 69: I recommend we do not include in the final component something that's there only to support the internal tests in Reactist.

I recommend that you forward any extra props received by the avatar to the box. You may need some TypeScript tweaking, in which case let me know if you need help. But maybe you can get away with merely adding the `...props` to the Avatar props deconstruction and then forwarding that as `<Box myPropsFirst {...props}>`

Then you pass this `data-testid` in the test only.

gnapse (changes requested):

src/new-components/avatar/avatar.test.tsx
scottlovegrove: Not a clue, I didn't add them, they were already present, so I just left them there.

src/new-components/avatar/avatar.test.tsx
scottlovegrove: This test is actually for the `colorList` prop which we've just removed, so I guess I should just remove this test anyway.

gnapse:
@scottlovegrove you marked my latest comments as resolved, but I do not see any new changes having been pushed after them. Perhaps you forgot to push something?

scottlovegrove:
> @scottlovegrove you marked my latest comments as resolved, but I do not see any new changes having been pushed after them. Perhaps you forgot to push something?

@gnapse I was waiting for an answer about the snapshot tests and whether I should just remove them.

gnapse:
I do not see value on them in this case, especially in the cases where I commented (the ones where you are already making an extra assertion that validates what's described in the test case title).

scottlovegrove:
> I do not see value on them in this case, especially in the cases where I commented (the ones where you are already making an extra assertion that validates what's described in the test case title).

I have removed them then.

gnapse (approved):

PR #614, opened 2021-12-05
fix: Improve ModalActions behaviour with null child items

### LINKED ISSUE #613: ModalActions: Don't add spacing for null child items
### Current behavior

I'm creating a wizard within a `Modal` component and making use of the `ModalActions` feature, but the buttons being displayed are dynamic in their nature depending on which step of the wizard you are at. 

My code for the `ModalActions` looks like this
```jsx
<ModalActions>
    {buttonData.back ? (
        <Button variant="secondary" onClick={backClicked}>
            <>{t('addLeague.back')}</>
        </Button>
    ) : null}
    {buttonData.next ? (
        <Button variant="primary" disabled={busy} type="submit" loading={busy}>
            <>{t('addLeague.next')}</>
        </Button>
    ) : null}
    {!buttonData.next && !buttonData.back ? (
        <Button variant="primary" onClick={close}>
            <>{t('addLeague.close')}</>
        </Button>
    ) : null}
</ModalActions>
```

At some point, some of these buttons won't be displayed and will instead return `null`, but they still seem to become children for the `ModalActions` which results in empty divs being created, and the default spacing (large) applied to those empty divs, which results in the following 
image

Create a `Modal`, set the `ModalActions` as follows:

```jsx
<ModalActions>
    {null}
    {null}
    {null}
</ModalActions>
```

Then inspect the resulting html, it will give you 
image

### Expected behavior

It should not be adding empty divs when the child is null

### Possible solutions

https://github.com/Doist/reactist/blob/main/src/new-components/modal/modal.tsx#L361-L363

This part could filter out any children that are null/undefined

### Environment

-   `@doist/reactist` version: v11.0.0
-   `react` version: 17.0.2
-   Browser: Chrome

scottlovegrove:
Closes #613 

## Short description

This change ensures that if child items passed into the `ModalActions` component are null, then they do not have an empty div, with spacing, created. It also ensures that if all child items passed into the component are null, then just don't create the footer in the first place.

## PR Checklist

## Versioning

This is not a breaking change, but I haven't updated the version.


src/new-components/modal/modal.tsx
```diff
@@ -355,12 +355,14 @@ export function ModalFooter({
...
+
+    // If all child items are null/undefined, then don't show the footer at all and alleviate any
+    // excess spacing within the modal.
+    return childItems.length === 0 ? null : (
         <ModalFooter {...props}>
             <Inline align="right" space="large">
-                {React.Children.map(children, (child) => (
-                    <div>{child}</div>
-                ))}
+                {childItems.map((child, i) => (child ? <div key={i}>{child}</div> : null))}
```
henningmu, line 365: I don't think we need to check here for falsy values as we filter them out above already. All `child` should be truthy 👍

---

A question for @gnapse: why do we need to wrap the children in `div`s in the first place? Shouldn't `Inline` work with anything we throw at it? And if not should we pass this burden to the user?

scottlovegrove: > I don't think we need to check here for falsy values as we filter them out above already. All child should be truthy 👍

Say hello to the original fix 😄 I'll remove that check now.

gnapse: > why do we need to wrap the children in `div`s in the first place? Shouldn't `Inline` work with anything we throw at it? And if not should we pass this burden to the user?

Nope. This was a compromise to avoid having `Inline` and `Stack` generate excessive markup all the time. See https://github.com/Doist/reactist/wiki/Simplified-markup-of-the-Stack-and-Inline-components

gnapse (changes requested):
Thanks for this, Scott. It certainly is an oversight.

However, this has made me realize that there's another related oversight that we might as well fix alongside the original issue. Imagine something like this:

```jsx
<ModalActions>
  {canSubmit ? (
    <>
      <Button type="submit">Delete one</Button>
      <Button type="submit">Delete all</Button>
    </>
  ) : null}
  <Button>Cancel</Button>
</ModalActions>
```

The first two buttons, being wrapped by a fragment, will be part of a single child of the `Inline` wrapper, thus lacking a spacing between them. What a consume of this component would expect is that the above `ModalAction` is treated as a `Inline` with 3 spaced-out children when the `canSubmit` condition is `true`.

A better fix for this issue goes along the lines of using `react-keyed-flatten-children`. An example of this is present in [the `Stack` component implementation](https://github.com/Doist/reactist/blob/3dd6bdfd1e4c150f87306838559fae9f321bdd96/src/new-components/stack/stack.tsx#L58).

Since Scott is on time off for a few weeks, I may take this over and do the change myself.

src/new-components/modal/modal.tsx
henningmu: Makes sense and awesome that we have documentation on this ❤️

gnapse:
Closed without merging, superseded by #615.

PR #615, opened 2021-12-07
fix: Various fixes to the `ModalActions` component

### LINKED ISSUE #613: ModalActions: Don't add spacing for null child items

gnapse:
Closes #613
Supersedes PR #614 

## Short description

This PR performs the following changes on the `ModalActions` component:

- ignore `null` children
- flatten children wrapped in React fragments
- removes the unneeded extra div wrapping each child element (following the same logic described [here](https://github.com/Doist/reactist/wiki/Simplified-markup-of-the-Stack-and-Inline-components)).

Additionally, this removes an extra `Box` wrapping the main element rendered by `Inline`. I was surprised to see it, and did several manual testing (both in the storybook, and running twist-web with my local version of Reactist with these changes in place) and, as I suspected, this outer `div` is not needed.

## PR Checklist

## Versioning

Queued, to be included in an upcoming release.

### DIFF
--- a/src/new-components/modal/modal.test.tsx
+++ b/src/new-components/modal/modal.test.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react'
 import { render, screen, within } from '@testing-library/react'
-import { Modal, ModalHeader, ModalFooter, ModalBody, ModalCloseButton } from './modal'
+import { Modal, ModalHeader, ModalFooter, ModalActions, ModalBody, ModalCloseButton } from './modal'
 import userEvent from '@testing-library/user-event'
 import { axe } from 'jest-axe'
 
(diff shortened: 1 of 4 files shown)


src/new-components/modal/modal.test.tsx
```diff
@@ -189,6 +189,86 @@ describe('ModalFooter', () => {
...
+        rerender(<ModalActions withDivider>Hello</ModalActions>)
+        expect(screen.getByRole('separator')).toBeInTheDocument()
+    })
+
+    it('ignores null children', () => {
+        render(
+            <ModalActions data-testid="modal-actions">
+                <button>OK</button>
+                <button>Cancel</button>
+                {null}
```
nats12, line 252: Wouldn't this make it harder to debug `null` values as a consumer? Say we are passing down a `div` containing a `null` value. We would at least be able to see that the `div` is correctly being rendered, it just doesn't have a value. Not against it though, just wanted more regarding the decision to ignore them

nats12 (approved):
Looks good! Just left a question around the ignoring of `null` children, but nothing that's blocking

src/new-components/modal/modal.test.tsx
gnapse: > Say we are passing down a `div` containing a `null` value.

What do you mean with "a `div` containing a `null` value"? And what do you mean with "debug `null` values as a consumer"?

nats12: What I understand from the above is that if we have:

```
<ModalActions data-testid="modal-actions">
                <button>OK</button>
                <button>Cancel</button>
                <button>{some null value</button>
</ModalActions>
```

We would see:

```
<ModalActions data-testid="modal-actions">
                <button>OK</button>
                <button>Cancel</button>
 </ModalActions>
```

Is that right? Or would we render an empty button tag?

> And what do you mean with "debug null values as a consumer"?

If I've understood it right and the above example is how it would work, wouldn't it be confusing to eliminate the empty button tag with the `null` value? Especially as we inspect the DOM to see why it hasn't rendered?

gnapse: What the above example shows is that if one of the **direct** children of `ModalActions` is `null`, it will not generate a "slot" in the sequence of inlined elements that are spaced out. It has nothing to do with having non-null children that have their children `null`.

If this component that we're testing receives a `<button>{some null value}</button>`, it has no way of knowing that the component has null children. It only knows if its own children are null.

henningmu:
@gnapse let's merge this?

gnapse:
> @gnapse let's merge this?

Thanks for the reminder, I had forgotten about this.

I'll do a smoke test running Twist locally against this branch, to make sure that it does not break anything. I think I did this at some point, but I'm not 100% sure.

gnapse:
I found no major consequences. The only noticeable thing is that under certain situations (e.g. a `Inline` directly inside a `Stack`) we may need to manually wrap the `Inline` inside a `div` or `Box`. I only found one instance of this happening in Twist, and in any case the actual visual consequence is almost not noticeable.

CleanShot 2022-01-04 at 11 36 48

I'll fix it in Twist in the PR that will upgrade to the new Reactist version, which I'll create this week.

PR #620, opened 2022-01-13
ci: set sensible timeout

pkoch:
Based on historical data, clamped between 5 and 60.

See https://github.com/Doist/Issues/issues/5259


.github/workflows/codeql-analysis.yml
```diff
@@ -1,56 +0,0 @@
-name: 'CodeQL'
```
pkoch, line 1: @henningmu Did you mean to delete this?

henningmu: Yup, it didn't run for 7 months anyway, cleaning the house a little 🧹

PR #633, opened 2022-02-15
fix: Hide Modal Header close button column when the button is not rendered

pedroalves0:
## Short description

This PR doesn't implement any functional change, only updates the Modal Header layout. The problem that it solves is that when the close button is hidden (its prop set `false`/`null`/`undefined`), the button column is rendered, although the button isn't. The column follows the parent `Columns` component spacing specification and renders a `large` spacing, even though there's isn't actual content rendered inside it.

The following screenshot exemplifies the problem, marking the extra column in a different color.

<a href="https://codesandbox.io/s/clever-yalow-675wis">
  extra-spacing
</a>

[^1]

[^1]: https://codesandbox.io/s/clever-yalow-675wis

The solution is to not render the column when the button is hidden.

## PR Checklist

-   [X] Added tests for bugs / new features
-   [X] Executed `npm run validate` and made sure no errors / warnings were shown
-   [X] Described changes in `CHANGELOG.md`

## Versioning

Won't be deployed for now. To be included in an upcoming release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 
 # next
 
+-   [Fix] Hidding `ModalHeader` close button no longer renders the button wrapper column
 -   [Fix] Using `start` as a value for `Box`'s `textAlign` now adds a class to properly set the `text-align` value
 -   [Fix] `TextArea`'s `rows` prop is not added to the component's type definition
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 
 # next
 
--   [Fix] Hidding `ModalHeader` close button no longer renders the button wrapper column
+-   [Fix] Hiding `ModalHeader` close button no longer renders the button wrapper column
 -   [Fix] Using `start` as a value for `Box`'s `textAlign` now adds a class to properly set the `text-align` value
 -   [Fix] `TextArea`'s `rows` prop is not added to the component's type definition
 
--- a/src/new-components/modal/modal.tsx
+++ b/src/new-components/modal/modal.tsx
@@ -248,13 +248,15 @@ export function ModalHeader({
             >
                 <Columns space="large" alignY="center">
                     <Column width="auto">{children}</Column>
-                    <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
-                        {typeof button !== 'boolean' ? (
-                            button
-                        ) : button === true ? (
-                            <ModalCloseButton aria-label="Close modal" autoFocus={false} />
-                        ) : null}
-                    </Column>
+                    {button === false || button === null || button === undefined ? null : (
+                        <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
+                            {typeof button === 'boolean' ? (
+                                <ModalCloseButton aria-label="Close modal" autoFocus={false} />
+                            ) : (
+                                button
+                            )}
+                        </Column>
+                    )}
                 </Columns>
             </Box>
             {withDivider ? <Divider /> : null}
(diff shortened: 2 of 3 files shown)


src/new-components/modal/modal.tsx
```diff
@@ -248,13 +248,15 @@ export function ModalHeader({
...
                     <Column width="auto">{children}</Column>
-                    <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
-                        {typeof button !== 'boolean' ? (
-                            button
-                        ) : button === true ? (
-                            <ModalCloseButton aria-label="Close modal" autoFocus={false} />
-                        ) : null}
-                    </Column>
+                    {button === false || button === null || button === undefined ? null : (
+                        <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
```
gnapse, line 252: Could you add a test about this in the `modal.test.tsx` file (under the `describe('ModalHeader',` section maybe).

I'm not sure how to check this in the test, though. Specifically the part about how to check if this `Column` container is rendered, since it's just a div, it can be hard to address.

I suspect that in jest tests the CSS modules class names are used as is. So this container should have `class="buttonContainer"` in the resulting HTML.

```js
expect(document.querySelector('.buttonContainer')).not.toBeInTheDocument()
```

You can also add a `data-testid` attribute if that makes it easier.

gnapse (approved):
This is awesome, @pedroalves0. Thanks for contributing!

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 
 # next
 
+-   [Fix] Hidding `ModalHeader` close button no longer renders the button wrapper column
```
frankieyan, line 7:
```suggestion
-   [Fix] Hiding `ModalHeader` close button no longer renders the button wrapper column
```

src/new-components/modal/modal.tsx
```diff
@@ -248,13 +248,15 @@ export function ModalHeader({
...
                 <Columns space="large" alignY="center">
                     <Column width="auto">{children}</Column>
-                    <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
-                        {typeof button !== 'boolean' ? (
-                            button
-                        ) : button === true ? (
-                            <ModalCloseButton aria-label="Close modal" autoFocus={false} />
-                        ) : null}
-                    </Column>
+                    {button === false || button === null || button === undefined ? null : (
```
frankieyan, line 251: Can this be a single falsy check instead?

src/new-components/modal/modal.tsx
```diff
@@ -248,13 +248,15 @@ export function ModalHeader({
...
+                        <Column width="content" exceptionallySetClassName={styles.buttonContainer}>
+                            {typeof button === 'boolean' ? (
```
frankieyan, line 253: It won't make any functional difference, but since we've already eliminated false (or falsy) values from `button`, would checking for `true` explicitly make more sense? Otherwise it could read like we want to render the button even when it's false.

frankieyan (approved):
Just a few minor comments 👍

src/new-components/modal/modal.tsx
pedroalves0: I opted not to use it. The reason is that `0` should be a valid button content.

I could have written `button === false || button == null`, but then it triggers an eslint rule, so I wrote the full statement.

src/new-components/modal/modal.tsx
pedroalves0: In my opinion, testing for `true` would be redundant and less semantic. Above we do a truthy/falsy check and the second condition does a type check (although it's true that testing `true` would match the type too). I think that's more readable than doing another boolean comparison.

src/new-components/modal/modal.tsx
pedroalves0: I previously skimmed the test suite for this component and was under the impression there was a test case that tested the hidden button, but I was wrong.

I've added the test case as per your suggestion.

src/new-components/modal/modal.tsx
gnapse: Indeed. Technically it is. That being said, technically I do not mind if we'd overlook that. Another reason for a falsy check would be to also consider the empty string as there being no button to show, which IMO is valid. But I have no strong opinion about having to do something about it right now.

PR #634, opened 2022-02-23
feat: Allow columns to customize their height

gnapse:
## Short description

One issue I found while creating the demo for my Doistalk about the design system components. Having all columns to automatically fill in the entire height is not desirable. It can cause issues when used alongside `<Columns alignY="center" />`. Take the following scenario:

```jsx
<Columns alignY="center">
  <Column><Placeholder height="100" /></Column>
  <Column><Placeholder height="200" /></Column>
</Columns>
```

Since the columns expand their height fully, the `alignY` ceases to have the desired effect, because the columns are all the same height. However, their content will move towards the top, thus not being aligned vertically towards the centre. This can be seen below. The placeholders are not of the same height, so they should align by the centre. However, since the columns are filling in the full height, the `alignY="center"` effect is lost. The columns shield the placeholder elements from aligning vertically as intended.

image

This is currently the cause of misaligned content in the Twist settings:

image

## PR Checklist

## Versioning

Not urgent. To be merged and queued for an upcoming release to include it.

### DIFF
--- a/src/new-components/columns/columns.tsx
+++ b/src/new-components/columns/columns.tsx
@@ -23,11 +23,12 @@ type ColumnWidth =
     | '4/5'
 
 interface ColumnProps {
+    height?: 'full' | 'content'
     width?: ColumnWidth
 }
 
 const Column = polymorphicComponent<'div', ColumnProps>(function Column(
-    { width = 'auto', children, exceptionallySetClassName, ...props },
+    { width = 'auto', height = 'content', children, exceptionallySetClassName, ...props },
     ref,
 ) {
     return (
@@ -42,7 +43,7 @@ const Column = polymorphicComponent<'div', ColumnProps>(function Column(
             ]}
             minWidth={0}
             width={width !== 'content' ? 'full' : undefined}
-            height="full"
+            height={height === 'full' ? 'full' : undefined}
             flexShrink={width === 'content' ? 0 : undefined}
             ref={ref}
         >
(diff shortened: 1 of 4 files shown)


nats12 (dismissed):
Makes sense, LGTM 👍🏼

src/new-components/columns/columns.tsx
```diff
@@ -23,11 +23,12 @@ type ColumnWidth =
     | '4/5'
 
 interface ColumnProps {
+    height?: 'full' | 'content'
     width?: ColumnWidth
 }
 
 const Column = polymorphicComponent<'div', ColumnProps>(function Column(
-    { width = 'auto', children, exceptionallySetClassName, ...props },
+    { width = 'auto', height = 'content', children, exceptionallySetClassName, ...props },
```
frankieyan, line 31: @gnapse I wonder if we should default to `full` to retain the previous behaviour? I'm not sure whether we have use cases that relied on the columns to be equal in height, but it may not be easy to discover them all across Todoist and Twist.

gnapse: @frankieyan ignore this PR for now and do not release it. I need to think more about it. I'm inclined to think that we may not need this, and this should be solved seamlessly by the system.

There's no rush to make a decision on this, as it is solving an issue that I myself raised, and there's no user reports on it.

gnapse: @frankieyan can you please add a review requesting changes? I cannot do it as the PR author.

frankieyan (changes requested):
Requesting changes as you requested :)

gnapse:
I'm closing this one for now, as we're not sure of the need for it, and there has not been any recent issues related to this. We can revisit and try this again in the future if needed.

PR #619, opened 2022-01-12
fix: Render menu lists in a React portal by default

gnapse:
## Short description

This PR changes menu popovers so that, by default, they render in a React portal, rather than by the side of the trigger button position in the DOM.

The change in this PR was motivated by a specific issue in Twist (https://github.com/Doist/Issues/issues/4239), where the fix is to make the menu render in a portal. This can be achieved in app code without changing Reactist, by passing the `modal={true}` prop to the top-level `Menu` (i.e. `<Menu modal={true}>`).

However, I think that it is saner to make this the default, as in general, floating elements are better off being rendered near the root of the DOM. Among the reasons, for instance, is the fact that while in a deeper part of the DOM, they could inherit unwanted CSS that affects the way they look.

From [reakit's menu component documentation](https://reakit.io/docs/menu/), these below are the implications of rendering menus in this way, that are IMO also ok and even desired:

> `preventBodyScroll` is automatically enabled, focus is trapped within the dialog and the dialog is rendered within a `Portal` by default.

BTW, this change is not preventing consumers of our menu component from reverting to a non-portal menu if they so wish. They can always do so via `<Menu modal={false}>`.

Last, but not least: doing this required upgrading the Reakit version to the latest, because the menu positioning while in this portal mode was not working correctly, but in the latest version it does.

## Test plan

1. Run storybook
    - [x] see that menus work as usual.

2. Now run twist-web while linked to Reactist locally, following these steps:

    ```shell
    # In reactist's folder under this branch
    npm install
    npm run start # leave it running

    # in another terminal, in twist-web folder under the main branch
    ide twist-web run --link-reactist
    ```

    Now check the following:

    - [ ] Try out various menus all around Twist and see that they work.
    - [ ] Make sure that menus are still rendered floating nearby the button that triggered them.
    - [x] Make sure in the HTML inspector dev tool that the menu is now rendered in a portal div that's mounted right inside the `<body>` element at the bottom (see https://cln.sh/nyXpWq)
    - [ ] Try specifically the menus in the thread list. Make sure that clicking on the outside of the menu to dismiss it does not trigger navigation (see [Valente's description below about this issue](https://github.com/Doist/reactist/pull/619#issuecomment-1012086512), that he experienced during his initial review of this PR).
    - [x] Check that the Firefox-specific bug described in https://github.com/Doist/Issues/issues/4239 is no longer happening when using this setup.

## PR Checklist

-   [ ] Added tests for bugs / new features
-   [ ] Updated docs (storybooks, readme)
[...]

PR #635, opened 2022-02-23
fix: Change how the menu and its backdrop are positioned in the markup

gnapse:
Supersedes #619
Related to fixing https://github.com/Doist/Issues/issues/4239

## Short description

This PR performs a change very similar to what I already attempted in #619, but with an added change that apparently contributes to the fix: rendering the menu backdrop (overlay) not as a parent of the menu, but as a sibling instead.

So while #619 attempted to do something like this (tag names are for reference, but they're all divs in reality):

```markup
<react-portal>
  <backdrop-div>
    <menu-popover>…</menu-popover>
  </backdrop-div>
</react-portal>
```

This PR does it like this

```markup
<react-portal>
  <backdrop-div />
</react-portal>
<react-portal>
    <menu-popover>…</menu-popover>
</react-portal>
```

## Test plan

1. Run storybook
    - [ ] see that menus work as usual.

2. In twist-web, remove [this CSS](https://github.com/Doist/twist-web/blob/9daaf17d0094326e0d02af0d46a9502b46af4844/src/components/styles/reactist.less#L255-L257) (which will have to be removed in the PR that ends up upgrading Reactist wit this change in place)

3. Now run twist-web while linked to Reactist locally, following these steps:

    ```shell
    # In reactist's folder under this branch
    npm install
    npm run start # leave it running

    # in another terminal, in twist-web folder under the main branch
    ide twist-web run --link-reactist
    ```

4. Now check the following:

    - [ ] Try out various menus all around Twist and see that they work. Specifically, try to trigger menus that might break due to z-index issues (e.g. the thread view header menus when the view header is already "lifted" on top of the scrolling content, things like that). In general, try to see if you can break any menu in twist-web.
    - [ ] Make sure that menus are still rendered floating nearby the button that triggered them.
    - [ ] Make sure in the HTML inspector dev tool that the menu is now rendered in a couple of portal div that are mounted right inside the `<body>` element at the bottom (see https://cln.sh/nyXpWq)
    - [ ] Try specifically the menus in the thread list (each thread's options menu, ). Make sure that clicking on the outside of the menu to dismiss it does not trigger navigation (see [Valente's description below about this issue](https://github.com/Doist/reactist/pull/619#issuecomment-1012086512), that he experienced during his initial review of this PR).
    - [ ] Check that the Firefox-specific bug described in https://github.com/Doist/Issues/issues/4239 is no longer happening when using this setup.

## Versioning

Patch release (bug fix, no new features or breaking changes).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,8 @@
                 "react-focus-lock": "^2.5.2",
                 "react-keyed-flatten-children": "^1.3.0",
                 "react-markdown": "^5.0.3",
-                "reakit": "1.3.0"
+                "reakit": "1.3.11",
+                "reakit-utils": "^0.15.2"
             },
             "devDependencies": {
                 "@babel/core": "^7.9.6",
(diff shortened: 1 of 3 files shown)


gnapse:
Ok, so I'm giving up. Just after posting this PR I found out that this is still "flaky". See the video below:

https://user-images.githubusercontent.com/15199/155425837-0f40b333-471f-4c84-8654-062a7cdf8049.mp4

Sometimes, for some reason, the menus appear in the top left corner, instead of floating right next to the menu button. It seems totally random.

I'm not sure what else to do beyond re-implementing our menus with something other than reakit.

gnapse:
No longer needed. The original issue that led to this was addressed differently anyway (rendering the button outside the surrounding link, and using CSS to make it look like it's inside).

PR #636, opened 2022-02-28
feat: Allow buttons to customize tooltip `gapSize`

rfgamaral:
## Short description

Exposes the  `gapSize` prop from `<Tooltip>` as a `tooltipGapSize` prop in `<Button>`. This is required by the new editor toolbar:

image

We need to push the tooltip further up a bit.

## PR Checklist

## Versioning

New feature required for the new editor toolbar, to be released as soon as it's merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 -   [Fix] Hiding `ModalHeader` close button no longer renders the button wrapper column
 -   [Fix] Using `start` as a value for `Box`'s `textAlign` now adds a class to properly set the `text-align` value
 -   [Fix] `TextArea`'s `rows` prop is not added to the component's type definition
+-   [Feat] Updates `Button` to support `gapSize` from `Tooltip` (as `tooltipGapSize`)
 
 # v11.1.0
 
--- a/src/new-components/base-button/base-button.tsx
+++ b/src/new-components/base-button/base-button.tsx
@@ -47,6 +47,10 @@ type CommonProps = {
      * A tooltip linked to the button element.
      */
     tooltip?: TooltipProps['content']
+    /**
+     * The distance between the button element and the linked tooltip.
+     */
+    tooltipGapSize?: TooltipProps['gapSize']
 }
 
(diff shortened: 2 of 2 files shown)


frankieyan (approved):

gnapse:
I want to challenge this feature. The rationale behind having the convenient `tooltip="…"` prop in the `Button` is that's just for convenience. For the default use case. Once you need to pass extra props to the tooltip, you should go back to use the tooltip directly. Nothing prevents you from doing this:

```jsx
import { Button, Tooltip } from '@doist/reactist'

function View() {
  return (
    <Tooltip content="Hello" gapSize={8}>
        <Button variant="primary">Click me</Button>
    </Tooltip>
  )
}
```

The reason for me to object to this is that this is a slippery slope. `Tooltip` has other props. What do we do? Do we add them all to the button to be a proxy that forwards props to the tooltip?

rfgamaral:
@gnapse I didn't quite remember if I tried the approach you are suggesting above, but it's not working for me when `<Button>` is an icon-only button. An icon-only button requires an `aria-label`, and that is taking precedence over the wrapped `<Tooltip>`.

I agree with your reasoning, so if that can be fixed as well, I'm okay with using the suggested approach.

gnapse:
Indeed, that's a very good point. I'll see how we can fix this without having to forward all props down.

gnapse:
I'm going to make it so that consumers will be able to suppress the implicit icon-only tooltip by explicitly passing down `tooltip={null}`:

```jsx
<Tooltip content="Add task now">
    <Button
        variant="primary"
        aria-label="Add task"
        icon={<Icon />}
        tooltip={null}
    />
</Tooltip>
```

Sounds good?

rfgamaral:
> Sounds good?

Perfect 👌

---

I guess I can close the draft PR I opened, since it can't be merged without that fix, which is coming in #649, right? And I'm assuming you will take care of updating the apps with the breaking release that will come out of that PR.

gnapse:
> I'm assuming you will take care of updating the apps with the breaking release that will come out of that PR.

Yes, I will.

rfgamaral:
Ok, cool. Thanks 🙏

PR #638, opened 2022-03-04
Adds button CSS border colour variables

craigcarlyle:
## Short description

This pull request adds CSS border colour variables for the `Button` component:

Screen Shot 2022-03-03 at 4 22 42 PM

This is to handle styling usecases like the `Cancel` buttons in Todoist:

Screen Shot 2022-03-03 at 4 31 20 PM

## PR Checklist


gnapse:
Is there a Figma file for how all the button variants look like in Todoist? Do we need border for them all? I'd check up on that before judging the changes in this PR on its own.

craigcarlyle:
@gnapse I don't think we have Figma mockups for individual button styles (correct me if I'm wrong, @alexmuench) but the Figma mockups for the Task View show a border on the `normal`/`secondary` button at the very least.

I could reduce this PR to only apply to that button if you prefer, but I figured it'd be best to add this to every style for consistency (and in case anyone inside or outside of Doist ever needed the functionality).

gnapse:
> but the Figma mockups for the Task View show a border on the `normal`/`secondary` button at the very least.

Can you point me to that Figma file? (via DM or via a thread in Twist) I'd really want to have a clear picture of what's needed from our buttons from a high-level before jumping in on ad-hoc changes in the library. I was under the impression (hope?) that our buttons across our two products would be more homogeneous, but seeing that in Todoist we plan for them to have borders, while they do not have borders in Twist, worries me a bit. Maybe we can continue to discuss in a thread.

alexmuench:
Ah yes! We definitely should unify. It looks like we are still using many outdated buttons in Todoist. I was under the impression making such a fundamental change would be hard to achieve, so I just followed through with what we have. Looks like the correct button style got lost in translation.

Ideally, yes, both products use the same button styles (with other colors of course).

I just checked our Design [Product Web library](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product---Web?node-id=3046%3A81842) and you can see [buttons for Todoist here](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product---Web?node-id=1286%3A25656) and [buttons for Twist here](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product---Web?node-id=3046%3A81842).

Looks like Todoist's secondary buttons still use a border, but I don't think we need them. The fill is supposed to be similar to Twist but with a neutral grey.

**Before-after Todoist:**
image


**Compare to Twist:**
image

craigcarlyle:
Thanks @alexmuench! I'll close this off then.

PR #640, opened 2022-03-22
feat: Update Heading component to support a semibold weight value

nats12:
## Short description

Adds `semibold` to the list of accepted values for the `Heading` component's `weight` prop.

## PR Checklist

## Versioning

New heading weight required for the new [guided onboarding screen](https://www.figma.com/file/oLzMM2zBFGmjIpo4X8vo5u/TD-Onboarding---Retention-DO?node-id=298%3A70262), to be released as soon as it's merged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 
 # Next
 
--   n/a
+-   [Feat] Updates `Heading` to support `semibold` weight prop value.
 
 # v11.2.0
 
--- a/src/new-components/heading/heading.module.css
+++ b/src/new-components/heading/heading.module.css
@@ -7,6 +7,9 @@
 .weight-light {
     font-weight: var(--reactist-font-weight-regular);
 }
+.weight-semibold {
+    font-weight: var(--reactist-font-weight-medium);
+}
 
 /* tone */
 
--- a/src/new-components/heading/heading.stories.tsx
+++ b/src/new-components/heading/heading.stories.tsx
@@ -99,7 +99,7 @@ export function ResponsiveHeadingStory(props: React.ComponentProps<typeof Headin
 ResponsiveHeadingStory.argTypes = {
     level: select(['1', '2', '3', '4', '5', '6'], '1'),
     size: selectWithNone(['largest', 'larger', 'smaller'], 'none'),
-    weight: select(['regular', 'light'], 'regular'),
+    weight: select(['regular', 'light', 'semibold'], 'regular'),
     lineClamp: selectWithNone([1, 2, 3, 4, 5], 'none'),
     tone: select(['normal', 'secondary', 'danger'], 'normal'),
     align: { control: false },
@@ -120,7 +120,7 @@ export function HeadingPlaygroundStory(props: React.ComponentProps<typeof Headin
 HeadingPlaygroundStory.argTypes = {
     level: select(['1', '2', '3', '4', '5', '6'], '1'),
     size: selectWithNone(['largest', 'larger', 'smaller'], 'none'),
-    weight: select(['regular', 'light'], 'regular'),
+    weight: select(['regular', 'light', 'semibold'], 'regular'),
     lineClamp: selectWithNone([1, 2, 3, 4, 5], 'none'),
     tone: select(['normal', 'secondary', 'danger'], 'normal'),
     align: selectWithNone(['start', 'center', 'end', 'justify'], 'none'),
--- a/src/new-components/heading/heading.test.tsx
+++ b/src/new-components/heading/heading.test.tsx
@@ -82,13 +82,19 @@ describe('Heading', () => {
             const textElement = screen.getByTestId('heading-element')
             expect(textElement).not.toHaveClass('weight-regular')
             expect(textElement).not.toHaveClass('weight-light')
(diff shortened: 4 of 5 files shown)


gnapse:
This is for the "thank you for your answers" text in that Figma link you shared, right? If so, I'm hesitant to consider that a `Heading` element. That's more of a `Text` element. Can't you achieve the desired size, weight, etc. with a `<Text />`?

nats12:
> This is for the "thank you for your answers" text in that Figma link you shared, right?

@gnapse Correct :)


> That's more of a Text element. Can't you achieve the desired size, weight, etc. with a <Text />?

I tried this after Frankie suggested the same in a recent PR ([ref](https://github.com/Doist/todoist-web/pull/4181#discussion_r831987978)) but I wouldn't be able to get the same size here as the max value `Text` accepts is `subtitle` which is `16px`. The mocks set it to `20px`. Since 20 is what we use as `Heading` text I saw it fit to use this one.

gnapse:
> I wouldn't be able to get the same size here as the max value `Text` accepts is `subtitle` which is `16px`. The mocks set it to `20px`. Since 20 is what we use as `Heading` text I saw it fit to use this one.

I would then ask the designers, why we have a text element that has a size that's not supported in the documented sizes that text elements can have. This is a prime example of a situation where we should push back instead of simply trying to achieve a desired visual outcome.

nats12:
@gnapse do we know what designers use to refer to our design system components' values? I.e. are they referring to something other than the Reactist site? Perhaps a paper?

gnapse:
> @gnapse do we know what documentation designers use to refer to our design system components' values? I.e. are they referring to something other than the Reactist site? Perhaps a paper?

I'd assume that their source of truth is on Figma. Not sure exactly where. That's why we should ask them. Maybe we can ping the @Doist/design-hero right here.

@Doist/design-hero can you point us to a Figma file or some other source that documents our headers typography styles in the design system?

nats12:
Could be this one @gnapse - I found it under Product Libraries. It's the only one I could find specific to text styles. According to this file, the header in question from the o[nboarding mocks](https://www.figma.com/file/oLzMM2zBFGmjIpo4X8vo5u/TD-Onboarding---Retention-DO?node-id=298%3A70262) (`Header 3` under the inspect column) documents a `semibold` weight. So does `Header 2`. In this case, should we not align Reactist with this file and include it?

gnapse:
@nats12 I'd really want to wait until we have a word from Design about what's the entire set of possible styles that a heading should support, before adding things on Reactist. For the time being, if this is blocking you, go ahead with a `exceptionallySetClassName` prop. Would that work for you?

nats12:
> I'd really want to wait until we have a word from Design about what's the entire set of possible styles that a heading should support, before adding things on Reactist

Of course, makes sense. I realise my question sounded like we should unify missing styles, if any, right away. It was more a 'we should think about doing this' if indeed design confirm that those are the supported styles. But in the meantime, yes a classname prop would work 👍🏼

benbreckler:
Hey! Sorry for the delay here. All our latest text styles can be found in the [Figma Global Library](https://www.figma.com/file/xo9yAsH8PQUpi0eTJh9pmR/Product---Global?node-id=2527%3A3732). Design and Frontend might not be aligned on the same system yet but that's what we are striving to achieve long term. 

@tsamoudakis has also been working on updating the [Todoist Web Libraries](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/?node-id=1080%3A0). You can also check out the Web Templates for reference.

nats12:
Just realised I never linked to it in my last comment 😅 but when I said "this one", that's the exact one I found ([ref](https://www.figma.com/file/xo9yAsH8PQUpi0eTJh9pmR/Product---Global?node-id=2527%3A3732)).  Thanks for confirming @benbreckler 👍🏼

nats12:
@gnapse design has now confirmed that the element from the mocks is a header (See Ana's comment [here](https://www.figma.com/file/oLzMM2zBFGmjIpo4X8vo5u?node-id=298:70262#168571185)). Given this along with Ben's confirmation of design's [source of truth](https://www.figma.com/file/xo9yAsH8PQUpi0eTJh9pmR/Product---Global?node-id=2527%3A3732), would you say it's safe to include this new weighting and make this element a header in my DO?

gnapse (commented):
When looking at [the linked Figma file](https://www.figma.com/file/xo9yAsH8PQUpi0eTJh9pmR/Product---Global?node-id=2527%3A3732), I think we should do something different from just adding a new weight value. See this below

image

It's not that we can select at will the bold level of a header. It's that different heading levels use different levels of boldness. So a "Header 3" is supposed to be semi-bold by virtue of its level alone.

Do you interpret the same thing @nats12? Perhaps @benbreckler can confirm?

If confirmed, we do not need to add this new weight value, but adjust the default styles so that each header level conforms to what's documented in that Figma design.

nats12:
> Do you interpret the same thing @nats12? 

I do @gnapse 👍🏼 it's also not just the weight but font-size as well for each header, but I will wait for Ben to confirm.

benbreckler:
@gnapse @nats12 Yes, the header styles differ in font weight, size but also line height and letter spacing. It's always recommended to check those properties for each style because they can vary from the other styles.

nats12:
@gnapse I'll be closing this PR and opening a new one where I'll align the headers in general with the file, not just the one I need.

PR #647, opened 2022-04-13
feat: <Button width="full" align="start|center|end">

gnapse:
## Short description

During my work in the new Task details view in Todoist, I've noticed the need for buttons to be optionally expanding their width instead of growing as little as needed to have their label fit. Also, when in this mode, we want to customize the label alignment as well.

image

I recommend looking at the diff with whitespace changes hidden.

## Test plan

Using storybook (`npm run storybook`), do the following:

- [x] The most important thing is to validate that buttons not using this new feature are 100% unaffected visually (I recommend using [the button playground story](http://localhost:6006/?path=/docs/design-system-button--playground)).
- [x] Play with the new full-width button in [the button full-width story](http://localhost:6006/?path=/docs/design-system-button--full-width). Play with different alignment values (left, center and right), and also change some other attributes (variant, size, long vs. short labels, etc.) See that it all looks good.

## PR Checklist

-   [ ] ~~Bumped version in `package.json` and `package-lock.json` (`npm --no-git-tag-version version <major|minor|patch>`) [ref](https://docs.npmjs.com/cli/v6/commands/npm-version)~~
-   [ ] ~~Updated all static build artifacts (`npm run build-all`)~~

## Versioning

New minor release.

### DIFF
--- a/src/new-components/base-button/base-button.module.css
+++ b/src/new-components/base-button/base-button.module.css
@@ -137,10 +137,6 @@
     --reactist-btn-font-size: var(--reactist-button-large-font-size);
 }
 
-.baseButton.width-full {
-    width: 100%;
-}
-
 /* setup the color variables for all the different states */
 /* variables are then set per variant and tone so that we do not repeat these definitions below */
 
--- a/src/new-components/base-button/base-button.module.css
+++ b/src/new-components/base-button/base-button.module.css
@@ -91,16 +91,16 @@
     text-align: center;
 }
 
-.align-left .label {
-    text-align: left;
+.align-start .label {
+    text-align: start;
 }
 
 .align-center .label {
     text-align: center;
 }
 
-.align-right .label {
-    text-align: right;
+.align-end .label {
+    text-align: end;
 }
 
 /* nice little animation when clicked to give some visual feedback */
--- a/src/new-components/base-button/base-button.module.css
+++ b/src/new-components/base-button/base-button.module.css
@@ -57,12 +57,6 @@
     --reactist-actionable-secondary-destructive-disabled-tint: #f1b7b2;
 }
 
-.label {
-    text-overflow: ellipsis;
-    white-space: nowrap;
-    overflow: hidden;
-}
-
 /* general styles for all buttons */
 .baseButton {
     max-width: 100%;
@@ -87,6 +81,26 @@
     transition-duration: 0.3s;
     transition-property: color, background-color;
     transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
+}
+
+.label {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    width: 100%;
+    text-align: center;
+}
+
+.align-left .label {
+    text-align: left;
+}
+
+.align-center .label {
+    text-align: center;
+}
+
+.align-right .label {
+    text-align: right;
 }
 
 /* nice little animation when clicked to give some visual feedback */
@@ -121,6 +135,10 @@
     --reactist-btn-height: var(--reactist-button-large-height);
     --reactist-btn-spacing: var(--reactist-button-large-spacing);
     --reactist-btn-font-size: var(--reactist-button-large-font-size);
+}
+
+.baseButton.width-full {
+    width: 100%;
 }
 
 /* setup the color variables for all the different states */
--- a/src/new-components/base-button/base-button.tsx
+++ b/src/new-components/base-button/base-button.tsx
@@ -54,7 +54,7 @@ type CommonProps = {
 }
 
 type AlignmentProps = {
-    width: 'full'
+    width: 'xsmall' | 'small' | 'medium' | 'large' | 'xlarge' | 'full'
     align?: 'left' | 'center' | 'right'
 }
 
@@ -119,6 +119,7 @@ export const BaseButton = polymorphicComponent<'div', BaseButtonProps>(function 
             ref={ref}
             aria-disabled={isDisabled}
             onClick={isDisabled ? preventDefault : onClick}
+            width={icon ? undefined : width}
             className={[
                 exceptionallySetClassName,
                 styles.baseButton,
@@ -126,7 +127,6 @@ export const BaseButton = polymorphicComponent<'div', BaseButtonProps>(function 
                 styles[`tone-${tone}`],
                 styles[`size-${size}`],
                 width !== 'auto' && icon == null && align != null ? styles[`align-${align}`] : null,
-                width !== 'auto' && icon == null ? styles[`width-${width}`] : null,
                 icon ? styles.iconButton : null,
                 disabled ? styles.disabled : null,
             ]}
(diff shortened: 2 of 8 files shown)


src/new-components/base-button/base-button.tsx
```diff
@@ -111,6 +117,8 @@ export const BaseButton = polymorphicComponent<'div', BaseButtonProps>(function
                 styles[`variant-${variant}`],
                 styles[`tone-${tone}`],
                 styles[`size-${size}`],
+                width !== 'auto' && icon == null && align != null ? styles[`align-${align}`] : null,
```
frankieyan, line 120: Since `align` is only meant to be used with the `width="full"` prop, it might be helpful to guard against them with something like this (haven't tried if it works though):

```TS
type AlignmentProps = {
	width: 'full'
	align?: 'left' | 'center' | 'right'
}

type AutoWidthProps = {
	width?: 'auto'
	align?: never
}

type LabelledButtonProps = { ... } & (AlignmentProps | AutoWidthProps)
```

src/new-components/button-link/button-link.stories.mdx
```diff
@@ -478,6 +478,92 @@ export function LoadingButtonLink(props) {
...
+            disabled: { control: false },
+            startIcon: { control: false },
+            endIcon: { control: false },
+            icon: { control: false },
+            tooltip: { control: false },
+            exceptionallySetClassName: { control: false },
+        }}
+        name="Full-width"
+    >
+        {FullWidthTemplate.bind({})}
```
frankieyan, line 563: I guess we can only pick between having controls vs having code snippets. Right now this shows up like this when you click on show code:

image

src/new-components/button-link/button-link.test.tsx
```diff
@@ -217,6 +217,27 @@ describe('ButtonLink', () => {
...
+            </ButtonLink>,
+        )
+        const buttonLink = screen.getByRole('link', { name: 'Click me' })
+        expect(buttonLink).toHaveClass('align-right')
+        expect(buttonLink).toHaveClass('width-full')
+    })
+
+    it('ignores align when width is not full', () => {
+        render(
+            <ButtonLink href="/" variant="primary" align="right">
```
frankieyan, line 233: Would be great if this can trigger `@ts-expect-error`

src/new-components/button/button.stories.mdx
```diff
@@ -467,9 +467,95 @@ export function LoadingButton(props) {
...
+        argTypes={{
+            label: {
+                control: { type: 'select' },
+                options: [
+                    'OK',
+                    'Submit',
+                    'Mark as done',
+                    'Yes, cancel my subscription',
+                    'Click me <em>now</em>',
+                    'If you click me now, youʼll take away the biggest part of me',
```
frankieyan, line 524: 😂

src/new-components/button-link/button-link.stories.mdx
```diff
@@ -478,6 +478,92 @@ export function LoadingButtonLink(props) {
...
+                options: ['small', 'normal', 'large'],
+                control: { type: 'inline-radio' },
+                defaultValue: 'normal',
+            },
+            disabled: { control: false },
+            startIcon: { control: false },
+            endIcon: { control: false },
+            icon: { control: false },
+            tooltip: { control: false },
+            exceptionallySetClassName: { control: false },
```
frankieyan, line 559: Can we add `{ control: false }` for `align` and `width` as well? Right now they render controls that don't do anything.

frankieyan (approved):
Looks good, just a minor comment to see if it's possible to guard the align and width props against each other

src/new-components/button-link/button-link.stories.mdx
gnapse: Yup, something we need to figure out with storybook. Perhaps it has bothered someone else already and there's a way around it. Not an issue introduced here, as all the other examples using templates are the same.

src/new-components/button-link/button-link.stories.mdx
gnapse: Good point.

src/new-components/base-button/base-button.tsx
gnapse: I thought about it, and as you know, we already do something along those lines to take care of the icon-only vs. the full-visual-text-label cases, as these have even more striking differences. I was hesitant, though, to introduce this level of complexity in the types. But maybe is something that the library can contain.

Another alternative that I thought was to not introduce the `width` prop at all. And make the button be full width whenever the `align` property is provided. But this on the other hand seemed a bit obscure.

Thoughts?

frankieyan: Yeah I don't think it should be omitted. It's not obvious `<Button align="center">` should be expanded just by looking at it 🤔 I think the explicitness of having `width="full"` in there would be more preferable.

gnapse: That's what I thought too.

I'll tinker the types a bit to see if I can achieve this without complicating things too much.

gnapse: Your suggestion seems to be working well. I'll validate a few things, but it looks like we'll make it.

src/new-components/button-link/button-link.test.tsx
gnapse: Done

src/new-components/button-link/button-link.stories.mdx
gnapse: Done

gnapse:
@frankieyan, in addition to adding your suggestion about the stricter types, I realized that the concept of `width` is already implemented in `Box`, and a button renders a `Box`, so I changed it so that we need no extra CSS for the width in `base-button.module.css`.

Moreover, I realized that I could expand this to not be only about `width="full"`, but any of the predefined box width values will also work. Hence, I added that change as well.

Since this is a non-trivial change, I'll wait for one last sanity check from you before I merge this. There's no hurry, and I'd rather be cautious with library changes like this.

frankieyan:
@gnapse my only comment is that I think the widths seem a little large for buttons (e.g. `small` is 400px). Even on the `<Box>` component, the `width` prop isn't getting too much use other than for its `full` value. That said, we could always introduce this now and come up with different values with the design team later, which shouldn't be too hard if we don't have a lot of use cases right now.

src/new-components/base-button/base-button.module.css
```diff
@@ -89,6 +83,26 @@
     transition-timing-function: cubic-bezier(0.4, 0, 0.2, 1);
 }
 
+.label {
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+    width: 100%;
```
pauloslund, line 90: @gnapse I believe this will introduce a breaking change to the Twist login page when Reactist is updated to this version:
| Before | After |
| --- | --- |
| image | image |

The only way around it seems to be adding custom CSS. Would it make sense for `width: 100%;` to only be applied when width is `full`?

gnapse: Indeed, if that fixes it, it makes sense that the label does not need to stretch in all situations. Does that fix it?

pauloslund: Yeah it seems to fix it. I'll create a PR for it!

PR #655, opened 2022-05-13
fix: Modal header's min-height is now applied when header has no button

gnapse:
## Short description

This PR makes sure that the modal header retains the same minimum height that it would have if it featured a close button on the right. For modals with a button visible, the header has the correct padding around the title because the button height ensures so.

I believe this may have been introduced in #633, hence my request for Pedro Alves to review (to make sure this fixes the issue without re-introducing the issue that #633 was attempting to fix).

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

CleanShot 2022-05-13 at 16 37 10@2x

</td>
<td>

CleanShot 2022-05-13 at 16 34 59@2x

</td>
</tr>
</table>

## PR Checklist

-   [x] Updated all static build artifacts (`npm run build-all`) _**(will do so after this is approved)**_

## Versioning

New patch release.


pedroalves0 (approved):
It would be nicer if the parent elements set the minimum height instead of having a dummy div. I tried to implement that quickly but hit some alignment issues. I understand that could require some layout changes, so I'm approving as is.

gnapse:
> It would be nicer if the parent elements set the minimum height instead of having a dummy div

I tried, but the parent element has a padding, hence I cannot set its height to 32px because that's not the final height that it ends up with. Its final height is 32px plus its top and bottom padding. I could compute that and use it, but then we'd need to update it if we ever change that padding. Hence, the most convenient approach is to make sure its content has 32px min-height, which will ensure the same overall height as when it has a close button.

PR #661, opened 2022-05-19
chore: Upgrade React@v17 TypeScript@v4.6 Storybook@v6.5

frankieyan:
## Short description

This bumps a few of our major dependencies up in order to [replace Reakit with Ariakit](https://github.com/Doist/reactist/pull/662).

- [x] React v17 
	- v16 is no longer a valid peer dependency for Ariakit
- [x] TypeScript
	- The reason I'm updating TypeScript is because some of [ariakit's](https://github.com/Doist/reactist/pull/662) generic and polymorphic types (ie. the `as` prop) can't be resolved correctly with v3.9, while they can with VSCode's built in version
	~~- Typechecking is running correctly after forcing tsdx is use the same version, but eslint and jest will have some noise around mismatched versions. As they still pass I'm not sure what the impact is yet~~
- [x] Storybook v6.5
  - There are a few logs in the console after the upgrade: 
	  - https://github.com/storybookjs/storybook/issues/18103
	  - [`argType.defaultValue` is deprecated](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-inferring-default-values-of-args)
- [x] eslint / @typescript-eslint / @doist/eslint-config
  ~~- The `lint` npm task is controlled by `tsdx` which uses eslint configs and plugins that aren't hoisted. When updating either eslint or typescript-eslint, it ends up running from the root level which causes the nested configs not to be resolved. There is also a whole can of peer dependencies that fails to be met due to old versions of `react-app` and `react-hooks`. If we upgrade `react-app`, then we introduce a lot of new violations that aren't a part of @doist/eslint-config. We're better off running eslint separately from tsdx so we can fully manage the config, but it won't be done as part of the upgrades here.~~
	- eslint is now running separate from tsdx so its related dependencies can be updated
- [x] Jest / ts-jest 
	~~- Tsdx fails with a promise rejected error before the test suite is run when these are updated~~
	- Like Eslint, Jest is now also run independently from tsdx to allow its dependencies to be updated
	- We're now using babel instead of ts-jest to transform TypeScript for tests, so that type errors don't block tests from running
	- The typeahead plugins when filtering for tests are pretty cool so they have been kept
- [x] tsdx v0.14

## PR Checklist

## Versioning

Major


package.json
```diff
@@ -22,7 +22,7 @@
     ],
     "engines": {
         "node": "^16.0.0",
-        "npm": "^7.0.0 || ^8.0.0"
+        "npm": "^8.3.0"
```
frankieyan, line 25: First version where `overrides` was introduced: https://github.com/npm/cli/releases/tag/v8.3.0

pauloslund (approved):
These changes all make sense, and I smoke tested storybook and everything seems to be working fine. I'm curious what your release plan for this is?

frankieyan:
> These changes all make sense, and I smoke tested storybook and everything seems to be working fine. I'm curious what your release plan for this is?

@pauloslund I plan to release this together with https://github.com/Doist/reactist/pull/662 with a major version bump, and pull it into both TD/TW, taking care of the breaking change with the `Tabs` component.

frankieyan:
Some size comparisons below. This should be attributed to setting the `targets` option for `@babel/preset-env` to `defaults, not IE 11`:

|-|Before|After|
|-|-|-|
|CommonJS|529KB|489KB|
|ESModules|504KB|458KB|
|Concatenated build (reactist.cjs.production.min.js)|74KB|71KB|

PR #663, opened 2022-05-20
refactor: Make label full width configurable

pauloslund:
## Short description

The main motivation for this PR was to remove the forced 100% width on button labels. However, in doing so I also noticed an opportunity to remove some custom CSS in favour of using props on existing components. The 100% width on button labels introduced a regression that requires custom styling as highlighted [here](https://github.com/Doist/reactist/pull/647/files#r875136060)

## Test Plan

- Run storybook, and compare https://doist.github.io/reactist/?path=/docs/design-system-button--full-width to http://localhost:6006/?path=/docs/design-system-button--full-width
- [x] The buttons look the same between local and production

## PR Checklist

## Versioning

Patch fix that is safe to go out with the next release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 # Next
 
 -   [Feat] Allow buttons to have fully rounded/circular corners.
+-   [Fix] Only apply full width labels when the button is full width
 
 # v12.1.1
 
--- a/src/new-components/base-button/base-button.module.css
+++ b/src/new-components/base-button/base-button.module.css
@@ -86,6 +86,7 @@
 .label {
     text-overflow: ellipsis;
     white-space: nowrap;
+    font-size: inherit;
 }
 
 /* nice little animation when clicked to give some visual feedback */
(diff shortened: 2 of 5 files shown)


pauloslund:
@gnapse Before I update the tests, what are your thoughts on this approach? Generally I thinking utilizing props where possible is always preferred, but I can see an argument for not doing so inside a design system component like this.

gnapse:
> @gnapse Before I update the tests, what are your thoughts on this approach? Generally I thinking utilizing props where possible is always preferred, but I can see an argument for not doing so inside a design system component like this.

I'm fine with this. I'd even say I prefer it.

gnapse:
@pauloslund one thing I wanted to ask you: under these changes, labels will keep stretching when we set the button to have full width, right? I'm thinking about use cases like this:

CleanShot 2022-06-01 at 09 36 05@2x

We still want in this case that the label stretches so that the `endIcon` shows on the right.

If so, then what case does this solved? I guess that those buttons in the login screens that motivated you to bring this up are not set to have full width, but they are stretched for some other reason, right?

pauloslund:
@gnapse Yup, full width buttons will behave how they are currently behaving. This only impacts buttons that stretch to the width of their container based on parent styling.

frankieyan:
I had to add overrides to a few places (buttons in the auth flows and sidebar prompts) in order to push out the latest version to Twist: https://github.com/Doist/twist-web/pull/4898

Once this is released they can be removed

pauloslund:
@gnapse would you be able to finish your review on this?

gnapse (approved):

PR #665, opened 2022-05-26
fix: Ensure tooltip doesn't override its anchor's ref

frankieyan:
## Short description

In #662 I made a mistake in the Tooltip component, which causes it to override the ref being passed to its anchor element. This breaks components that both require a tooltip, and another component that needs it as an anchor. For example, the scheduler in the new task detail's sidebar, which uses popper to display the scheduler menu. By losing the ref, the scheduler menu can't be positioned correctly near its trigger.

Unfortunately I couldn't test the build in-app as I'm running into issues with react-focus-lock once I pull in this branch through npm link (`ide run --link-reactist`), as Todoist is running a newer version and is able to dedupe the different copies when installing through npm normally. Even when updating the version here to match, I'm seeing weird behaviour once linked into the app. I'll have to test again once this is released.

### [Menu](http://localhost:6006/?path=/docs/components-menu--simple-menu-example)
- [ ] With the "With sub-menu" example, a tooltip appears when you focus on the menu trigger
	- [ ] Pressing enter to select a menu item should place the focus back on the trigger, but not show the tooltip

### [Tooltip](http://localhost:6006/?path=/docs/components-tooltip--tooltip-playground-story)

- [ ] Hovering or focusing on the box shows its tooltip

## PR Checklist

## Versioning

Patch


frankieyan:
@pauloslund I'm going to release this as a patch as it's blocking some of my work. Let me know if you have anything you'd like to see addressed when you get a chance 👍

pauloslund (commented):
👍

PR #659, opened 2022-05-18
refactor: Extract type alias

gnapse:
Closes #...

## Short description

This PR adds a semantic error message to most form field components.

- [Discussion in Twist](https://twist.com/a/1585/ch/81455/t/3484406/)

## PR Checklist

-   [x] Added tests for bugs / new features
-   [x] Updated docs (storybooks, readme)
-   [x] Executed `npm run validate` and made sure no errors / warnings were shown
-   [ ] Described changes in `CHANGELOG.md`
-   [ ] Bumped version in `package.json` and `package-lock.json` (`npm --no-git-tag-version version <major|minor|patch>`) [ref](https://docs.npmjs.com/cli/v6/commands/npm-version)
-   [ ] Updated all static build artifacts (`npm run build-all`)

## Versioning

New minor release.

PR #667, opened 2022-06-01
feat: Add a semantic message and tone to most field components

gnapse:
## Short description

- [Discussion in Twist](https://twist.com/a/1585/ch/81455/t/3484406/)
- Supersedes #659 (which only added error messages)

## Test plan

Review manually the live components in the various new stories titled "Message Tone Story" in storybook

- [x] Reviewed for `TextField`
- [x] Reviewed for `SelectField`
- [x] Reviewed for `PasswordField`
- [x] Reviewed for `TextArea`

<details>
<summary><strong>Demo</strong></summary>

<table><tr><td>

CleanShot 2022-06-01 at 16 37 07@2x

</td></tr></table>
</details>

## PR Checklist

## Versioning

New minor release

### DIFF
--- a/src/new-components/base-field/base-field.tsx
+++ b/src/new-components/base-field/base-field.tsx
@@ -6,6 +6,7 @@ import styles from './base-field.module.css'
 import { Stack } from '../stack'
 
 import type { WithEnhancedClassName } from '../common-types'
+import { Spinner } from '../spinner'
 
 type FieldHintProps = {
     id: string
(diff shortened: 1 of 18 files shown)


src/new-components/base-field/base-field.tsx
```diff
@@ -6,28 +6,107 @@ import styles from './base-field.module.css'
...
+}
+
 //
 // BaseField
 //
 
+type ChildrenRenderProps = {
+    id: string
+    'aria-describedby'?: string
+    'aria-errormessage'?: string
```
frankieyan, line 96: Maybe it's not an issue since the markup here is similar to mdn's example, but the error message itself isn't read when the field receives focus. It only gets read when you traverse to the message itself, but even then there's no indication that it's linked to the text field:

image
image

frankieyan (approved):

src/new-components/base-field/base-field.tsx
gnapse: Apparently, support for this is not great on VoiceOver (or in general) 😕

https://a11ysupport.io/tech/aria/aria-errormessage_attribute

cc @craigcarlyle what do you think we could do? Maybe we make it part of the description all the time? I tested this, and VoiceOver still announces that the field is marked as invalid. But it does not announce the error text. So having it in the description would have it read out loud.

craigcarlyle: @gnapse Since `aria-errormessage` came out in WAI-ARIA 1.1 in _2017_ and is _still_ not supported across assistive software I think we have no choice but to use `aria-describedby`. 

I think the suggestion [here](https://www.davidmacd.com/blog/test-aria-describedby-errormessage-aria-live.html) is the way to go:
Screen Shot 2022-06-06 at 3 15 04 PM

gnapse: Done in the last commit.

PR #670, opened 2022-06-30
feat: Allow buttons to have fully rounded/circular corners

gnapse:
## Short description

See https://twist.com/a/1585/ch/81455/t/3592631/

## PR Checklist

## Versioning

New feature. New minor version.

### DIFF
--- a/src/new-components/base-button/base-button.module.css
+++ b/src/new-components/base-button/base-button.module.css
@@ -69,7 +69,7 @@
     color: inherit;
     border: none;
     background-color: transparent;
-    border-radius: 5px;
+    border-radius: var(--reactist-border-radius-small);
     white-space: nowrap;
     font-family: var(--reactist-font-family);
     font-weight: var(--reactist-font-weight-medium);
@@ -117,6 +117,10 @@
     /* spinner */
     --reactist-spinner-tint: var(--reactist-btn-idle-tint);
     --reactist-spinner-fill: var(--reactist-btn-idle-fill);
+}
+
+.baseButton.shape-rounded {
+    border-radius: 1000px;
 }
 
 .baseButton.size-small {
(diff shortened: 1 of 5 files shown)


Bloomca:
@gnapse sorry I don't fully get it -- is it ready for review, or is it still a draft?

gnapse:
> @gnapse sorry I don't fully get it -- is it ready for review, or is it still a draft?

Sorry for the confusion. I left it as draft while I went ahead and created the Twist thread. Since I was asking there about how to design the API, I was not sure if it would change. But the essence remains, regardless of how the prop and its value are designed.

I'll remove the draft status now, since it appears that the preferred approach is the one I already used here (`shape="rounded"`).

src/new-components/base-button/base-button.module.css
```diff
@@ -119,6 +119,10 @@
     --reactist-spinner-fill: var(--reactist-btn-idle-fill);
 }
 
+.baseButton.shape-rounded {
+    border-radius: 1000px;
```
Bloomca, line 123: it doesn't change anything, just curious why not to use `border-radius: 100%`?

Bloomca (approved):
Code looks good.

I started the storybook and I didn't find any examples of how rounded buttons would look like (the property is documented though). Do you think it is worth to add a section with examples? I imagine a regular rounded button, and then an icon to show that it becomes round.

src/new-components/base-button/base-button.module.css
gnapse: I've never understood why, but that has an unintended effect:

CleanShot 2022-07-04 at 20 05 28@2x

I've always had to use a very large number when wanting a full round border radius.

frankieyan: We could reserve this one for `shape="fish"`?

Issue #676, opened 2022-08-09
Remove Enzyme


frankieyan:
Enzyme is dead at this point and a v18 adapter has no hope of being developed. See[ an article written by the author of v17's adapter here](https://dev.to/wojtekmaj/enzyme-is-dead-now-what-ekl).

There are 12 components with tests written in Enzyme - with some components being so old that we might even be able to remove if they're unused in Todoist and Twist. We should deprecate/remove components and rewrite the tests for the ones that remain.

frankieyan:
After https://github.com/Doist/reactist/pull/672 and https://github.com/Doist/reactist/pull/680, there will only be 9 test suites left with Enzyme.

PR #680, opened 2022-08-11
chore: Deprecate Dropdown, Input, Select; Remove Popover

frankieyan:
## Short description

As titled, this removes a couple components that have been deprecated for a while as we no longer use them in Todoist or Twist, and deprecates a few more components that will be removed in the future.

This allows us to reduce the scope of https://github.com/Doist/reactist/issues/676 so we have fewer legacy tests to maintain.

#### Deprecations
* `Dropdown`: still used in Twist and its export has been renamed to `DeprecatedDropdown`
* `Input`: still used in Twist and its export has been renamed to `DeprecatedInput`. `TextField` should be used instead
* `Select`: one instance in Twist and one in Todoist. `SelectField` should be used instead, renamed and will be removed in the next release

#### Removals
* `Popover`: this was undocumented and wasn't used in our codebases

---

image

Twist discussion around replacements for `Dropdown`: https://twist.com/a/1585/ch/190200/t/3685978/

## PR Checklist

## Versioning

Major


frankieyan:
On hold as this overlaps some of the work from https://github.com/Doist/reactist/pull/670

src/components/color-picker/color-picker.tsx
```diff
@@ -41,8 +41,8 @@ type Props = {
 
 function ColorPicker({ color = 0, small, onChange, colorList = COLORS }: Props) {
```
gnapse, line 42: I wonder if we should deprecate this one, given that:

- It is relying on a deprecated component (technically not a reason for deprecating it, as it should be re-implemented internally to use something else)
- More importantly: I think we should create a new one eventually that will probably have a different interface), and I'd also rename the new one to be a "select" and not a "picker".

Anyway, maybe not do anything about it now. I'm just dumping my thoughts.

frankieyan: Logged under https://github.com/Doist/reactist/issues/682 👍

PR #681, opened 2022-08-12
breaking: Support `themed` and `neutral` Tab variations

frankieyan:
## Short description

This changes the `Tab` component's API and drops a few variations that we no longer use. The `color` prop has been removed, and `variant` now accepts `themed` and `neutral` as valid options.

More details are in https://github.com/Doist/Issues/issues/6854 and https://twist.com/a/1585/inbox/t/3688056/.

image

## PR Checklist

## Versioning

Major

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 
 # Next
 
+-   [Breaking] The `color` prop has been removed from `Tabs`, and `variant` now only supports `themed` and `neutral` as options.
 -   [Fix] `Tooltip` will now correctly use an anchor component with an `as` prop
 
 # v13.0.0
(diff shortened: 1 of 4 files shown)


rfgamaral (approved):
Looks good to me, although I'm curious about the introduction of `Required`. Doesn't this introduce a breaking change? Do we want that?

frankieyan:
> Looks good to me, although I'm curious about the introduction of `Required`. Doesn't this introduce a breaking change? Do we want that?

image

@rfgamaral you mean this change right? That is for the context internal to the component, and not the component's props themselves which remain optional. Since we have default values for those props they should never end up being undefined in the context. The only thing it does is it saves us from doing a `contextValue.space ?? ''` when trying to use them as strings to get the css classnames.

rfgamaral:
Yes, that's what I was referring to. Thanks for the clarification.

frankieyan:
Hi @rfgamaral, sorry I've updated the component after a discussion with the design team [here](https://twist.com/a/1585/inbox/t/3688056/), and this is now a breaking change as I've removed older variants that we will no longer use. Can you please take another look :)

rfgamaral (approved):
Looks good to me 🚀

PR #683, opened 2022-08-15
feat: Field and textarea components improvements

gnapse:
## Short description

A handful of improvements and fixes related to field components. Most prominently:

1. Allow them to render without a label.
    - This was already technically possible, but it rendered a gap at the top of the component, the spacing between it and its labels. That gap is now gone if there are no labels to render.
    - The label prop is still not optional. Users must be explicit about not wanting a label by passing `label=""` or `label={null}`
2. Add tests for `TextArea` and a couple of fixes uncovered by the new tests.
    - See `CHANGELOG` for details

## PR Checklist

## Versioning

This will be merged and queued to an upcoming release.

The changes here are not breaking. This PR on its own should involve a minor release, as it only introduces fixes and a new feature (the removal of the gap, making the labelless field possibility official).

## Demo

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

CleanShot 2022-08-15 at 13 27 59@2x

</td>
<td>

CleanShot 2022-08-15 at 13 28 37@2x

</td>
</tr>
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@ Reactist follows [semantic versioning](https://semver.org/) and doesn't introduc
 # Next-Next
 
 -   [Feat] Fields without label show no spacing above the field
+-   [Fix] `TextArea` can now be hidden
+-   [Fix] `TextArea` now properly support receiving an explicit alternative aria-describedby attribute
 
 # Next
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# Next-Next
+
+-   [Feat] Fields without label show no spacing above the field
+
 # Next
 
 # v15.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,11 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
-# Next-Next
+# Next
 
 -   [Feat] Fields without label show no spacing above the field
 -   [Fix] `TextArea` can now be hidden
 -   [Fix] `TextArea` now properly support receiving an explicit alternative aria-describedby attribute
-
-# Next
 
 # v15.0.0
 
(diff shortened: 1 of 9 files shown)


gnapse:
@frankieyan I added you as an additional reviewer in case this can still make it to the upcoming release you'll be making of Reactist. But if this is not approved by the time you have to release, feel free to ignore this, as it is low priority for now.

src/new-components/text-area/text-area.stories.tsx
```diff
@@ -112,3 +114,28 @@ export function MessageToneStory() {
         </Stack>
     )
 }
+
+export function TextAreaWithoutLabelStory() {
```
frankieyan, line 118: Can we duplicate this story for other fields like SelectField and TextField?

src/new-components/base-field/base-field.tsx
```diff
@@ -94,8 +94,8 @@ type BaseFieldProps = WithEnhancedClassName &
          *
          * Avoid providing interactive elements in the label. Prefer `auxiliaryLabel` for that.
          *
-         * @see secondaryLabel
-         * @see auxiliaryLabel
+         * @see BaseFieldProps['secondaryLabel']
```
frankieyan, line 97: Maybe we can mention why we want this field to be required, but still allow `null` to be passed in in the description above.

frankieyan (commented):
This will make it easier to move away from the old Input component and support use cases like the link text box: 
image

Just a few minor comments - it might not make it to this upcoming release, but I have one immediately after for the new Tabs update so this can be a part of that release 👍

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# Next-Next
```
nats12, line 5: This may be documented somewhere and I've missed it, but why the need for a Next-Next? Why not add below?

nats12 (approved):
Looks good Ernesto, I don't have anything to add to Frankie's comments 👍🏼

src/new-components/base-field/base-field.tsx
gnapse: Well, technically, we do not do this for every prop in our components that accepts a `React.ReactNode`. And this type allows values such as `null` and `""`.

But yeah, I guess it wouldn't hurt to add a note in the jsdoc. I'll do it.

src/new-components/text-area/text-area.stories.tsx
gnapse: Ok

src/new-components/base-field/base-field.tsx
frankieyan: I'm assuming it is because we want to remind ourselves to always have a label associated with the field, but allow it to be omitted if we provide one externally 👀

PR #685, opened 2022-08-18
chore: Remove Enzyme

frankieyan:
Closes https://github.com/Doist/reactist/issues/676

## Short description

This removes Enzyme from our dependencies and tests that were using them are rewritten with React Testing Library. Note that not every test can be moved over 1:1 as some were digging into the components' internals. 

A few `data-testid` attributes were added, and we don't have a process to remove them from the production build, but the impact should be minor.

## PR Checklist

## Versioning

No release needed


pauloslund (approved):
> A few data-testid attributes were added, and we don't have a process to remove them from the production build, but the impact should be minor.

I think this is probably fine for this PR. It seems like the test ids could likely be replaced with better accessible roles, however I like that this PR doesn't change any functionality. Also, considering components like the dropdown are deprecated I think it's fine to not continue improving them.

Issue #679, opened 2022-08-09
Help update @reach/dialog to be compatible with React 18, or migrate to another compatible backing component


frankieyan:
We rely on Reach-UI's [Dialog](https://github.com/reach/reach-ui/tree/v0.17.0/packages/dialog) component to power our [Modal](https://github.com/Doist/reactist/tree/v12.1.1/src/new-components/modal). 

In Todoist, there are 11 modals built using the Reactist Modal, and 27 in Twist. Todoist also pulls in @reach/dialog directly and uses it to implement another 27 modals, bringing the total to 38.

There is an opportunity to contribute back to Reach-UI - its author is looking for help in making the component suite React 18 compatible ([ref](https://github.com/reach/reach-ui/issues/916#issuecomment-1205371907)). 

An alternative we can consider is to use [Ariakit's Dialog](https://ariakit.org/examples/dialog). It works with React 18 and we already have it as a dependency after migrating other Reakit-backed components over. Ideally, we pick one backing library to power Reactist as there are shared dependencies and lower-level components each of these components use.

Twist discussion: https://twist.com/a/1585/ch/190200/t/3686215/

PR #686, opened 2022-08-18
Breaking: Deprecate Reach-UI based modal and add Ariakit-based modal

frankieyan:
Closes https://github.com/Doist/reactist/issues/679

## Short description

This re-implements the modal with Ariakit. It tries to match the previous implementation as close as possible, both API and markup wise, so we don't leverage Ariakit's default modal backdrop.

## Test plan

Test that all of the previous modal's functionality have been retained in the storybook demos.

Old: http://localhost:6006/?path=/docs/components-modal
New: http://localhost:6006/?path=/story/design-system-modal

- [x] The modal can be closed by clicking on the backdrop or pressing Esc
- [x] Right clicking on the backdrop doesn't close the modal
- [x] The modal isn't rendered in the DOM when it's closed
- [x] Other elements that are siblings to the modal have `aria-hidden` set when the modal is open
- [x] Content outside the modal isn't scrollable when it's open
- [x] The modal's contents should be scrollable
- [x] The width/size props should be respected
- [x] Go to "Canvas" mode for the "Modal with standard actions footer" demo and open the modal. In the overflow menu, click on each option and check that they're logged to the "Actions" tab below
- [x] In the "Autofocus" demo, check that the "Address" text field is focused
- [x] In the "Stacking modals" demo, check that you can open a modal on top of another, and switch its props independently via the form within the modal 

## PR Checklist

## Versioning

Major


scottlovegrove (approved):
✅ All works as expected

frankieyan:
@scottlovegrove I added a couple more fixes after testing it in Todoist:

* Added a storybook demo to show rendering a modal on top of another
* Added a `Menu` component to the first demo to show menu items can be selected, and that they can be traversed with the keyboard

Could you please have a quick look at the last few commits and give them a spin in Storybook?

* [Add story for modal on top of modal](https://github.com/Doist/reactist/pull/686/commits/affe7e1394e295a5ded674b1384d49da5edcec21)
* [Prevent backdrop click event from propagating](https://github.com/Doist/reactist/pull/686/commits/3dc481838b8eb062e071796041b53b5ddcf78def)
* [Add menu to modal story](https://github.com/Doist/reactist/pull/686/commits/8602515fb3c708e8dbc5dcf0e0e26e58710a3b01)
* [Set default z-index for menu](https://github.com/Doist/reactist/pull/686/commits/94b37cddb26cd84dd69ee8c86b01918a445d9cf5)
* [Allow click events portaled from the modal to be handled](https://github.com/Doist/reactist/pull/686/commits/b4d4d5c464bec2e46ef40eda5b93f9c45aabcf94)
* [Add FocusLock to menu to allow it to trap focus on top of the modal](https://github.com/Doist/reactist/pull/686/commits/2521b52a4724c9cb79f8fc8be32006c31f8820bc)

src/new-components/modal/modal.stories.mdx
```diff
@@ -77,10 +82,30 @@ behaviour, see [ARIA: dialog role](https://developer.mozilla.org/en-US/docs/Web/
...
     >
         <ModalStoryStateProvider>
-            <Button variant="primary">Open modal</Button>
+            <Button variant="primary" action="open">
+                Open modal
+            </Button>
             <Modal aria-label="Modal with standard actions footer">
                 <ModalHeader>
-                    <Heading level="1">Modal with standard actions footer</Heading>
+                    <Columns>
```
gnapse, line 90: I think this fixes the misaligned modal header:

```suggestion
                    <Columns alignY="center">
```

Although, IMO, this extra button should be put in the header button slot, instead of putting it in the title slot. Something along these lines:

```jsx
<ModalHeader
  button={
    <Columns space="small">
      <Column><CustomButton /></Column>
      <Column><ModalCloseButton /></Column>
    </Columns>
  }
>
  <Heading level="1">Modal with standard actions footer</Heading>
</ModalHeader>
```

gnapse (approved):
Preliminary review, before I get into the new steps in the test plan:

1. **The header title is not aligned to the close button:**

    CleanShot 2022-08-29 at 11 06 43@2x

    This only happens in the first story, the one titled "Modal with standard actions footer". So it's most probably a problem with the story, and not with the modal component itself. It probably has to do with how you put this custom menu button in the header. I left a comment in code on what I think may fix this.

    <details>
    <summary>More info</summary>

    I realized while switching between two tabs: one with the new modal, and the other one with the old modal.

    CleanShot 2022-08-29 at 11 09 05

    </details>

2. In the "Stacking modal" story, I suggest changing the nested modal to have a different size than the parent modal initially, so that we can see how a modal is floating in top of the other one. That is, a different height and width setting.

Issue #678, opened 2022-08-09
Update storybook-addon-badges to React 18 compatible version


frankieyan:
We're on an older version ([0.0.4](https://github.com/geometricpanda/geometricpanda/releases/tag/0.0.4)) of storybook-addon-badges, but even the newest version [pulls in an older version of @storybook/theming at 6.2.9](https://github.com/geometricpanda/geometricpanda/blob/v0.2.1/package.json#L33), which lists `@emotion/core` (Emotion 10) as its dependency and isn't React 18 compatible. 

Newer versions of @storybook/theming have already upgraded to Emotion 11, and it should be kept on the same version as the rest of our storybook dependencies. We may be able to force npm to reconcile to the newer version of @storybook/theming using package.json's overrides field, and in the meantime submit a PR to addon-badges to update storybook's version upstream.

Issue #677, opened 2022-08-09
Replace @storybook/addon-knobs with addon-controls in Reactist


frankieyan:

HenryKenya:
Will this involve re-writing the stories using args? Sounds interesting and I'd like to give it a shot @frankieyan

frankieyan:
Hi @HenryKenya - yes `args` and `argTypes` can both be used. The latter is to give more control into what kind of UI we want to show in order to change the props: https://storybook.js.org/docs/react/essentials/controls#choosing-the-control-type

The two stories are:
https://github.com/Doist/reactist/blob/v14.0.0/stories/components/Avatar.stories.tsx
https://github.com/Doist/reactist/blob/v14.0.0/stories/components/KeyboardShortcut.stories.tsx

I believe they already have `argTypes` set up, but `Avatar` still declares knobs as a decorator even though it is not needed, and `KeyboardShortcut` seems to have both declared.

Once both of these have knobs removed, we can then remove the `@storybook/addon-knobs` dependency.

Thank you for offering to do this :) Let me know how it goes 🙌

PR #687, opened 2022-08-23
deps: Update @geometricpanda/storybook-addon-badges to v0.2.2

frankieyan:
Closes https://github.com/Doist/reactist/issues/678

## Short description

This updates @geometricpanda/storybook-addon-badges to the latest version, because the previous version we were using has an older version of @storybook/theming as a dependency, which includes the non-React 18 compatible Emotion 10.

~~The interesting thing is that the newest version of addon-badges lists @storybook/theming@6.2.9 as a prod dependency without using the `^` semver range, but once installed in Reactist, the semver symbol is included, allowing it to use v6.5.3 and deduped between other dependencies that use the theming package. I have no idea why this happened though~~ I've been looking at the wrong package.json the whole time in the repo. This is the correct one: https://github.com/geometricpanda/geometricpanda/blob/72ac16e6eb121480efccc426a4aac275d3b09b4d/libs/storybook-addon-badges/package.json#L40

image
image

The only remaining package that pulls in Emotion 10 is addon-knobs, which will be removed with https://github.com/Doist/reactist/issues/677

image

How the badges are styled were deprecated as well, and has been updated to the new API ([ref](https://github.com/geometricpanda/geometricpanda/tree/v0.2.1/libs/storybook-addon-badges#deprecation-notice)).

## Test plan

- [ ] Test that running npm install on this branch results in the package tree above (same versions of @storybook/theming throughout, only one instance of @emotion/core left)
- [ ] Run `npm run storybook` and check that the badges resembles the deployed docs at https://doist.github.io/reactist (it's by default positioned a little further away from the tabs which I thought looked better)

|Before|After|
|-|-|
|image|image|

## PR Checklist

## Versioning

No release needed


frankieyan:
> 💭 (nit): The badge has more padding on the left than on the right: image

@pawelgrimm the padding/margin around the badges are set here: https://github.com/geometricpanda/geometricpanda/blob/72ac16e6eb121480efccc426a4aac275d3b09b4d/libs/storybook-addon-badges/src/lib/blocks/badges.tsx#L57

I've tried setting `layoutMargin` on the theme (in `.storybook/theme.js`), but it doesn't seem to allow being overridden. Its [type](https://github.com/storybookjs/storybook/blob/f2324595c8953682dec3be6a3201aca577ad9467/lib/theming/src/types.ts#L5) and [docs](https://storybook.js.org/docs/react/configure/theming) also exclude it from `create`'s config options. I tried to run the addon's storybook, but I ran into some errors on that end too. We can log an issue later so they can be aware of the issue though.

PR #689, opened 2022-08-23
fix: Set max-height on reactist_menulist

jvalente:
Closes https://github.com/Doist/Issues/issues/7375

## Short description

Applies `min-height: 44px` and `max-height: var(--popover-available-height)` to `.reactist_menulist` in order to allow the menu list popover to scroll when it's height is greater than the viewport.

The min-height is set according to the [minimum target size for pointer inputs](https://www.w3.org/TR/WCAG21/#target-size).

This fixed is based on the ariakit [menu example implementation](https://github.com/ariakit/ariakit/blob/9b2af70dc86fa767ffb8fb268ade591bc988cce7/packages/ariakit/src/menu/__examples__/menu/style.css#L4).

## Test plan

-   Run the Reactist storybook
- Navigate into the `Components > Menu > Simple Menu Example` canvas tab.
- Expand the Simple Menu
- [x] After reducing the canvas height the simple menu popover contents become scrollable.

## PR Checklist

### DIFF
--- a/src/components/menu/menu.less
+++ b/src/components/menu/menu.less
@@ -4,6 +4,9 @@
 // Ref: https://github.com/ariakit/ariakit/blob/454aef1237ada31e8292ee4966adca3ed0f6e299/packages/ariakit/src/menu/menu-list.ts#L150
 .reactist_menulist[role='menu'],
 .reactist_menulist[role='menubar'] {
+    min-height: 44px; // minimum accessible target area
+    max-height: var(--popover-available-height); // defined by ariakit
+    overflow: auto;
     display: block;
     white-space: nowrap;
     background: hsla(0, 100%, 100%, 0.99);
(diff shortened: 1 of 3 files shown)


src/components/menu/menu.less
```diff
@@ -4,6 +4,9 @@
 // Ref: https://github.com/ariakit/ariakit/blob/454aef1237ada31e8292ee4966adca3ed0f6e299/packages/ariakit/src/menu/menu-list.ts#L150
 .reactist_menulist[role='menu'],
 .reactist_menulist[role='menubar'] {
+    min-height: 44px; // minimum accessible target area
+    max-height: var(--popover-available-height);
```
pedroalves0, line 8: Perhaps it makes sense to add a comment here indicating that `--popover-available-height` is defined by Ariakit?

pedroalves0 (approved):

src/components/menu/menu.less
jvalente: Yep! Makes complete sense.

PR #697, opened 2022-09-01
feat: Enhance the `label` property in `CheckboxField` to be of the type `React.ReactNode`

rfgamaral:
## Short description

This change enhances the `label` property in the `CheckboxField` component to be of the type `React.ReactNode`, allowing one to use a React component as label, instead of only a `string`.

This will permit that the checkbox, icon, and text label, in the items such as the ones below, are all clickable (thus toggling the check state on and off):

image

## PR Checklist

## Versioning

Minor: `v15.1.0`

### DIFF
--- a/src/new-components/checkbox-field/checkbox-field.test.tsx
+++ b/src/new-components/checkbox-field/checkbox-field.test.tsx
@@ -54,12 +54,12 @@ describe('CheckboxField', () => {
         expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show unread badge')
     })
 
-    it('is labelled by its `String` label', () => {
+    it('is labelled by its label (string)', () => {
         render(<CheckboxField data-testid="checkbox-field" label="Show completed tasks" />)
         expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show completed tasks')
     })
 
-    it('is labelled by its `ReactNode` label', () => {
+    it('is labelled by its label (ReactNode)', () => {
         render(
             <CheckboxField
                 data-testid="checkbox-field"
@@ -67,7 +67,7 @@ describe('CheckboxField', () => {
             />,
         )
         expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show completed tasks')
-        expect(screen.getByTestId('react-node-label')).toHaveTextContent('Show completed tasks')
+        expect(screen.getByTestId('react-node-label')).toBeInTheDocument()
     })
 
     it('issues a warning if no label is given', () => {
--- a/src/new-components/checkbox-field/checkbox-field.test.tsx
+++ b/src/new-components/checkbox-field/checkbox-field.test.tsx
@@ -54,9 +54,20 @@ describe('CheckboxField', () => {
         expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show unread badge')
     })
 
-    it('is labelled by its label', () => {
+    it('is labelled by its `String` label', () => {
         render(<CheckboxField data-testid="checkbox-field" label="Show completed tasks" />)
         expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show completed tasks')
+    })
+
+    it('is labelled by its `ReactNode` label', () => {
+        render(
+            <CheckboxField
+                data-testid="checkbox-field"
+                label={<span data-testid="react-node-label">Show completed tasks</span>}
+            />,
+        )
+        expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show completed tasks')
+        expect(screen.getByTestId('react-node-label')).toHaveTextContent('Show completed tasks')
     })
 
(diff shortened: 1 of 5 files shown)


rfgamaral:
cc @gnapse, this follows-up our conversation over DM on Twist

src/new-components/checkbox-field/checkbox-field.test.tsx
```diff
@@ -54,11 +54,22 @@ describe('CheckboxField', () => {
...
 
+    it('is labelled by its `ReactNode` label', () => {
+        render(
+            <CheckboxField
+                data-testid="checkbox-field"
+                label={<span data-testid="react-node-label">Show completed tasks</span>}
+            />,
+        )
+        expect(screen.getByTestId('checkbox-field')).toHaveAccessibleName('Show completed tasks')
+        expect(screen.getByTestId('react-node-label')).toHaveTextContent('Show completed tasks')
```
gnapse, line 70: This line is not testing anything about the library. It is testing something you rendered in the test.

That is, there's nothing you can change in the component code that would make this assertion fail.

I suggest we remove it:

```suggestion
```

src/new-components/checkbox-field/checkbox-field.tsx
```diff
@@ -104,7 +104,7 @@ const CheckboxField = React.forwardRef<HTMLInputElement, CheckboxFieldProps>(fun
                 indeterminate={indeterminate}
                 aria-hidden
             />
-            {label ? <Text>{label}</Text> : null}
+            {label ? typeof label === 'string' ? <Text>{label}</Text> : label : null}
```
gnapse, line 107: I think we need to put it inside `<Text />` regardless. Otherwise, it will not pick up the correct text color, for instance, when in dark mode.

I'm approving the PR, but with the expectation that this needs to be fixed before merging.

gnapse (approved):
I'm approving, but one of my comments needs to be addressed before this is merged and released.

src/new-components/checkbox-field/checkbox-field.tsx
rfgamaral: But since this now supports _any_ `ReactNode`, we can pass along anything, won't the `<Text>` conflict with more complex labels? Although I don't see why one would do that, but capability is there.

I'm wondering if we should add a `startIcon` instead, like the `<Button>`. What do you think? We'll need a way to control the margin between the checkbox / icon / label, but I guess we can do that from the consumer side with proper selectors.

src/new-components/checkbox-field/checkbox-field.test.tsx
rfgamaral: I understand what you mean, but I added this as a way to test that you can pass a `ReactNode` to the component, and that it's rendered properly. This test correctly asserts for that. I feel that removing this assertion, renders the new test I added irrelevant, don't you think?

src/new-components/checkbox-field/checkbox-field.tsx
gnapse: I'm not a fan of a start icon. Where does it end? Will we need an end icon too? And if we want the icon, or any other markup in the middle? For instance, what if we want this:

```jsx
<CheckboxField
  label={<>This is <strong>important</strong> so pay attention}
/>
```

The labels in all the other fields already support `React.ReactNode`, this was an oversight with the checkbox label.

> But since this now supports _any_ `ReactNode`, we can pass along anything, won't the `<Text>` conflict with more complex labels? Although I don't see why one would do that, but capability is there.

The capability is there, but the primary goal of the label is to display text, and `<Text />` is what ensures that this text plays nicely with the design system theming and styles. You can see in `BaseField` (the component that serves as the basis for all fields except the checkbox) that we already wrap all the label are inside `<Text />`.

src/new-components/checkbox-field/checkbox-field.test.tsx
gnapse: If you can make a single change in the component code that breaks this assertion, and this assertion only, I'll agree with you.

gnapse: If what you want to test is about making sure that the react node content is rendered verbatim, I'd say there are clearer ways to do that. For instance, use a `label={<>Hello <strong>world</strong></>}` and then assert:

```js
expect(screen.getByText('world').tagName).toBe('strong')
```

Or maybe you can do:

```suggestion
        expect(screen.getByTestId('react-node-label')).toBeInTheDocument()
```

No need to test the text content twice. I think that's what tipped me off a bit.

rfgamaral: > No need to test the text content twice. I think that's what tipped me off a bit.

In the end, all 3 alternatives will assert the same thing, I'm okay with switching to one of those simpler options 👍

src/new-components/checkbox-field/checkbox-field.tsx
rfgamaral: That makes sense, thanks for clarifying.

PR #699, opened 2022-09-06
refactor: Change deprecated way to import React

gnapse:
## Reference

- https://twist.com/a/1585/ch/190200/t/3739793/
- https://twitter.com/dan_abramov/status/1308739731551858689

## Test plan

Not really anything to test, but I still want explicit approval.

### DIFF
--- a/.storybook/BaseDecorator.js
+++ b/.storybook/BaseDecorator.js
@@ -1,6 +1,6 @@
 /**
  * This decorator acts as a base decorator for all stories.
  */
-import React from 'react'
+import * as React from 'react'
 
 export default (Story) => Story()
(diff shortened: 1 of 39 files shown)


nats12 (approved):
Looks good 👍🏼  from the thread's poll we may need to change this again in the near future though to named imports.

gnapse:
> Looks good 👍🏼 from the thread's poll we may need to change this again in the near future though to named imports.

The poll seems to have the wildcard import winning. But even if it did not, making the move to named imports is not easily achievable via search and replace. It may need a codemod script or something like that. In the meantime, we move away from the default React import that's deprecated.

PR #714, opened 2022-11-16
feat: Add support for Menu as context menu

frankieyan:
Closes https://github.com/Doist/Issues/issues/7077

## Short description

This adds a `<MenuContextMenuTrigger>` component to allow an element to open a menu as a context menu.

image

Reference: https://ariakit.org/examples/menu-context-menu

## PR Checklist

## Test plan

- [x] Start storybook and go to the Menu story. In the context menu example, you should be able to trigger the menu using the right mouse button, or ctrl + click on a Mac, using the primary button. You can also trigger the same menu by clicking on the secondary button.

## Versioning

Minor

### DIFF
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -123,9 +123,9 @@ const MenuButton = polymorphicComponent<'button', MenuButtonProps>(function Menu
 })
 
 //
-// MenuContextMenuTrigger
+// ContextMenuTrigger
 //
-const MenuContextMenuTrigger = polymorphicComponent<'div', unknown>(function MenuContextMenuTrigger(
+const ContextMenuTrigger = polymorphicComponent<'div', unknown>(function ContextMenuTrigger(
     { as: component = 'div', ...props },
     ref,
 ) {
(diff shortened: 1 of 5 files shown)


scottlovegrove (approved):

src/components/menu/menu.tsx
```diff
@@ -107,6 +122,26 @@ const MenuButton = polymorphicComponent<'button', MenuButtonProps>(function Menu
     )
 })
 
+//
+// MenuContextMenuTrigger
+//
+const MenuContextMenuTrigger = polymorphicComponent<'div', unknown>(function MenuContextMenuTrigger(
```
gnapse, line 128: @frankieyan 

This is a bit of a nit-pick, but I'm having trouble parsing this name in my mind.

What about `ContextMenuTrigger`?

gnapse (approved):
One thing I want to know about this solution is: how does it look like when we want to have the same menu be available as a context menu, and as a menu triggered by a button?

Do we have to repeat the menu items list in both places? No big deal, just checking. I guess there's not much of a way around it. And we should always have the menu list as a reusable component to DRY things up.

src/components/menu/menu.tsx
frankieyan: I was torn on this one, as the other components here have `Menu` as a prefix, but let's go with your suggestion 👍

PR #717, opened 2022-11-25
feat: Add `startIcon` support to `TextField`

frankieyan:
## Short description

image

This adds the `startIcon` prop to the `TextField` component, as we occasionally need to show an icon to convey the functionality of a text field, e.g. for a search field.

~~`useCombinedRefs` has also been extracted from the TD codebase and made available in Reactist.~~

## Test plan

- [x] An icon appears in the `TextField`'s Icon story
- [x] The icon doesn't interfere with the text field's functionality. You can focus and enter your input as normal
- [x] Clicking on the icon focuses the text field

## PR Checklist

## Versioning

Minor


src/new-components/text-field/text-field.module.css
```diff
@@ -53,3 +53,9 @@
 .inputWrapper:not(.bordered) input {
     padding: var(--tmp-vertical-padding) 10px;
 }
+
+.startIcon {
+    margin-right: calc(var(--reactist-spacing-xsmall) * -1);
+    padding-left: 10px;
+    cursor: text;
+}
```
pauloslund, line 61: 💭 Do you think `.startIcon` should have a max height? On buttons no matter how big the icon is it won't cause the button to be larger than normal, however that is not currently the case with text fields.

src/new-components/text-field/text-field.module.css
```diff
@@ -53,3 +53,9 @@
 .inputWrapper:not(.bordered) input {
     padding: var(--tmp-vertical-padding) 10px;
 }
+
+.startIcon {
+    margin-right: calc(var(--reactist-spacing-xsmall) * -1);
```
pauloslund, line 58: 💭 You could add `marginRight="-xsmall"` to the `Box` component.

src/new-components/text-field/text-field.tsx
```diff
@@ -25,10 +27,18 @@ const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(function Te
...
+        startIcon,
         ...props
     },
     ref,
 ) {
+    const internalRef = React.useRef<HTMLInputElement>()
+    const combinedRef = useCombinedRefs(ref, internalRef)
+
+    function focusOnIconClick() {
+        internalRef.current?.focus()
```
pauloslund, line 39: Thoughts on adding a test for the icon that ensures this focus works?
```jsx
function SearchIcon() {
    return (
        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" data-testid="SearchIcon">
            <path
                d="M10.5 3a7.5 7.5 0 015.645 12.438l4.709 4.708a.502.502 0 01-.708.708l-4.708-4.709A7.5 7.5 0 1110.5 3zm0 1a6.5 6.5 0 100 13 6.5 6.5 0 000-13z"
                fill="currentColor"
            />
        </svg>
    )
}

it('supports rendering an icon', () => {
    render(<TextField label="Whatʼs your job title?" startIcon={<SearchIcon />} />)

    userEvent.click(screen.getByTestId('icon'))

    const inputElement = screen.getByRole('textbox', { name: 'Whatʼs your job title?' })
    expect(inputElement).toHaveFocus()
})
```

pauloslund (approved):
Looks good! I left a couple small comments/suggestions on it.

gnapse:
Why not an `endIcon` too?

Why not both

src/hooks/use-combined-refs/use-combined-refs.stories.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+function MyComponent(props, ref) {
+    const myRef = useRef()
+    const combinedRef = useCombinedRefs(ref, myRef)
+
+    return (
+        <Box ref={combinedRef} />
+    )
+}
+```
```
gnapse, line 21: I suspect that the functionality provided by this hook must be already offered by the generic hooks library we use. So my point is: do we really want to export this from Reactist? It forces us to either maintain it in the future, or go through a cumbersome deprecation cycle if we decide to remove it. Especially now that we no longer offer any hooks from within Reactist.

frankieyan: Good call, [useMergeRefs](https://github.com/theKashey/use-callback-ref#usemergerefs) works identically here 👍

PR #718, opened 2022-11-29
feat: Allow defining a custom class on the Modal's overlay

pauloslund:
## Short description

We need a way for consumers to define custom classes on the new Modal's overlay. This is needed to support custom styling for the overlay (eg. both the command menu, and quick add modal in Todoist), and for adding custom animations.

## PR Checklist

## Versioning

Patch


pauloslund:
@gnapse Wanted to get your thoughts on this addition. I know generally we don't want to introduce exceptions to the design system, but I haven't been able to come up with an alternative approach that I liked. To clarify, _some_ modals need to be able to restyle the included overlay. Generally this is by removing the background colour entirely, but they have also been adding classes to manage custom animations.

gnapse (approved):
This looks good as a short-term solution. I still think that long term we need to look at it differently, but it is precisely these learnings and needs that will drive future changes that can make this less of an escape hatch. For now, this is good IMO.

PR #722, opened 2022-12-06
feat: Add support for preventing Modal from hiding when clicking overlay

pedroalves0:
## Short description

Certain Todoist modals are not dismissable if their overlay is clicked. We're adding here a new prop, `hideOnOverlayClick`, to control that behavior.

## PR Checklist

## Versioning

Minor

### DIFF
--- a/src/new-components/modal/modal-stories-components.tsx
+++ b/src/new-components/modal/modal-stories-components.tsx
@@ -17,6 +17,7 @@ function Link({ children, ...props }: JSX.IntrinsicElements['a']) {
 
 type ModalStoryState = Pick<ModalProps, 'width' | 'height'> & {
     button: 'true' | 'false' | 'custom'
+    hideOnOverlayClick: boolean
     withScrollableContent: boolean
 }
 
@@ -24,6 +25,7 @@ const defaultInitialState: ModalStoryState = {
     width: 'medium',
     height: 'expand',
     button: 'true',
+    hideOnOverlayClick: true,
     withScrollableContent: false,
 }
 
@@ -99,9 +101,14 @@ function ScrollableContent({ label = 'Item', count = 20 }: { label?: string; cou
 }
 
 function ModalOptionsForm({ title }: { title?: React.ReactNode }) {
-    const { button, width, height, withScrollableContent, onChange } = React.useContext(
-        ModalStoryContext,
-    )
+    const {
+        button,
+        width,
+        height,
+        hideOnOverlayClick,
+        withScrollableContent,
+        onChange,
+    } = React.useContext(ModalStoryContext)
     return (
         <Stack space="large">
             {title}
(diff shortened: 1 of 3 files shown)


pedroalves0:
I'd like to mention that I did check what other libraries support, for instance ariakit's Dialog has [hideOnInteractOutside](https://github.com/ariakit/ariakit/blob/main/packages/ariakit/src/dialog/dialog.tsx#L113). This said, a modal might have different requirements than a more general dialog, so let's discuss your questions.

> Last but not least a question: what's the use case in Todoist for this to be needed?

Better to answer this one first, it's Quick Add.

>  If what we really want is to only prevent the dialog closing when the overlay is clicked, I think that we do need the new prop anyway. Hence my question below to see if we really need this feature.

Quick Add is special in the sense that sometimes it accepts being dismissed when the overlay is clicked and sometimes it doesn't. For instance, it's mounted as a modal in the `QuickAddPage` component even though it's the only component. Clicking on the surrounding blank space there shouldn't hide the modal.

> I mean, ultimately, the consumer of the dialog controls the ability to dismiss it (via the onDismiss callback). Couldn't this be the means to decide if the modal should be dismissed or not?

I had considered this before, but `onDismiss` doesn't pass any information of what triggered the close action. Therefore there isn't a way to cancel `onDismiss` just when clicking the overlay in the `QuickAddPage` component.

---

Now, I'm in the process of migrating the remaining Todoist reach-ui modals to Reactist Modal. I'm sure that we could refactor `QuickAddPage` to not mount Quick Add as a Modal, and thus void this use case, but I'm trying to refactor as little as possible to minimize the scope of the changes footprint. Secondly, given that other libraries support similar props,  I thought this would be a well-received addition for future use cases.

As you're strongly opinionated on this area your guidance is appreciated, so please let me know if these points weren't enough to convince you.

pedroalves0:
As I wrote my answer above I just realized that `Modal` uses Ariakit's `Dialog` underneath. Instead of adding this new prop, how does lifting the `hideOnEscape` and `hideOnInteractOutside` sound to you?

frankieyan:
@pedroalves0 please tag me when you've made the change to lift ariakit's props 👍

pedroalves0:
> @pedroalves0 please tag me when you've made the change to lift ariakit's props 👍

Done @frankieyan .

src/new-components/modal/modal-stories-components.tsx
```diff
@@ -17,13 +17,15 @@ function Link({ children, ...props }: JSX.IntrinsicElements['a']) {
 
 type ModalStoryState = Pick<ModalProps, 'width' | 'height'> & {
     button: 'true' | 'false' | 'custom'
+    hideOn: 'escapeAndOverlay' | 'escape' | 'overlay'
```
frankieyan, line 20: 💭 Not super important but I would prefer to see this follow the same naming as the props, ie. `escapeAndInteractOutside` and `interactOutside`

Issue #688, opened 2022-08-23
Ariakit, used as a prod dependency, should have its version pinned instead of allowing a range


frankieyan:
### Current behavior

Discovered in @scottlovegrove's personal project, when Reactist is installed as a dependency, newer versions of Ariakit with breaking changes can end up being installed. This is because Ariakit uses prerelease tags, and the caret semver range does not prevent upgrades between them: [ref](https://docs.npmjs.com/cli/v6/using-npm/semver#caret-ranges-123-025-004)

Breaking change introduced in:
https://github.com/ariakit/ariakit/pull/1521
https://github.com/ariakit/ariakit/releases/tag/ariakit%402.0.0-next.32

### Steps to reproduce the bug

* Install Reactist in another project. You should see a version of Ariakit newer than `2.0.0-next.27` to be installed.

### Expected behavior

We should explicitly declare Ariakit versions we are using

PR #728, opened 2022-12-16
chore: Update ariakit dependency to 2.0.0-next.30

pedroalves0:
## Short description

Reactist is currently a blocker to the undergoing efforts to enable React v18 in Todoist (https://github.com/Doist/todoist-web/issues/4910). The reason is that one of its dependencies, Ariakit, throws at times a `Cannot call an event handler while rendering.` error from its own internal `useEvent` function, and it shouldn't. While visually it's not a breaking issue, several unit tests fail.

This problem has been fixed in Ariakit a while ago (https://github.com/ariakit/ariakit/pull/1481), but we froze Ariakit's version because newer versions brought breaking changes (https://github.com/Doist/reactist/issues/688). The good news is that the fix is present in a version (`2.0.0-next.30`) that was deployed just before the one that's problematic(`2.0.0-next.32`).

## Test plan

For peace of mind, you can run storybook and verify that the components run without errors. Special focus to the components that use Ariakit's `Dialog` underneath (`2.0.0-next.32` breaking changes modified some of its props).

## PR Checklist

## Versioning

Minor

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,8 +12,8 @@
             "dependencies": {
                 "@reach/dialog": "^0.16.0",
                 "aria-hidden": "^1.2.1",
-                "ariakit": "2.0.0-next.27",
-                "ariakit-utils": "0.17.0-next.18",
+                "ariakit": "2.0.0-next.30",
+                "ariakit-utils": "0.17.0-next.21",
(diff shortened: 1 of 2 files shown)


craigcarlyle (approved):
🚀

PR #730, opened 2022-12-27
test: Introduce Chromatic for visual regression tests

gnapse:
## Short description

This is work in progress, not even sure it's going to make it to `main` yet.

## PR Checklist


frankieyan:
@gnapse I'm going to re-roll as I'll be off the first half of next week. Super interesting though, I'll definitely check it out when I'm back 👍

scottlovegrove (approved):
Was able to view the storybook without the chromatic account. 
Was able to create a chromatic account and once in, it showed me the Reactist project. There were two there, one is `Reactist` the other is `reactist`. I think the former is the one we are really expecting, correct?

gnapse:
> There were two there, one is `Reactist` the other is `reactist`. I think the former is the one we are really expecting, correct?

Yup. I had not notice that, and do not know how that other one came into being, but I now removed it.

PR #731, opened 2022-12-29
refactor: Refactor modal stories, new modal prop portalElement

gnapse:
## Short description

This is work extracted off of #730, that I deemed useful regardless of whether we adopt Chromatic or not. It is mostly about improvements I had to make to the modal stories to take advantage of Chromatic visual regression tests on modals, but these improvements add value on their own without adopting Chromatic.

Here's a list of the relevant changes:

1. Rewrote modal stories into two modules:
    - MDX for text docs (mostly documenting component APIs and props)
    - `.tsx` for modal example stories that actually render the modal component
2. Introduce a new `portalElement` prop on the modal, allowing consumers to customize where in the DOM they want their portal element to be mounted.
3. Added [storybooks interaction testing](https://storybook.js.org/docs/react/writing-tests/interaction-testing) to modal stories.
    - This now makes it so that modals automatically open in modal stories.
    - This opens the door to use storybooks as a place where we run in-browser component tests.
4. Various other small improvements to stories in other components.

## PR Checklist

## Versioning

A new minor version change, due to the new feature introduced to the modal component.


package.json
```diff
@@ -1,11 +1,18 @@
 {
     "name": "@doist/reactist",
     "description": "Open source React components by Doist",
-    "author": "Henning Muszynski <henning@doist.com> (http://doist.com)",
+    "author": {
```
rfgamaral, line 4: 💭 Nothing against Henning 😅, but doesn't it make more sense to remove this?

PR #732, opened 2022-12-29
test: Introduce visual regression tests using Playwright

gnapse:
## Short description

Work in progress.

See https://twist.com/a/1585/ch/607587/t/3963022/

## PR Checklist


gnapse:
Something introduced in this PR has broken the tsdx build command. I noticed first on CI, but when I run it locally I can reproduce. To do so, run this command locally:

```sh
tsdx build --tsconfig tsconfig.dist.json
```

The closest I've found online related to the error message is https://github.com/rollup/rollup/issues/3591. Maybe tsdx is using a rollup version older than the one in which a fix for that issue was released. Meaning we really need to get rid of tsdx soon.

henningmu:
> Maybe tsdx is using a rollup version older than the one in which a fix for that issue was released

According to package-lock, tsdx is using 2.79.1. The fix went live with 2.11.0. Must be something else 😞

> Meaning we really need to get rid of tsdx soon

This is true regardless, and might be the easier solution than figuring out the root cause here 😅

PR #737, opened 2023-01-05
feat: Badge component improvements

gnapse:
## Short description

While preparing for [upcoming work](https://twist.com/a/1585/ch/628546/t/4031093/), I noticed that we're still using in Todoist an in-app badge component in some cases, alongside cases that are using the newer shared `Badge` component from Reactist.

As I prepared to unify all `Badge` uses in Todoist with the one from Reactist, I also realized that this one needs some improvements (most prominently, making it render as an inline element).

Other change: this removes the `text-transform: uppercase` style. According to most of our references to badges in design mockups, we're not expecting our badges to be all uppercase anymore.

## PR Checklist

## Versioning

New minor release.

### DIFF
--- a/src/new-components/badge/badge.module.css
+++ b/src/new-components/badge/badge.module.css
@@ -2,7 +2,7 @@
     --reactist-badge-neutral-background: rgb(238, 238, 238);
     --reactist-badge-neutral-foreground: rgb(102, 102, 102);
 
-    --reactist-badge-positive-background: rgba(5, 133, 39, 0.1);
+    --reactist-badge-positive-background: rgb(224, 241, 228);
     --reactist-badge-positive-foreground: rgb(5, 133, 39);
 
     --reactist-badge-color-background: rgb(250, 234, 209);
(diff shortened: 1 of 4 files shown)


src/new-components/badge/badge.module.css
```diff
@@ -2,7 +2,7 @@
     --reactist-badge-neutral-background: rgb(238, 238, 238);
     --reactist-badge-neutral-foreground: rgb(102, 102, 102);
 
-    --reactist-badge-positive-background: rgba(5, 133, 39, 0.1);
+    --reactist-badge-positive-background: rgb(224, 241, 228);
```
gnapse, line 5: This change preserves the same color. The thing is, having had the color specified with some transparency, it worked great when the background is white, but when the background is something else, the color changed in unwanted ways.

So I changed the color to be the one that appears when on white, but hard-coded and not relying on transparency-over-white-bg to get it right. Now, the color is the same regardless of the background on which the badge appears.

PR #742, opened 2023-01-12
feat: Improved tooltip, support for arrow in tooltips

gnapse:
## Short description

See CHANGELOG in the diff for a summary of the changes in this PR.

## PR Checklist

## Versioning

New major release, as this introduces a breaking change (mentioned above).

## Demo

image


engfragui:
@gnapse Do you expect reviewers to test this in Reactist via Storybook (which it's what I did) or also trying to use the component in todoist-web (which I have not done)?

CHANGELOG.md
```diff
@@ -2,6 +2,15 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# next
+
+-   [BREAKING] Renames the CSS variable for primary text content inside toasts and tooltip. Previous name was `--reactist-content-toast`, now replaced with the new name `--reactist-toast-content-primary`.
+-   [Feat] New CSS variable to control the color of secondary text inside toasts and tooltip: `--reactist-toast-content-secondary`.
+-   [BREAKING] The `Tooltip` component no longer accepts generic props to be forwarded as generic HTML attributes to the tooltip content container. In particular, it no longer accepts a `className` to customize its appearance.
```
engfragui, line 9: 💭 From what you've seen in todoist-web and twist-web, does it look like there are occurrences that will have to be updated (i.e. stop passing `className`)?

engfragui (approved):

src/components/styles/constants.less
```diff
@@ -9,7 +9,6 @@
 @color_grey_10: #ececec;
 @color_grey_11: #fafafa;
 @color_border_grey: #dcdcdc;
-@color_almost_black: #333333;
```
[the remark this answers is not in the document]
gnapse: It's no longer being used. After having transitioned the tooltip to the new way of styling based on CSS variables within the newer Design System components, it now gets its colour from a CSS variable (the same one that colors the toasts, BTW). So this older approach of styling via .less variables is no longer used.

src/new-components/tooltip/tooltip.tsx
```diff
@@ -1,23 +1,74 @@
...
+     *
+     * It should be an interactive element accessible both via mouse and keyboard interactions.
+     */
     children: React.ReactNode
+
+    /**
+     * The content to show in the tooltip.
+     *
+     * It can be rich content provided via React elements, or string content. It should not include
+     * interactive elements inside it. This includes links or buttons.
```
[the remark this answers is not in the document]
gnapse: Correct. Although apparently similar to other components that allow to do this, a tooltip is not meant for offering interactions inside it. Semantically it is not allowed (e.g. on the WCAG/aria specs, for instance). A tooltip-like element that does allow for interactions should technically be a `role="dialog"` (a non-modal `dialog`). I'm planning to offer a standardized implementation via [ariakit's hover card](https://ariakit.org/components/hovercard).

CHANGELOG.md
gnapse: I have found some in twist-web though. Maybe I'll add a `exceptionallySetClassName` backdoor for now, to make future upgrades easier.

engfragui: I just checked (since you asked) and found three:

Todoist:
- https://github.com/Doist/todoist-web/blob/main/src/react_components/empty-state/empty-state-tooltip.tsx#L17

Twist:
- https://github.com/Doist/twist-web/blob/main/src/components/markdown-content/renderers/link-renderer-channel.tsx#L35
- https://github.com/Doist/twist-web/blob/main/src/components/markdown-content/renderers/link-renderer-thread.tsx#L53

gnapse:
> Do you expect reviewers to test this in Reactist via Storybook (which it's what I did) or also trying to use the component in todoist-web (which I have not done)?

Unless there's an explicit mention of this in a test plan in the PR description, I do not think it's necessary. Though it's always welcome if the reviewer thinks it's worth it to double-check.

CHANGELOG.md
gnapse: Yeah, those are a good find (I'm talking about the ones you found in Todoist). Especially because it's the kind of case where I would simply remove the custom class name, or at least check that I can do it without much consequences. A custom border and shadow on a tooltip is something we should not keep. So my thinking when being aggressive and removing the capabality to pass custom class names is that we could even remove our current instances where we do it.

Anyway, I added the `exceptionallySetClassName` backdoor for now. Which still does not mean I'll keep the current instances where we do this. I'll still try to get rid of as many cases as I can.

PR #743, opened 2023-01-13
fix: Set pointer-events none for button elements to ensure GA tracking

engfragui:
## Short description

While reviewing https://github.com/Doist/todoist-web/pull/5709, [Ernesto suggested](https://github.com/Doist/todoist-web/pull/5709#discussion_r1068081791) that, instead of sprinkling our todoist-web code with:
```
button[data-gtm-id] span {
    pointer-events: none;
}
```
and
```
a[data-gtm-id] span {
    pointer-events: none;
}
``` 
to make sure that our GA tracking works well on buttons, we can actually set that in Reactist ("at the source") so that we don't need to bother with it in todoist-web.

The result is that the inner `span` of any button element (both Reactist's `Button`s and `ButtonLink`s) has the `pointer-events: none` property set:

Screenshot 2023-01-13 at 15 22 35

### Some more background

If we don't make this change (a.k.a. if we don't add `pointer-events: none` to our buttons), the user click will happen on the span nested inside the `button` (for `Button`s) or inside the anchor tag `a` (for `ButtonLink`s). 

However, since the `data-gtm-id` prop needed for our Google Analytics tracking gets set on the `button` or anchor tag `a` (and not on the span nested inside them), the tracking of the user click will not work. 

What we do instead is to make the span non-clickable so that the click will happen on the `button` or anchor tag `a` instead, which is achieved by setting `pointer-events: none` on the elements inside the button.

## PR Checklist

## Versioning

Just a patch bump (v17.10.0 -> v17.0.1) as this is not really a new feature. However, I would not bet my family on the fact that this doesn't break anything for clients... 😅 So I appreciate any input.


src/new-components/button-link/button-link.test.tsx
```diff
@@ -240,6 +240,24 @@ describe('ButtonLink', () => {
         expect(buttonLink).not.toHaveClass('align-end')
     })
 
+    it('contains a non-clickable nested span', () => {
```
engfragui, line 243: Note that this test fails, as the `onClick` function gets called anyway. I was hoping that, if `pointer-events: none` was set, then the `onClick` function would not been called, but I guess I'm wrong 😅 Trying to think of another way to test that the nested span is non-clickable... 🤔

src/new-components/base-button/base-button.module.css
```diff
@@ -170,6 +170,14 @@
     margin-right: -6px;
 }
 
+/**
+ * Make sure that any `span` nested inside our button elements is non-clickable.
+ * This ultimately ensures that GA tracking works successfully for button elements.
+ */
```
engfragui, line 176: Note to reviewer: this is the actual change in this PR.

src/new-components/button-link/button-link.test.tsx
gnapse: The reason is that in jest tests, the stylesheet is not loaded. So the `pointer-events: none` style is not applied. I took the liberty to push an update to this branch that removes the test, and it also applies the new style rule to any element inside the button, and not just to the span. There’s for instance, the icons that can possibly be inside a button.

PR #746, opened 2023-01-18
chore: Introduce new component template

gnapse:
## Short description

Introduces a new tool called [plop](https://plopjs.com/), leveraged in this PR to provide an easy way to create bare-bones source files to implement a new component. The generated source files include the component implementation with sample props and styles, plus a small test file and storybook source files as well.

## Demo

https://user-images.githubusercontent.com/15199/213154676-7926118f-f772-420f-baf8-cdae93efda1d.mp4

## Test plan

1. Run `npm install`
2. Run `npm run plop component`[^1]
3. Provide a name for your component (e.g. `dropdown select`, `NavListItem` or `sortable-table`)
4. Leave the other two prompts at their default value by pressing "Enter" twice

[^1]: Alternatively, you can `npm install -g plop` and then you can use the command `plop component` directly.

Now check the following:

- [x] A new folder was created with all the new source files in it (e.g. check by running `git status`). Check the new source files if you're curious.
- [x] Run `npm run test:watch` to see the new minimal component test in action. The test should pass.
- [x] Run `npm run storybook` and check the stories for the new component. Stories should work and allow you to change the props and see the visual effects (in the "Playground" story for your new component).

## Versioning

No need to release. It's an internal change for devs only, that does not change what the library provides to its consumers.

### DIFF
--- a/plopfile.mjs
+++ b/plopfile.mjs
@@ -1,0 +1,90 @@
+// See https://plopjs.com/
+
+/** @typedef {import('plop').NodePlopAPI} NodePlopAPI */
+/** @typedef {import('plop').ActionConfig} ActionConfig */
+
+/**
+ * @param {string} name
+ * @returns string
+ */
+function templateFile(name) {
+    return `.plop/templates/${name}.hbs`
+}
+
+/** @param {NodePlopAPI} plop */
+function componentGenerator(plop) {
+    plop.setGenerator('component', {
+        description: 'React component with tests, styles, and storybook docs',
+
+        prompts: [
+            {
+                type: 'input',
+                name: 'name',
+                message: 'Component name',
+            },
+            {
+                type: 'input',
+                name: 'location',
+                message: 'Component folder (inside src/)',
+                default: 'new-components',
+            },
+            {
+                type: 'list',
+                name: 'storyType',
+                message: 'Story type',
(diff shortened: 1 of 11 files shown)


plopfile.mjs
```diff
@@ -0,0 +1,90 @@
...
+        prompts: [
+            {
+                type: 'input',
+                name: 'name',
+                message: 'Component name',
+            },
+            {
+                type: 'input',
+                name: 'location',
+                message: 'Component folder (inside src/)',
```
Bloomca, line 28: I feel it is worth to leave a note that you probably want to use the default folder here.

plopfile.mjs
```diff
@@ -0,0 +1,90 @@
...
+                default: 'new-components',
+            },
+            {
+                type: 'list',
+                name: 'storyType',
+                message: 'Story type',
+                choices: ['.mdx', '.tsx', 'both'],
```
Bloomca, line 35: Probably the same about choosing the default when in doubt.

Bloomca (approved):
Works as expected. Maybe we can improve messages on steps a bit, but it's minor.

plopfile.mjs
gnapse: Maybe we do not need to ask for this info at all. I was always in doubt to include it, and did it mostly as a proof of concept of what's possible.

PR #747, opened 2023-01-19
fix: Allow z-index override for Tooltip

frankieyan:
## Short description

This allows `Tooltip` to have its z-index overridden, as it can end up being hidden behind elements on a higher z-index:

image

## PR Checklist

## Versioning

Patch


src/new-components/tooltip/tooltip.module.css
```diff
@@ -1,4 +1,8 @@
+:root {
+    --reactist-tooltip-zindex: 1000;
```
gnapse, line 2: _(This is the comment for which I'm requesting changes, I do not recall the emoji we agreed to use for this)_

We already use a naming convention for zindex-related variables: `--reactist-stacking-order-*`

Also, we already have a convention of not declaring its value, and use a default value when the variable is used. See my next comment below.

src/new-components/tooltip/tooltip.module.css
```diff
@@ -1,4 +1,8 @@
...
+}
+
 .tooltip {
     text-overflow: ellipsis;
-    z-index: 1000;
+    z-index: var(--reactist-tooltip-zindex);
```
gnapse, line 7: _(optional but encouraged for consistency)_

We already have a convention of not declaring its value, and use a default value when the variable is used. See my next comment below.

```suggestion
    z-index: var(--reactist-stacking-order-tooltip, 1000);
```

This includes removing the declaration in `:root`.

Alternatively, we might decide to move in the other direction and declare all stacking order variables explicitly. No need to do it in this PR, though.

src/new-components/tooltip/tooltip.stories.tsx
```diff
@@ -127,3 +128,42 @@ TooltipRichContentStory.argTypes = {
...
+                    </Tooltip>
+                </Box>
+            </Stack>
+        </>
+    )
+}
+
+function PortalToHead({ children }: React.PropsWithChildren<unknown>) {
+    return createPortal(children, document.head)
+}
```
gnapse, line 169: nit: I do not think we need to create a story for this.

gnapse (changes requested):

gnapse:
@frankieyan while you're at it, this PR is a great opportunity to add a new style to tooltips that is being requested: setting a max-width for them, as requested [here](https://twist.com/a/1585/ch/634263/t/4031093/c/81279438).

> Yes, a tooltip width limit would be great. In our library we have it set to 250px. Maybe 300px works too 🤔

Do you think we can slip in that change here? It would be hardcoded in the tooltip styles. I do not think we need to give the ability to customize this via a CSS variable. But maybe we do and you prefer to do it that way. What do you think?

frankieyan:
> here

I don't mind doing it, but I don't think we should include it in the changes here to keep it focused. I'll create a separate PR.

src/new-components/tooltip/tooltip.stories.tsx
frankieyan: I'm inclined to keep it, not just to provide this tidbit but also as a way to test the bug/fix in storybook. Let me know if you feel more strongly here 👍

src/new-components/tooltip/tooltip.module.css
frankieyan: I went with your second suggestion and pulled all the stacking-order properties over to `default-styles.css` 👍

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# Next
+
+-   [Fix] `Tooltip`s can now have its z-index customized using the `--reactist-tooltip-zindex` custom property.
```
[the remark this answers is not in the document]
frankieyan: Yes! Noticed this while I was cutting the release and ninja fixed it 👍

PR #750, opened 2023-01-26
feat: New Banner component

engfragui:
Fixes https://github.com/Doist/Issues/issues/9215

## Short description

New Banner component! 🎉

https://user-images.githubusercontent.com/1509326/214834007-542e1905-8631-4faf-8c25-5ae03037afeb.mp4

----

Note: [in this piece of code](https://github.com/Doist/todoist-web/blob/main/src/react_components/project-view/project-view-header.tsx#L336-L371) you can see the current implementation of this type of banners that we're already using in production. 

Our new Reactist banner component needs to be equivalent (or very close) so that we can use in todoist-web instead of the current custom implementation.

- [x] Waiting for design answer here https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk?node-id=8321:272616#353688502

## Reference

- [Design mocks](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product-Library---Web?node-id=8321%3A272616&t=b3zdZZ8napvDN8lz-0)

## PR Checklist

## Versioning

New feature (MINOR bump: v19.0.1 -> v19.1.0)


src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,79 @@
...
+                <Column>
+                    <Box paddingY="xsmall" paddingRight={'small'}>
+                        {description ? (
+                            <Box className={[styles.title, styles[`title-${tone}`]]}>{title}</Box>
+                        ) : (
+                            <Box
+                                className={[
+                                    styles.title,
+                                    // If the banner does not have a description, we need to slightly tweak
+                                    // the styling of the title applying an extra css class
```
engfragui, line 57: Note to self: Errr there must be a better way to do this 😅

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,80 @@
...
+     */
+    action?: React.ReactChild
+}
+
+function Banner({ id, tone, icon, title, description, action, ...props }: BannerProps) {
+    return (
+        <Box
+            {...props}
+            id={id}
+            role="banner"
```
engfragui, line 38: A note: if I set this `role="banner"` and I try to render more than one banner at the time (i.e. one stacked top of each other) I see that the a11y doesn't pass:

```
● Banner › renders with no a11y violations

    expect(received).toHaveNoViolations(expected)

    Expected the HTML found at $('.banner-info') to have no violations:

    <div role="banner" aria-live="polite" class="banner banner-info box borderRadius-standard">

    Received:

    "Document should not have more than one banner landmark (landmark-no-duplicate-banner)"

    Fix any of the following:
      Document has more than one banner landmark

    You can find more information on this issue here:
    https://dequeuniversity.com/rules/axe/4.2/landmark-no-duplicate-banner?application=axeAPI

    ────────

    Expected the HTML found at $('.banner-info') to have no violations:

    <div role="banner" aria-live="polite" class="banner banner-info box borderRadius-standard">

    Received:

    "Ensures landmarks are unique (landmark-unique)"

    Fix any of the following:
      The landmark must have a unique aria-label, aria-labelledby, or title to make landmarks distinguishable

    You can find more information on this issue here:
    https://dequeuniversity.com/rules/axe/4.2/landmark-unique?application=axeAPI

      79 |         )
      80 |         const results = await axe(container)
    > 81 |         expect(results).toHaveNoViolations()
         |                         ^
      82 |     })
      83 | })
      84 |

      at Object.toHaveNoViolations (src/new-components/banner/banner.test.tsx:81:25)
```

More info at this link: https://dequeuniversity.com/rules/axe/4.2/landmark-no-duplicate-banner?application=axeAPI

> The role="banner" ARIA landmark should only appear once on an element that is unique to the HTML document

src/new-components/banner/banner.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+            {buttonText}
+        </ButtonLink>
+    )
+}
+
+export function BannerWrapper({ tone, icon, title, description, action }) {
+    return (
+        <Banner tone={tone} icon={icon} title={title} description={description} action={action} />
+    )
+}
```
gnapse, line 74: What's the point of this wrapper if all it does is forward the props without doing anything else?

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,79 @@
...
+     */
+    icon: React.ReactChild
+    /**
+     * The title to be displayed at the top of the Banner.
+     */
+    title: string
+    /**
+     * An optional description to be displayed inside the Banner.
+     */
+    description?: string
```
gnapse, line 26: I tend to prefer to declare these as type `React.ReactNode` instead.

```suggestion
    /**
     * The title to be displayed at the top of the Banner.
     */
    title: React.ReactNode
    /**
     * An optional description to be displayed inside the Banner.
     */
    description?: React.ReactNode
```

It enables passing rich text, for instance, containing links or bold or emphasized text:

```jsx
<Banner
  title={<>This is really <strong>important</strong></>}
  description={<>If you're curious, <a href="#">learn more</a>.</>}
/>
```

src/new-components/banner/banner.tsx
gnapse: If I understood correctly, this is about making the title bold only when a description is present, right? If so, check out how the `Toast` component does it:

https://github.com/Doist/reactist/blob/62102eb395c7e121904461d44646ff261c96a948/src/new-components/toast/static-toast.tsx#L96-L103

It uses the `Text` component, though. I'm not sure how that would affect what you're doing here. For instance, the `Text` component would need some work to make it adopt the text colors you need. I have some ideas on how to do that somewhat decently, but I have no hard objections to this approach. We can ship it and improve it later if needed. As long as it outwardly achieves what it's supposed to, that's ok.

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,79 @@
...
+        <Box
+            id={id}
+            role="banner"
+            aria-live="polite"
+            borderRadius="standard"
+            className={[styles.banner, styles[`banner-${tone}`]]}
+        >
+            <Columns space="medium" alignY="center">
+                <Column width="content" aria-hidden>
+                    <Box display="flex" className={styles.icon} aria-hidden>
```
gnapse, line 45: I do not see this `.icon` class in the css module source. I think we can remove this. Also, why do we need this `display="flex"` for? Furthermore, do we really need the wrapper box around the icon? (not sure about that last one, just asking)

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,79 @@
...
+                        {icon}
+                    </Box>
+                </Column>
+                <Column>
+                    <Box paddingY="xsmall" paddingRight={'small'}>
```
gnapse, line 50: It looks fishy that we need to have this `paddingRight` here. Why isn't the inter-column spacing enough?

src/new-components/banner/banner.stories.mdx
```diff
@@ -0,0 +1,234 @@
...
+</Canvas>
+
+## Examples
+
+<Canvas>
+    <Story
+        name="Main demo"
+        parameters={{
+            docs: { source: { type: 'code' } },
+            chromatic: { disableSnapshot: false },
```
gnapse, line 165: Nice! I like that you realized that we want to enable visual regression testing in this kind of story that showcases the different look and feels all at once. It's good to know that that was somehow clear, even though I have not documented yet our "Chromatic best practices".

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,79 @@
...
+     * An optional description to be displayed inside the Banner.
+     */
+    description?: string
+    /**
+     * An optional action to displayed inside the Banner.
+     */
+    action?: React.ReactChild
+}
+
+function Banner({ id, tone, icon, title, description, action }: BannerProps) {
```
gnapse: Anticipating that we might need it, I'd recommend that you enable forward-refs to this component's container:

```tsx
const Banner = forwardRef<HTMLDivElement, BannerProps>(
    function Banner({ … }, ref) {
        // pass ref={ref} to the outer `Box`
    }
)
```

gnapse (changes requested):
This is looking good so far. Great work!

Main reason for requesting changes is the comment about the types for the `title` and `description` props, which I think are a hard requirement (e.g. when we pass translations that need the `<Interpolate />` component in Todoist).

The rest of the comments vary in optionality.

src/new-components/banner/banner.tsx
engfragui: This makes a lot of sense! 😅 I added a unit test and also an example in the stories. Which lead me to:

Screenshot 2023-01-26 at 16 07 15

Screenshot 2023-01-26 at 16 07 19

Is this (colors for href) something that I should already ask Design team about or should we just move forward since we don't have any use cases (in todoist-web) at the moment?

src/new-components/banner/banner.tsx
engfragui: Let me test all of this. I think I was playing with different paddings and the like and forgot to clean things up.

src/new-components/banner/banner.stories.mdx
engfragui: I mean, thank you for putting a lot of respect and trust in me, but I must admit that I would've put this even if the story didn't have all the "variants" 😅

Actually, now that I think about it, maybe this `chromatic: { disableSnapshot: false },` should go even on the dark mode story that I added at the end of this file. Why. not?

src/new-components/banner/banner.tsx
gnapse: Wow, it's good that we have these jest axe tests in place.

Indeed, it seems this component should not use `role="banner"`. In spite of our choice of name, in terms of aria roles, "banner" seems to something else entirely. According to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Banner_Role):

> The `banner` role is for defining a global site header, which usually includes a logo, company name, search feature, and possibly the global navigation or a slogan. It is generally located at the top of the page.

So it's nothing like what you are developing here.

Looking at the list of existing aria roles, I think this component is in line with [`role="note"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/note_role). Check that out, and if you agree, make the change.

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,82 @@
...
+            className={[styles.banner, styles[`banner-${tone}`]]}
+        >
+            <Columns space="medium" alignY="center">
+                <Column width="content" aria-hidden>
+                    {icon}
+                </Column>
+                <Column>
+                    <Box paddingY="xsmall">
+                        {description ? (
+                            <Box className={[styles.title, styles[`title-${tone}`]]}>{title}</Box>
```
gnapse, line 54: If you go with the `role="note"` suggestion, then I think we should make sure that the title of the banner becomes the semantic accessible name for the note element. To ensure this, please add a test that queries the element by role and accessible name:

```tsx
it('uses the title as the accessible name', () => {
  render(<Banner title="Hello world" />)
  expect(screen.getByRole('note', { name: 'Hello world' })).toBeInTheDocument()
})
```

If this does not work out of the box right now (which is most likely the case), then we may need to associate the title element to the container via `aria-labelledby`. Here's some pseudo-code:

```tsx
function Banner(props) {
  const titleId = useId() // import { useId } from 'react'
  return (
    <Box role="note" aria-labelledby={titleId} {…}>
      <Box id={titleId}>{title}</Box>
      // …
    </Box>
  )
}
```

Hope it helps.

src/new-components/banner/banner.tsx
engfragui: @gnapse Thank you for all the context! 🙏 I think `role="note"` is perfect, as this is just something slightly more prominent, that we want the users to pay attention to, but nothing more.

Does this in any way suggest that we should rename our new reactist component to something else? I felt like `Banner` worked well, but I also feel like I don't know English enough maybe.

gnapse: Let's ask @craigcarlyle, not only as a native English speaker, but as our accessibility DRD.

src/new-components/banner/banner.tsx
gnapse: Oh, and to round things up, then the description (when present) should become the accessible description of the element. Using `aria-describedby` along the same lines.

engfragui: Indeed, it didn't workout _out of the box_, but then I put the appropriate props _in the Box_ and it worked! 😅

Thank you, all your notes were perfect. 👌

src/new-components/banner/banner.test.tsx
```diff
@@ -94,8 +94,30 @@ describe('Banner', () => {
...
+    it('uses the description as the accessible description', () => {
+        render(
+            <Banner
+                tone="info"
+                icon={'💚'}
+                title={'Hello World'}
+                description={'Welcome to the world, Linus!'}
+            />,
+        )
+        expect(screen.getByRole('note', { name: 'Hello World' })).toHaveAccessibleDescription()
```
gnapse, line 116: You can assert the content of the description is the expected one

```suggestion
        expect(screen.getByRole('note', { name: 'Hello World' })).toHaveAccessibleDescription(
            'Welcome to the world, Linus!',
        )
```

src/new-components/banner/banner.tsx
gnapse: What if you use `<TextLink />` instead of `<a />`? Though it probably won't get the right color either, because we do not have colors for dark mode here in Reactist. Unless you also provide the necessary CSS variable override as you do with the rest of the colors for dark mode in this story.

gnapse (approved):
I can approve already, even though we may still wait for Craig's input on the component name, and any other minor issue.

src/new-components/banner/banner.tsx
craigcarlyle: First of all, I'm happy to see that the automated tests caught this 🎉 

I think setting the `role` to `note` is okay, but thinking about our use cases maybe we should follow the example of the [Banner component in the Polaris design system](https://polaris.shopify.com/components/banner) and: 

* Set the `role` to `status`
* Set the `aria-live` property to `polite`
* Add `tabindex="0"` (so that it can be navigated to with a keyboard)

This means that users with screen reading software would:
* Get notified eventually that they're viewing an achieved project
* Get notified eventually that they're previewing a project
* Get notified eventually that they've hit a project limit

If we went with the `note` option the user would have to manually navigate to the content to discover it instead of the content being announced automatically. I think it may be the better option.

And to answer the last question, I think the "Banner" name is fine. The name is for developers rather than the end consumer 🙂

gnapse: @craigcarlyle, name-similarities aside, it does not look like Polaris' component is meant for the same use cases as ours.

- That one looks more like something that appears triggered by some user action (judging by that default example of "Order archived") and it is dismissable.
- Ours is meant to appear from the start if the conditions for it are met, and it is not dismissable, but stays around. At least not dismissable by clicking a close button. The only way you can dismiss a "This project is preview only" is to click on "Join project", a permanent action that removes the condition that made the banner appear.

I'm not saying these are reasons good enough to not use `role="status"`, but I do see the difference with what, at a glance, the Polaris' component seems to be about. With all this in mind, do you still think we should use that option? The part about this message being announced might be interesting. But, again, this message does not suddenly appear due to a user action.

gnapse: Actually, that Polaris component seems more like Reactist's `Alert`, a component that is visually similar to this one, but with a different semantics. Check out [this comment](https://github.com/Doist/Issues/issues/7617#issuecomment-1397364983) where I proposed to make these different things in the design system.

craigcarlyle: @gnapse I still believe it's the better option and I think it's differentiated between the `Alert` component enough with the role difference (`status` vs `alert`). However, I don't feel too strongly about it.

Looking at Polaris closer, our use case may be closer to the [Callout card](https://polaris.shopify.com/components/callout-card) component which has no special `role` or `aria-*` properties at all. Maybe we're overthinking it? 🤷‍♂️

gnapse: Yes, viewed under that light I see your point. Comparing `role="status"` vs. `role="alert"`, I see this about `status`:

> The `status` role defines a [live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) containing advisory information for the user that is not important enough to be an [alert](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/alert_role).

[source](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/status_role)

Thanks!

What do you think @engfragui? Do we have a winner? 😄

engfragui: Thank you both for this discussion! 🙏 Let's go with role "status", which feels like in between "alert" and "note" and might fit well with our intent of "we want the user to pay attention to this, but not really alert them".

Just to confirm, this is what I have:

```
            role="status"
            aria-labelledby={titleId}
            aria-describedby={descriptionId}
            aria-live="polite"
            tabIndex={0}
```

engfragui:
Looking at https://github.com/Doist/todoist-web/pull/5846 (whose design is from [this design mock](https://www.figma.com/file/ypFhMfTimwHRViD9zAZhAD/TD-Teams-Billing?node-id=67%3A197199&viewport=1173%2C791%2C1&t=to3e1o5mAcq1e828-0)) it seems like we might need/want to add some sort of "subtle" tone as well to our Banner (no background, gray button).

Happy to do this as a follow-up PR, possibly after getting feedback/confirmation from the Design team (https://www.figma.com/file/ypFhMfTimwHRViD9zAZhAD?node-id=67:197199#355231551).

engfragui:
Status update: I'm mostly just trying to get the final requirements from Design team (see this comment https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk?node-id=8321:272616#353688502)

src/new-components/banner/banner.tsx
```diff
@@ -0,0 +1,101 @@
...
+            borderRadius="standard"
+            className={[styles.banner, styles[`banner-${tone}`]]}
+        >
+            <Columns space="medium" alignY="center">
+                <Column
+                    width="content"
+                    aria-hidden
+                    style={{
+                        /* Make sure the icon is centered vertically */
+                        lineHeight: 0,
```
engfragui, line 59: This comes from the existing banner-like component we implemented in todoist-web https://github.com/Doist/todoist-web/blob/main/src/react_components/project-view/project-view-header.tsx#L349-L353.

I didn't have this at first and the icons were placed slightly off center, which was somewhat noticeable for banners without description (only title).

PR #755, opened 2023-02-17
fix: Change Badge component to bold + set explicit line height

engfragui:
## Short description

This PR tweaks our (existing) Badge component:
1. Content of the Badge is now bold (as per [this request on Twist](https://twist.com/a/1585/ch/640947/t/4158034/c/81940918)). We should be able to see this very clearly via our visual regression tests, but here's some screenshots:
<table>
<tr><td>Before</td><td>After</td></tr>
<tr><td>
Screenshot 2023-02-17 at 12 57 43
</td><td>
Screenshot 2023-02-17 at 12 25 37
</td></tr>
</table>

2. The appropriate line-height is enforced. No line-height set was causing the background of the badge to expand vertically when inside flex containers:
Screenshot 2023-02-17 at 12 51 43
Although this can be easily set in the client, I figured we would want to enforce this in Reactist _once and for all_.

I'm also taking this chance to fix a couple of typos in the Banner stories (25 instead of 24 px -- I was probably drunk, who knows 🤷‍♀️).

## PR Checklist

## Versioning

I think this is a minor fix, so I'm just bumping the patch.


engfragui:
> We should be able to see this very clearly via our visual regression tests

Errr not true (see [this build](https://www.chromatic.com/build?appId=638e186cd0b36e1276f510de&number=57) -- you can see that there's no side by side comparison of all the badges). Looking into it now 👀

src/badge/badge.module.css
```diff
@@ -16,14 +18,15 @@
...
+    font-weight: var(--reactist-font-weight-strong);
     font-size: var(--reactist-badge-font-size);
     text-transform: uppercase;
     letter-spacing: 0.1em;
     border-radius: 3px;
-    padding: 1px var(--reactist-spacing-xsmall);
+    padding: var(--reactist-badge-paddingY) var(--reactist-badge-paddingX);
     color: var(--reactist-badge-tint);
     background-color: var(--reactist-badge-fill);
+    line-height: calc(var(--reactist-badge-font-size) + 2 * var(--reactist-badge-paddingY));
```
gnapse, line 29: I think it would be a bit simpler if instead of applying a vertical padding, we just apply a line height and that's it. That alone would enforce the height and the vertical spacing around the text, even if it's not padding, right?

With that in mind, I would also hard-code the line-height to be 12px, instead of doing all this calculation. Something along these lines:

```css
.badge {
  padding: 0 var(--reactist-spacing-xsmall);
  line-height: calc(var(--reactist-badge-font-size) + 2);
}
```

I also feel that offering badge-specific CSS variables for the padding of the badge may be unnecessary. I'd rather add that API later if needed, than committing to it today. Any new CSS variables we add becomes public Reactist API that we later need to support or deprecate if we want to remove it.

gnapse (approved):

gnapse:
@engfragui upon inspecting the Chromatic snapshots:

CleanShot 2023-02-17 at 10 03 46@2x

I find it odd that in there the vertical padding looks almost non-existent. Could it be that 1px is too little?

engfragui:
> Could it be that 1px is too little?

@gnapse Errr the [Figma mocks](https://www.figma.com/file/ZtwbWZH9f8l0aWfPolQ90E/23H1-%7C-Todoist-Fun%3A-Pro-in-product-education-(Mockups)?node-id=558%3A275021&t=jJk7eByVSQsZzQAP-0) say the padding should be 3px 😅

Screenshot 2023-02-17 at 14 06 16

I think it still looked good because the background was anyway expanding vertically to fill the entire vertical height, so we never noticed this. I will change to 3px.

src/badge/badge.module.css
engfragui: Err actually it seems like I still need to set the padding value to what I actually want the padding to be.

Screenshot 2023-02-17 at 14 19 00

vs.

Screenshot 2023-02-17 at 14 18 50

PR #757, opened 2023-02-20
fix: Tweak badge height as per design mocks

engfragui:
## Short description

Minor follow-up to https://github.com/Doist/reactist/pull/755, after testing the new version in todoist-web. Looks like this:

Screenshot 2023-02-20 at 16 42 11

which looks _too tall_.

However, what we know is that:
- font size should be (and is) 10px
- height of the badge should be 15px ([as per mocks](https://www.figma.com/file/ZtwbWZH9f8l0aWfPolQ90E/23H1-%7C-Todoist-Fun%3A-Pro-in-product-education-(Mockups)?node-id=408%3A250997&t=dpdsRmaA2k7Ccnfu-0))

Therefore the padding should be 2.5px at the top and 2.5px at the bottom. I had 3px and 3px before this change. Apologies, I hadn't checked the mock carefully enough apparently.  

## PR Checklist

## Versioning

Patch bump.


engfragui:
@gnapse I'm running into an issue in the [Chromatic build](https://github.com/Doist/reactist/actions/runs/4224987339/jobs/7336654969):

```
Error: error:0308010C:digital envelope routines::unsupported
```

Based on ["the internet"](https://stackoverflow.com/questions/69692842/error-message-error0308010cdigital-envelope-routinesunsupported), it seems like this is due to fact that that build is running on `"nodeVersion": "v18.14.1",`.

Have you encountered this before? I feel like a possible solution could be to ensure that we're using node v16 in the chromatic configs https://github.com/Doist/reactist/blob/main/.github/workflows/chromatic.yml.

engfragui:
As per @gnapse, it seems like the actual culprit might instead be the error that shows up at the beginning of the build (instead of the one I was looking at):

```
ℹ Missing commit detected
When detecting git changes for TurboSnap, we couldn't find the commit (686f3ea) for the most recent build (#63).
To avoid re-snapshotting stories we know haven't changed, we copied from the most recent build (#55) that did have a commit (b5dd3e6) instead.
```

engfragui:
Fix for the Chromatic build is in https://github.com/Doist/reactist/pull/758

PR #770, opened 2023-05-11
BREAKING: Add startSlot and endSlot to TextField

gnapse:
## Short description

-   [BREAKING] `TextField`'s `startIcon` property is now renamed `startSlot`.
-   [Feat] New `startSlot` and `endSlot` props on the `TextField` component. These support rendering not only icons, but buttons too.
-   [Tweak] `PasswordField`'s toggle visibility button is now implemented by using the `TextField`'s `endSlot` feature.

About the last one in that list, I want to note that this change allowed me to remove the `PasswordField`'s stylesheet entirely, and also the change was so seamless that the `PasswordField`'s test file needed no changes at all. It brings in that feeling that things are falling more into place with this direction we're taking.

## Demo

https://github.com/Doist/reactist/assets/15199/65b096a0-69df-454b-b3cb-55f9f8cf4cf0

## PR Checklist

## Versioning

New major version.


src/text-field/text-field.tsx
```diff
@@ -5,12 +5,13 @@ import styles from './text-field.module.css'
 import type { FieldComponentProps } from '../base-field'
 import { useMergeRefs } from 'use-callback-ref'
 
-type TextFieldType = 'email' | 'search' | 'tel' | 'text' | 'url'
+type TextFieldType = 'email' | 'search' | 'tel' | 'text' | 'url' | 'password'
```
frankieyan, line 8: I'm wondering if exposing the `password` type here would offer too much flexibility. 

Unless we want to be able to present a password field without the visibility toggle, I would try to make this more explicit by using a prop like `__reactist_internal_password_field?: boolean`, and letting that set the input's type as password instead.

frankieyan (approved):
Good change @gnapse 👍 I have one suggestion about the type prop for your consideration.

src/text-field/text-field.tsx
gnapse: I also thought about this, and was leaning towards not being too protective, hence my solution of adding "password" as a valid type in TextField. However, seeing that you also thought about this, and you lean towards being restrictive, I fixed it back to how it was.

However, I did not go down the path you propose. Such an internal prop name would be exposed in the TypeScript autocomplete of the `TextField` props, hence spilling a bit into the public API the fact that we have this obscure prop. Instead, I remove "password" as a valid type value in `TextField`, but I still pass it inside the implementation of `PasswordField`, overcoming the TypeScript error with a `@ts-expect-error` comment.

Issue #675, opened 2022-07-26
Improve tooling of reactist


gnapse:
[Wishlist for Reactist tooling/setup improvements](https://twist.com/a/1585/ch/190200/t/2996915/)

- [ ] Put thread into spec template / issue so it can be worked on, then move into "Ready to be worked on"

frankieyan:
A proposal from @pawelgrimm ([ref](https://twist.com/a/1585/ch/190200/t/6451124/c/92384016)) is to switch away from publishing Reactist as an npm package, and leveraging shadcn's cli - which lives in https://github.com/shadcn/ui - allowing the Reactist source files to be copied into consumer codebases in an automated. way. This avoids having to rely on the "link" mechanism when dealing with local npm packages and docker.

Pros/cons:

> - 👍 Much easier to customize
> - 👍 No real build process needed in Reactist (shadcn already provides the script to “compile” the registry)
> - 👍 Better DX (develop directly against todoist-web, then commit the same patch to Reactist)
> - 👎 Worse upgrade experience (manually integrate updates from Reactist into todoist-web)
> - 👎 Looser constraints (i.e. Reactist can’t impose any hard rules, they can be overridden easily within the consuming code)

To make this work, we'll have to enforce (whether automated or not) on the todoist-web side that Reactist source files are not to be manipulated, similar to how we treat other build artifacts like colour variables.

Next step: small PoC to demonstrate to the team how this works.

gnapse:
Thanks for summarizing that down, @frankieyan. It's really helpful.

About those cons.

> Worse upgrade experience (manually integrate updates from Reactist into todoist-web)

Why is this exactly worse? Assuming we have a script to do it, it should not be too different from increasing a version number in `package.json` and running `npm install`. Or am I missing something?

> Looser constraints (i.e. Reactist can’t impose any hard rules, they can be overridden easily within the consuming code)

I could see that as a positive, depending on how easily things can be overridden. As in "can they be inadvertently overridden easily?" or is it more like "they can be easily overridden, but doing it is explicit enough so that you are unlike to do it by accident".

By "overriding" you refer to what you talk about at the end, regarding the modification of the Reactist artifacts? If so, then I agree that we should do something to prevent that, and possibly we can do that. Or is the "overriding con" about other types of overrides?

gnapse:
I updated the check boxes above, checking off the ones about automated releases, after Rui took care of that in https://github.com/Doist/reactist/pull/863

PR #782, opened 2023-06-14
chore: Deploy storybook automatically, without relying on build artifacts in repo

### LINKED ISSUE #252: Do we need built assets in git
### ⚙️ Improvement

I wonder if we need to have checked in git all the files generated from a build. That hasn't been my experience working in other libraries. Though I have to admit I am not knowledgeable at all when it comes to all the build/config parts of the job so I'm not even sure what the implications are in terms of publishing, etc.

henningmu:
Closes #252 
Part of https://github.com/Doist/reactist/issues/675

## Short description

Previously, one would be required to run `npm run build-all` and commit the updated files from `docs/` folder to this repository. This made reviews harder, was cumbersome and error-prone. With this PR, it's a thing of the past. We're now building and deploying the storybook from a GitHub Action on each commit on `main`. Proper CD if you will 💫


gnapse:
Applause

Thank you

PR #784, opened 2023-06-15
refactor: Rework modal padding

pawelgrimm:
## Short description

In this PR, we are adding a new story for a basic confirmation modal, which is a very common use case for modals:
image

In addition, we are refactoring how we apply padding to the `Modal*` components such that the inner components (`ModalHeader`, `ModalBody`, and `ModalFooter`) are no longer responsible for knowing or declaring their own padding, as this can be more easily (and consistently) managed in the `Modal` component. Now, the `Modal` component applies its own outer padding and it uses the newish `Box.gap` prop to apply inter-child padding.

## PR Checklist

### DIFF
--- a/src/modal/modal.tsx
+++ b/src/modal/modal.tsx
@@ -230,9 +230,12 @@ export function Modal({
                         background="default"
                         display="flex"
                         flexDirection="column"
+                        flexGrow={height === 'expand' ? 1 : 0}
                         overflow="hidden"
                         height={height === 'expand' ? 'full' : undefined}
-                        flexGrow={height === 'expand' ? 1 : 0}
+                        padding="large"
+                        paddingTop="medium"
+                        gap="large"
                         className={[exceptionallySetClassName, styles.container]}
                         // Disable focus lock as we set up our own using ReactFocusLock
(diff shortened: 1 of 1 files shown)


gnapse:
<blockquote>

The main issue is that the default amount of padding in the Reactist modal components doesn't match that of examples in the Design system and the only way to get our modals to align is to use `exceptionallySetClassName`. For example, check out our delete-section-confirmation-modal vs the example in my OP vs the [Figma mockup](https://www.figma.com/file/LYlWNzvhMDh907l07mPPQk/Product-Library---Web?type=design&node-id=1802-345&mode=design&t=vdZdac3HZEcKprmh-4):

<table>
<tr>
<th>
Todoist

<th>
This PR

<th>
Figma

<tr>
 <td>

image

 <td>

image

<td>

image

</table>
</blockquote>

I may be seeing something wrongly, but from those 3 screenshots, the ones on the left (actual implementation as seen in Todoist) and the one and the right (Figma source of truth) look the most alike. While the one in the middle, from this PR is the one that I find different from the other two (and I'm not talking about the primary button color).

PR #789, opened 2023-07-05
style: Apply a large right-padding to ModalHeader to match left side

nats12:
## Short description

Recently after working on the edit filter modal, it became obvious that the right hand side padding of the modal's header was much smaller, especially when our modal body's padding is set to large. 

CleanShot 2023-07-05 at 10 24 00

I see we've done it to account for button padding, but have also witnessed this style being overriden in some cases in todoist  to align it with the left i.e. task detail header. 

I feel it would look better if we align these at the root component, and also that it's not necessarily this component's job to account for potentially large buttons within it. Instead, it should be handled by consumers if need be. Happy to discuss this further if there are any strong opinions on this change though. 

## PR Checklist

- [ ] The `ModalHeader` has a `large` padding on both right and left sides.

## Versioning

Small fix, bumped to`v21.1.1`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# v21.1.1
+
+-   [Fix] Apply even right and left paddings for `ModalHeader`.
+
 # v21.1.0
 
 -   [Feat] The `Prose` component now supports links applied to inline code.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@doist/reactist",
-    "version": "21.1.0",
+    "version": "21.1.1",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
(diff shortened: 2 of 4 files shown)


nats12:
Looping in @gnapse too as design system DRD.

scottlovegrove:
Unless this is needed urgently, I would recommend waiting for @pawelgrimm as he is working on something in the same area that could fix this as part of the overall changes he's making. https://github.com/Doist/reactist/pull/784

gnapse:
The reason I recall for the right-side padding to be smaller is to give the impression that the close button icon's is right-aligned. It is better explained with a couple of screenshots:

CleanShot 2023-07-05 at 09 25 29@2x

See how we want the icon inside the button to be the thing that is right-aligned with the modal's content. However, the button element is bigger than the button icon, as can be seen in the following screenshot:

CleanShot 2023-07-05 at 09 26 26@2x

However, that being said, I agree that hard-coding a smaller spacing to the right of the header might not be the best solution. But whatever other solution we find, it should keep in mind this need for alignment with the modal's close button icon.

cc @pawelgrimm to keep it in mind on that pull request that you're working on in Reactist.

nats12:
> Unless this is needed urgently

It is it not, so I will close this PR as it would be superseded by the changes @pawelgrimm is working on.

Thank you both for the context @scottlovegrove, @gnapse 👍🏼

PR #790, opened 2023-07-06
feat: Expose input border colours as --reactist-inputs-* variables

frankieyan:
## Short description

This separates the CSS variables used for dividers and graphical elements' borders, from those used for input fields' borders.

To document these variables in Storybook, the four field-based components have had their documentation converted to MDX:

image

By having this separation, we'd be able to take advantage of the new `inputs` colours from Design ([ref](https://twist.com/a/1585/ch/656510/t/4343516/c/83606339)).

## PR Checklist

## Versioning

Minor. It's a bit of a grey area as the new variables are instantiated with the old `--reactist-divider` colours, keeping backwards compatibility, and existing overrides using these variables should not break.


Bloomca (approved):
Change looks good. I am not sure what I need (if I need to do anything) on Chromatic, though.

frankieyan:
@Bloomca I am wondering if I should stick with the old story names (i.e. ending in "Story") so that Chromatic can match up the diffs, or leave it as is and accept the new diffs. Thoughts? cc. @gnapse 

|Before|After|
|-|-|
|image|image|

gnapse:
>  I am wondering if I should stick with the old story names (i.e. ending in "Story") so that Chromatic can match up the diffs, or leave it as is and accept the new diffs. Thoughts?

We should not keep a name only because of Chromatic. If we really find the name change to be what we want, we should do the change, even if inconvenient.

I do agree the name change is for the better.

PR #794, opened 2023-07-31
chore: Fix links to Reactist components in storybook stories

engfragui:
## Short description

While browsing a couple of Reactist storybook stories, I've noticed that there are some broken links that don't include the path "reactist":

https://github.com/Doist/reactist/assets/1509326/26b1d8cf-3c44-41c7-97aa-b26bcb4fa96e

This PR should fix those links so that do include the correct path and therefore lead the viewer to the right place. 

## PR Checklist

-   [ ] Added tests for bugs / new features
-   [ ] Updated docs (storybooks, readme)
-   [x] Executed `npm run validate` and made sure no errors / warnings were shown
-   [x] Described changes in `CHANGELOG.md`
-   [x] Bumped version in `package.json` and `package-lock.json` (`npm --no-git-tag-version version <major|minor|patch>`) [ref](https://docs.npmjs.com/cli/v6/commands/npm-version)
-   [ ] Reviewed and approved Chromatic visual regression tests in CI

## Versioning

Bumped patch.

PR #795, opened 2023-07-31
chore: Fix links to Reactist components in storybook stories using LinkTo

engfragui:
## Short description

Follow-up to https://github.com/Doist/reactist/pull/794 where I tried to fix a couple of broken links in our storybook stories.

> While browsing a couple of Reactist storybook stories, I've noticed that there are some broken links that don't include the path "reactist":
> 
> https://github.com/Doist/reactist/assets/1509326/26b1d8cf-3c44-41c7-97aa-b26bcb4fa96e

It turns out I didn't actually fix anything at all as the links are still missing the appropriate `/reactist` path which it seems not to be included in `window.location.origin` 😅

In this PR, I initially tried to address this by concatenating `window.location.origin` with `window.location.pathname` to (hopefully) get the correct Reactist URL path. However, that doesn't work either, because `pathname` will include `/iframe.html` as well (since the link is inside an iframe) so what I'm doing in this PR doesn't work either. To be fair, even if the concatenation had worked, it would not have looked great in our code 😅

I changed my approach to actually use the official Reactist add-on that "should" be used to link from a story to another story: https://storybook.js.org/addons/@storybook/addon-links (https://github.com/storybookjs/storybook/issues/8618) and hopefully that will do the trick (for real, this time).

This seems to be working well when testing on the storybook dev version deployed via Chromatic (linked in the builds below) 

https://github.com/Doist/reactist/assets/1509326/018d7b54-7431-4f70-9779-c8fc279bc2c0

...so I have hopes it will work when it gets deployed to https://doist.github.io/reactist/

## PR Checklist

## Versioning

Bumped patch. This is just a doc change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# v21.2.2
+
+-   [Docs] Fix links to Reactist components in storybook stories.
+
 # v21.2.1
 
 -   [Docs] Fix links to Reactist components in storybook stories.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@doist/reactist",
-    "version": "21.2.1",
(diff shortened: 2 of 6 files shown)


engfragui:
I'm happy to confirm that this works as expected after being deployed to production 🎉


https://github.com/Doist/reactist/assets/1509326/35ff4c82-997d-4136-94a3-d57bd5966233

PR #809, opened 2023-12-11
chore: increase show tooltip delay to 1 second

Bloomca:
## Short description

Increase tooltip show delay to 1 second

## PR Checklist

## Versioning

This is a patch change.

### DIFF
--- a/src/tooltip/tooltip.tsx
+++ b/src/tooltip/tooltip.tsx
@@ -77,7 +77,7 @@ type TooltipProps = {
 }
 
 // These are exported to be used in the tests, they are not meant to be exported publicly
-export const SHOW_DELAY = 500
+export const SHOW_DELAY = 1000
 export const HIDE_DELAY = 100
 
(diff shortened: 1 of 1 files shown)


src/tooltip/tooltip.test.tsx
```diff
@@ -247,9 +247,12 @@ describe('Tooltip', () => {
...
         userEvent.tab()
 
-        await waitFor(() => {
-            expect(button).toHaveAccessibleDescription('tooltip content here')
-        })
+        await waitFor(
+            () => {
+                expect(button).toHaveAccessibleDescription('tooltip content here')
+            },
+            { timeout: SHOW_DELAY + 10 },
```
Bloomca, line 254: Default timeout is 1 second, so we basically just need one tick. This is a scalable solution, although each test would take 1 second (instead of 500ms before), so we can advance timers manually. Since we don't have that many tests, I think it is okay to keep it like that.

gnapse (approved):

PR #812, opened 2024-01-12
fix: Explicitly omit `crossOrigin` and `placeholder` props

pawelgrimm:
Relates to:
- https://github.com/creativetimofficial/material-tailwind/issues/427
- https://github.com/creativetimofficial/material-tailwind/issues/528

## Short description

Due to an update to `@types/react`, certain intrinsic HTML element type have been updated to remove unsupported props, namely:
- Removing `crossOrigin` from `<input>` and `<textarea>`
- Removing `placeholder` from `<h1>`, `<h2>`, etc.

This means that if a consumer of this package were to upgrade their `@types/react`, while this package retains the older types, usages of components such as `<TextInput>` and `<Heading>` would be met with type errors complaining about missing those props that have been removed.

The more correct solution here would be to update the `@types/react` package in THIS package, we cannot do so because it causes similar errors stemming from `ariakit`, which is not trivial to upgrade.

As an interim solution, we are explicitly excluding those removed props from affected Reactist component props.

## PR Checklist

## Versioning

This is technically a breaking change, as the interface of many components has changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "@doist/reactist",
-    "version": "22.3.3",
+    "version": "23.0.0",
     "lockfileVersion": 2,
     "requires": true,
     "packages": {
         "": {
             "name": "@doist/reactist",
-            "version": "22.3.3",
+            "version": "23.0.0",
             "hasInstallScript": true,
             "license": "MIT",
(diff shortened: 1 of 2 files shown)


craigcarlyle (approved):
🚀

PR #818, opened 2024-03-08
fix: Reset anchor rect after context menu is closed

gnapse:
## Short description

Last issues needed to be fixed for the Ariakit upgrade.

So far, there's only one.

## PR Checklist

### DIFF
--- a/src/menu/menu.tsx
+++ b/src/menu/menu.tsx
@@ -126,6 +126,11 @@ const ContextMenuTrigger = polymorphicComponent<'div', unknown>(function Context
         [setAnchorRect, menuStore],
     )
 
+    const isOpen = menuStore.useState('open')
+    React.useEffect(() => {
+        if (!isOpen) setAnchorRect(null)
+    }, [isOpen, setAnchorRect])
+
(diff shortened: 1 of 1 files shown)


gnapse:
> the fix was removed in https://github.com/Doist/reactist/pull/778

Ooops, I'm not sure why and do not fully recall now. Possibly a mishap, given how it is needed now.

gnapse:
@rfgamaral you requested information about the fix to the issue of the scheduler and other nested dropdowns not opening properly in the task menu. Here's the explanation:

> I debugged it by focusing in all the ways in which the menu can be dismissed, which can only happen if the `store.hide()` function of the menu’s Ariakit store is called. By placing breakpoints on each of the points in the code that call it, I found the culprit, which turned out to be one I added last week to fix another issue. But that other issue could be fixed differently, thus what I had added last week could be removed, and it fixed this particular issue. More details in [this commit](https://github.com/Doist/reactist/pull/818/commits/0b7ae84c60aa324162caa07cb5de99d77c429e64) from this PR.

rfgamaral:
Thank you, @gnapse, that helped 👍

PR #840, opened 2024-08-29
fix: Disable cross-frame focus lock in `<Modal>`

pawelgrimm:
## Short description

When using `<Modal>` in a Storybook story, controls would stop working. This was determined to be caused by the `<FocusLock>` component, which would prevent interaction with Storybooks preview/add-ons pane once focus was established within the modal (which would happen automatically as `autoFocus` defaults to `true`). To fix this, we're limiting `<FocusLock>`'s focus trap to the document/iframe the Modal is contained within, preventing it from bleeding over into Storybook's preview pane.

## PR Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 Reactist follows [semantic versioning](https://semver.org/) and doesn't introduce breaking changes (API-wise) in minor or patch releases. However, the appearance of a component might change in a minor or patch release so keep an eye on redesigns and make sure your app still looks and feels like you expect it.
 
+# v25.1.1
+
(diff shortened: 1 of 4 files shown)


frankieyan:
I just realized that the focus lock is actually causing problems for us right now 

If you open up a template's task detail in https://todoist.com/templates/todoist-ceo-setup (you'll have to use a browser window without a Todoist logged in session), and once you focus on its elements, you cannot focus on any of the host page's elements anymore, similar to what we're seeing in Storybook.

While we do allow Todoist to be embedded (we have no `X-Frame-Options` header), I'm not aware of any situations where we'd want to take over control from the host page, so I'd imagine this change to be safe to make 👍

PR #842, opened 2024-10-04
refactor: Update form field components to align with the new design spec

rfgamaral:
## Short description

This PR implements a few changes to adapt some of the form fields to the new [form fields spec](https://www.figma.com/design/5gTX7MuUxhCIvL6WK87JVA/24Q3-Foundation?node-id=1947-125676&t=GfBAZcEOXJtltxfU-1). Please refer to `CHANGELOG.md` (or the commit messages) to better understand the changes that were applied.

> [!TIP]
> Given the number of changes to bring the form field components to parity with the new form fields spec, a commit-by-commit review is recommended.

> [!WARNING]
> Don't be scared by the number of changes, most of them come from the last commit, **which is a temporary commit that includes a Reactist build** in this PR branch, so that I can pre-emptively open a PR on `ui-extensions` and `todoist-web` without having it break our CI checks.
>
> Before merging this PR, I'll remove that commit, and update both `ui-extensions` and `todoist-web` with the real package of Reactist v26.

## PR Checklist

## Versioning

Breaking change, some props were removed from the `BaseField` component.


src/base-field/base-field.tsx
```diff
@@ -7,6 +7,9 @@ import { Stack } from '../stack'
 
 import type { WithEnhancedClassName } from '../utils/common-types'
 import { Spinner } from '../spinner'
+import { Column, Columns } from '../columns'
+
+const MAX_LENGTH_THRESHOLD = 10
```
pedroalves0, line 12: I'd add a note in Figma for this implementation detail.

src/base-field/base-field.tsx
```diff
@@ -137,34 +196,72 @@ type BaseFieldProps = WithEnhancedClassName &
...
+                value: event.currentTarget.value,
+                maxLength,
+            })
+
+            setCharacterCount(inputLength.count)
+            setCharacterCountTone(inputLength.tone)
+        },
     }
 
+    React.useEffect(
```
pedroalves0, line 251: I don't think an effect (which delays the update one tick) is needed for this. I think that `useMemo` should be used instead for computing the tone, which would update whenever the character count/props change.

pedroalves0 (approved):
I assume that this PR is not meant to completely align the fields with the new design as there are several things missing or different (e.g. tooltips, invalid error state icon, search clear button).

Not sure if you're continuing this work into Q4, but if you are, I suggest creating stories that replicate 1:1 the Figma designs. Easier to review and to confirm that the client implementation follows the mockups.

rfgamaral:
> I assume that this PR is not meant to completely align the fields with the new design as there are several things missing or different (e.g. tooltips, invalid error state icon, search clear button).

The goal was to simplify the components (remove no longer needed things) and to keep them flexible enough to support the new designs (which they do, unless I missed something) from the consumer side.

> Not sure if you're continuing this work into Q4(...)

AFAIK, this work will be in the back burner for Q4, and might get picked up some time next year.

src/base-field/base-field.tsx
rfgamaral: I've shared the feedback with Julia.

src/base-field/base-field.tsx
rfgamaral: I believe the effect is still required to update the character count when the `value` prop changes, which happens in the following modal when clicking any of the filter example buttons:

image

PR #860, opened 2025-01-16
fix: ensure TextLink styles are applied to nested elements

rmartins90:
## Short description

Hover styles (text decoration and color) were only applied to the container element itself. This caused inconsistent behavior when the text link contained nested elements, as the hover styles wouldn't propagate to the children.

This PR uses `.container:hover > *` selector to target immediate children.

## PR Checklist

### DIFF
--- a/src/text-link/text-link.module.css
+++ b/src/text-link/text-link.module.css
@@ -14,7 +14,17 @@
     cursor: pointer;
 }
 
-.container:hover {
+.container > * {
+    text-decoration: var(--reactist-text-link-idle-decoration);
+    color: var(--reactist-text-link-idle-tint);
+}
+
+/**
+ * We need to target the child elements to ensure that styles are
+ * applied to the text independently of the DOM elements hierarchy.
+ */
+.container:hover,
+.container:hover > * {
     text-decoration: var(--reactist-text-link-hover-decoration);
     color: var(--reactist-text-link-hover-tint);
 }
--- a/src/badge/badge.module.css
+++ b/src/badge/badge.module.css
@@ -29,6 +29,12 @@
     background-color: var(--reactist-badge-fill);
     line-height: calc(var(--reactist-badge-font-size) + 3px);
     white-space: nowrap;
+    text-decoration: none;
+}
+
+.badge:hover {
+    color: var(--reactist-badge-tint);
+    text-decoration: none;
 }
 
 .badge-info {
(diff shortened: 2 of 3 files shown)


src/text-link/text-link.module.css
```diff
@@ -14,7 +14,17 @@
     cursor: pointer;
 }
 
-.container:hover {
+.container > * {
```
rmartins90, line 17: I'm not applying here other rules such as font-weight so they are preserved for nested elements

rmartins90:
> Looks good Rui! Just one comment, with Reactist updates, we tend to keep the versioning changes separate from the product updates.

That makes sense! We should then change the PR template, as it misleads. Do you agree? (see below)

> - Described changes in CHANGELOG.md
> - Bumped version in package.json and package-lock.json (npm --no-git-tag-version version <major|minor|patch>) [ref](https://docs.npmjs.com/cli/v6/commands/npm-version)

There are other examples where we change the version and CHANGELOG in the PR - https://github.com/Doist/reactist/commit/d88972ba169fefe7a40dd335b27de5fa80a47e11

Either way, this process should be automated soon 🙂

rmartins90:
Removed versioning from this PR @nats12. I'm merging this now.

nats12:
> That makes sense! We should then change the PR template, as it misleads. Do you agree? (see below)
> There are other examples where we change the version and CHANGELOG in the PR - https://github.com/Doist/reactist/commit/d88972ba169fefe7a40dd335b27de5fa80a47e11

Good point @rmartins90. @gnapse maybe can shed light on this: we are sometimes bumping versions in the same PR as the change, other times in separate PRs. Which process should we follow? ☺️

gnapse:
We've often bumped the version in the same PR as we make a product change ([ref](https://github.com/Doist/reactist/commit/d88972ba169fefe7a40dd335b27de5fa80a47e11), [ref](https://github.com/Doist/reactist/commit/5c79f80309d20a0b8f8ab5f6b03e2f41d16c1600), [ref](https://github.com/Doist/reactist/commit/75441cdb1ef09770f676c8f81597773c6265ce6e)). But not always.

I don't think there's a hard rule on this. Personally, I tend to bump the version with the change, unless I have two or more changes that I know I can release together. In which case I hold off on the version bump, merge two or more PRs in a row, then do a PR with the version bump.

Regarding the PR request template, I think we can remove the "versioning" section regardless, because we already know from the PR title prefix if it's a patch release (e.g. `fix:…`), a minor release (e.g. `feat:…`) or a major release (breaking change).

rmartins90:
Meanwhile, I've created a PR to automate this process https://github.com/Doist/reactist/pull/863

PR #879, opened 2025-01-22
ci: fix dependabot config syntax

rmartins90:
Closes #...

## Short description

Fix syntax on dependabot config. Ref: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference

## PR Checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,6 +3,7 @@ version: 2
 updates:
     - package-ecosystem: npm
       directory: /
-      interval: daily
+      schedule:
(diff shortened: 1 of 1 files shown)


henningmu:
@rmartins90 should we switch to Renovate over Dependabot as we do internally? 🤔

rmartins90:
I thought about that but didn't want to change it at this point, as Dependabot was already integrated with some pending PRs. Though, I don't have a strong preference here. What are your thoughts on that?

PR #883, opened 2025-01-28
feat(TextField): add support for customizable character count position

rmartins90:
## Short description

This PR introduces a new feature to customize the character count position in text fields. Major changes include:
- Added a new `characterCountPosition` prop to `BaseField` component with three options (two of them new):
  - `'below'` (default) - shows the counter below the field
  - `'inline'` - displays the counter inline with the field content
  - `'hidden'` - hides the character counter completely
- Updated `TextField` component to support the new character count positioning feature
- Added new stories and examples in the documentation
- Added test coverage for the new feature

CleanShot 2025-01-28 at 22 04 10@2x

### References
- Twist thread: https://twist.com/a/1585/ch/765851/t/6664583/

## PR Checklist

### DIFF
--- a/src/base-field/base-field.tsx
+++ b/src/base-field/base-field.tsx
@@ -94,6 +94,7 @@ type ChildrenRenderProps = {
     'aria-describedby'?: string
     'aria-invalid'?: true
     onChange?: React.ChangeEventHandler<HTMLInputElement | HTMLTextAreaElement | HTMLSelectElement>
+    characterCountElement?: React.ReactNode | null
 }
 
 type HtmlInputProps<T extends HTMLElement> = React.DetailedHTMLProps<
@@ -118,7 +119,7 @@ type BaseFieldVariantProps = {
     variant?: BaseFieldVariant
 }
 
-type BaseFieldProps = WithEnhancedClassName &
+export type BaseFieldProps = WithEnhancedClassName &
     Pick<HtmlInputProps<HTMLInputElement>, 'id' | 'hidden' | 'maxLength' | 'aria-describedby'> & {
         /**
          * The main label for this field element.
@@ -202,6 +203,14 @@ type BaseFieldProps = WithEnhancedClassName &
          * the public props of such components.
          */
         children: (props: ChildrenRenderProps) => React.ReactNode
+
+        /**
+         * The position of the character count element.
+         * It can be shown below the field or inline with the field.
+         *
+         * @default 'below'
+         */
+        characterCountPosition?: 'below' | 'inline' | 'hidden'
     }
 
 type FieldComponentProps<T extends HTMLElement> = Omit<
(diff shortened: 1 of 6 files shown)


src/text-field/text-field.tsx
```diff
@@ -97,13 +100,14 @@ const TextField = React.forwardRef<HTMLInputElement, TextFieldProps>(function Te
...
-                    {endSlot ? (
+                    {endSlot || characterCountElement ? (
                         <Box
                             className={styles.slot}
                             display="flex"
                             marginRight={variant === 'bordered' ? '-xsmall' : 'xsmall'}
                             marginLeft={variant === 'bordered' ? 'xsmall' : '-xsmall'}
                         >
+                            {characterCountElement}
                             {endSlot}
```
engfragui, line 111: 💭 Playing with storybook, it looks slightly unpolished/not aligned when I toggle `endSlot` and I also choose the inline position for the character count:

Screenshot 2025-01-30 at 09 23 54
 
I think this could be solved by adding vertical alignment to this <Box> with `alignItems="center"` (or something like that)

src/base-field/base-field.tsx
```diff
@@ -202,6 +203,14 @@ type BaseFieldProps = WithEnhancedClassName &
...
          */
         children: (props: ChildrenRenderProps) => React.ReactNode
+
+        /**
+         * The position of the character count element.
+         * It can be shown below the field or inline with the field.
+         *
+         * @default 'below'
+         */
+        characterCountPosition?: 'below' | 'inline' | 'hidden'
```
engfragui, line 213: 💭 Mmm `BaseField` is used by `SelectField` as well and I'm not sure if this new `characterCountPosition` prop would make sense for that specific field 🤔

src/text-field/text-field.tsx
rmartins90: Great catch @engfragui. Just pushed fix for this.

CleanShot 2025-01-30 at 22 51 55@2x

src/base-field/base-field.tsx
rmartins90: That's true, but it's also used by `TextArea`, and it makes sense there. Also, `maxLength` doesn't make sense in `SelectField`. So what I did was omitting `maxLength` and `characterCountPosition` from SelectField: https://github.com/Doist/reactist/pull/883/commits/6f8cc671a0332374e6d5541bbfc98c1b36d0c0c5

What do you think?

engfragui: That sounds good 👍

engfragui (approved):

engfragui:
@rmartins90 It seems like `characterCountPosition` doesn't work anymore in Reactist?


https://github.com/user-attachments/assets/09626e60-1e16-4aa7-89e6-7442edc42610

rmartins90:
I agree that for now we should disable it for TextArea @engfragui.

PR #897, opened 2025-02-18
feat(TextLink): Add color and underline styling options

rmartins90:
## Short description

- Introduce color prop with 'default' and 'inherit' variants
- Add underline prop to control text-decoration
- Extend stories and tests to cover new styling options

CleanShot 2025-02-18 at 20 28 26@2x

## PR Checklist

### DIFF
--- a/src/text-link/text-link.stories.mdx
+++ b/src/text-link/text-link.stories.mdx
@@ -47,13 +47,7 @@ A component used to create text-based hyperlinks.
 When using nested elements inside a `TextLink`, hover styles are properly applied to all children.
 
 <Canvas>
-    <Story
-        name="Nested elements"
-        parameters={{
-            docs: { source: { type: 'code' } },
-            chromatic: { disableSnapshot: false },
-        }}
-    >
+    <Story name="Nested elements">
         <Stack space="medium" align="start">
             <TextLink href="https://www.doist.com/">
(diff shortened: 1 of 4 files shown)


src/text-link/text-link.stories.mdx
```diff
@@ -74,3 +73,27 @@ When using nested elements inside a `TextLink`, hover styles are properly applie
         </Stack>
     </Story>
 </Canvas>
+
+## Style
```
Bloomca, line 77: I think these need to be changed to actual props. Currently it is defined separately, not in a table: [ref](https://638e186cd0b36e1276f510de-uamqrgqgzc.chromatic.com/?path=/docs/design-system-textlink--main-demo). E.g. look at the stack: [ref](https://638e186cd0b36e1276f510de-uamqrgqgzc.chromatic.com/?path=/docs/design-system-stack--interactive-props-story)

Bloomca (approved):
Looks good, although I think in the story is not defined in the best way, we should list props in a table. After that all should be good 👍

src/text-link/text-link.stories.mdx
rmartins90: that's right, props are not being infered automatically by Storybook. Probably because of polymorphism method, that is deprecated and I should remove soon. When I refactor that I will address that 🙂

PR #899, opened 2025-03-12
feat: Allow positional placement of `endSlot` in `TextField`

scottlovegrove:
## Short description

This change allows us to use the `bordered` variant of `TextField` and have the endSlot appear centrally, rather than inline with the input at the bottom. 

Before: 
image

After (with `endSlotPosition` set to `fullHeight`:
image

I know Design would like this to be the default behaviour, but given I only need it for one area in the app (new onboarding), I was reluctant to make my changes the default behaviour, so I haven't. This is opt-in, rather than opt-out.

## PR Checklist


package.json
engfragui, line 1: Intentional? 🤔

src/text-field/text-field.tsx
```diff
@@ -9,7 +9,7 @@ type TextFieldType = 'email' | 'search' | 'tel' | 'text' | 'url'
 
 interface TextFieldProps
     extends Omit<FieldComponentProps<HTMLInputElement>, 'type'>,
-        BaseFieldVariantProps,
+        Omit<BaseFieldVariantProps, 'supportsStartAndEndSlots'>,
```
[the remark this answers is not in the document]
engfragui: Oh, I think I got it. `supportsStartAndEndSlots` is actually included in `FieldComponentProps` and not in `BaseFieldVariantProps` so it should be omitted from there instead, am I right?

src/text-field/text-field.tsx
```diff
@@ -9,7 +9,7 @@ type TextFieldType = 'email' | 'search' | 'tel' | 'text' | 'url'
 
 interface TextFieldProps
     extends Omit<FieldComponentProps<HTMLInputElement>, 'type'>,
```
engfragui, line 11:
```suggestion
    extends Omit<FieldComponentProps<HTMLInputElement>, 'type' | 'supportsStartAndEndSlots'>,
```

src/base-field/base-field.tsx
```diff
@@ -213,7 +213,23 @@ export type BaseFieldProps = WithEnhancedClassName &
...
+        | {
+              supportsStartAndEndSlots?: false
+              endSlot?: never
+              endSlotPosition?: never
+          }
+        | {
+              supportsStartAndEndSlots: true
+              endSlot?: React.ReactElement | string | number
+              /**
+               * This is primarily for `bordered` variants of TextField. When set to `bottom` (the default),
```
engfragui, line 226: 💭 Looking at the code, it seems like this is "solely" more than "primarily" for `bordered` variants of TextField (we'd need to update this comment to make it more clear).

scottlovegrove: Fair.

src/text-field/text-field.tsx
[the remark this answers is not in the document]
scottlovegrove: I've tried all and I can't make the damn thing disappear from storybook 😄

src/base-field/base-field.tsx
engfragui: What I meant by `endSlotPosition` being "solely" for `bordered` variants of TextField is that if I have a `default` version but I still pass `endSlotPosition: fullHeight`, that will make so that what I passed as `endSlot` doesn't get honored:

Screenshot 2025-03-12 at 15 04 17

Screenshot 2025-03-12 at 15 04 25

I feel like we should just treat it as `endSlotPosition: bottom`.

src/text-field/text-field.tsx
scottlovegrove: For whatever reason, this doesn't remove it from Storybook.

engfragui: Mmm it does remove it for me (after restarting Storybook).

scottlovegrove: Which it does. Stupid Storybook.

src/base-field/base-field.tsx
scottlovegrove: So update the types again so that if it's not `bordered` then `endSlotPosition` can't be set?

engfragui:
`variant: bordered` + some `endSlot` passed in -> I would expect the `endSlot` to show up regardless of the value of `endSlotPosition` (even if it's undefined), right? 🤔

Screenshot 2025-03-12 at 15 59 13

engfragui (approved):

engfragui:
Storybook issue fixed in https://github.com/Doist/reactist/pull/902

PR #903, opened 2025-04-16
feat(Tabs)!: Add support for additional customization settings

rfgamaral:
## Short description

This PR adds support for the following customization settings for the `Tabs` component:

- CSS variables for hover and disabled states for the `Tab` component
    - `--reactist-tab-themed-hover` / `--reactist-tab-neutral-hover`
    - `--reactist-tab-themed-disabled` / `--reactist-tab-themed-disabled`
    - A tab can be individually disabled with the `disabled` prop
- CSS variable for a box shadow for the selected tab
    - `--reactist-tab-neutral-shadow` / `--reactist-tab-themed-shadow`
- Full width for the `TabList` component
    - The `width` prop can be `full` or `maxContent` (default)
- Tab alignment for the `TabList` component
    - The `align` prop can be `start` (default), `center`, or `end`
- CSS variable for a transition for the selected tab
    - `--reactist-tab-selected-transition`
- `exceptionallySetClassName` prop added to the `TabList` component
    - Allows to override the `--reactist-tab-border-radius` variable used in the "track" element

All these changes are necessary to help customize the `Tabs` component to help with the upcoming style updates, and they come with sensible defaults that match the current style and behaviour. This avoids breaking changes, and helps with a smoother transition. I've also added a few example stories to demonstrate some of the most relevant new settings.

## PR Checklist

### DIFF
--- a/src/tabs/tabs.tsx
+++ b/src/tabs/tabs.tsx
@@ -149,12 +149,27 @@ type TabListProps = (
      * Controls the spacing between tabs
      */
     space?: Space
+
+    /**
+     * The width of the tab list.
+     *
+     * - `'maxContent'`: Each tab will be as wide as its content.
+     * - `'full'`: Each tab will be as wide as the tab list.
+     *
+     * @default 'maxContent'
+     */
+    width?: 'maxContent' | 'full'
 }
 
 /**
  * A component used to group `<Tab>` elements together.
  */
-function TabList({ children, space, ...props }: TabListProps): React.ReactElement | null {
+function TabList({
+    children,
+    space,
+    width = 'maxContent',
+    ...props
+}: TabListProps): React.ReactElement | null {
     const tabContextValue = React.useContext(TabsContext)
 
     if (!tabContextValue) {
@@ -169,11 +184,18 @@ function TabList({ children, space, ...props }: TabListProps): React.ReactElemen
         <div>
             <BaseTabList
                 store={tabStore}
-                render={<Box position="relative" width="maxContent" />}
+                render={<Box position="relative" width={width} />}
                 {...props}
             >
                 <Box className={[styles.track, styles[`track-${variant}`]]} />
-                <Inline space={space}>{children}</Inline>
+                <Inline
+                    space={space}
+                    exceptionallySetClassName={classNames(
+                        width === 'full' ? styles.fullTabList : null,
+                    )}
+                >
+                    {children}
+                </Inline>
             </BaseTabList>
         </div>
     )
--- a/src/tabs/tabs.tsx
+++ b/src/tabs/tabs.tsx
@@ -168,7 +168,7 @@ type TabListProps = (
      * @default 'start'
      */
     align?: ResponsiveProp<'start' | 'center' | 'end'>
-}
+} & ObfuscatedClassName
 
 /**
  * A component used to group `<Tab>` elements together.
@@ -178,6 +178,7 @@ function TabList({
     space,
     width = 'maxContent',
     align = 'start',
+    exceptionallySetClassName,
     ...props
 }: TabListProps): React.ReactElement | null {
     const tabContextValue = React.useContext(TabsContext)
@@ -233,7 +234,9 @@ function TabList({
         >
             <BaseTabList
                 store={tabStore}
-                render={<Box position="relative" width={width} />}
+                render={
+                    <Box position="relative" width={width} className={exceptionallySetClassName} />
+                }
                 ref={tabListRef}
                 {...props}
             >
--- a/src/tabs/tabs.tsx
+++ b/src/tabs/tabs.tsx
@@ -86,13 +86,18 @@ interface TabProps
      * The tab's identifier. This must match its corresponding `<TabPanel>`'s id
      */
     id: string
+
+    /**
+     * Defines wether or not the tab is disabled.
+     */
+    disabled?: boolean
 }
 
 /**
  * Represents the individual tab elements within the group. Each `<Tab>` must have a corresponding `<TabPanel>` component.
  */
 const Tab = React.forwardRef<HTMLButtonElement, TabProps>(function Tab(
-    { children, id, exceptionallySetClassName, render, onClick },
+    { children, id, disabled, exceptionallySetClassName, render, onClick },
     ref,
 ): React.ReactElement | null {
     const tabContextValue = React.useContext(TabsContext)
@@ -105,6 +110,7 @@ const Tab = React.forwardRef<HTMLButtonElement, TabProps>(function Tab(
         <BaseTab
             id={id}
             ref={ref}
+            disabled={disabled}
             store={tabStore}
             render={render}
             className={className}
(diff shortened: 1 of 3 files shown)


src/tabs/tabs.tsx
```diff
@@ -143,33 +151,112 @@ type TabListProps = (
...
+}: TabListProps): React.ReactElement | null {
     const tabContextValue = React.useContext(TabsContext)
 
+    const [selectedTabElement, setSelectedTabElement] = React.useState<HTMLElement | null>(null)
+    const [selectedTabStyle, setSelectedTabStyle] = React.useState<React.CSSProperties>({})
+    const tabListRef = React.useRef<HTMLDivElement>(null)
+
+    const selectedId = tabContextValue?.tabStore.useState('selectedId')
+
+    React.useLayoutEffect(() => {
```
frankieyan, line 192: image

I noticed an issue with the tab's width and position when the viewport is resized due to them being calculated on render.

src/tabs/tabs.tsx
```diff
@@ -143,33 +151,112 @@ type TabListProps = (
...
-        // The extra <div> prevents <Inline>'s negative margins from collapsing when used in a flex container
-        // which will render the track with the wrong height
-        <div>
+        // This extra <Box> not only provides alignment for the tabs, but also prevents <Inline>'s
+        // negative margins from collapsing when used in a flex container which will render the
+        // track with the wrong height
+        <Box
+            display="flex"
+            justifyContent={
+                width === 'full'
```
frankieyan, line 226: Might be easier to see the relationship between align's and justifyContent's values if we constructed a hash like this:

```ts
const justifyContentAlignMap = { 
    start: 'flexStart',
    end: 'flexEnd',
    center: 'center'
}

justifyContent={width === 'full' ? 'center' : justfiyContentAlignMap[align]}
```

frankieyan (approved):
Looks good 👍 comments are not blocking but I'd like to see the tab's styles not break when resized. Perhaps there is a possible CSS-only solution?

src/tabs/tabs.tsx
rfgamaral: Great find, I'll fix this.

src/tabs/tabs.tsx
rfgamaral: Sounds good, I'll make the change.

rfgamaral:
> Is the border radius point meant to be a sub-bullet item here or are they separate points? I'm confused because we should be able to just override the `--reactist-tab-border-radius` without the class name override 😁

The `--reactist-tab-border-radius` variable is used in two places within the Tabs component, one of those is the [`.track` class](https://github.com/Doist/reactist/blob/b539bc974569a646964f05d3b360ed95c0bc4cee/src/tabs/tabs.module.css#L47), and that `.track` class is used [here](https://github.com/Doist/reactist/blob/b539bc974569a646964f05d3b360ed95c0bc4cee/src/tabs/tabs.tsx#L243). Without the `exceptionallySetClassName` you can't override that variable unless you create a wrapper container around the `TabsList` component (or the whole `Tabs` component), and you would be doing that just to override this variable, which doesn't make a lot of sense to me.

So, no, they are not separate points. The `exceptionallySetClassName` was added so that we are able to override `--reactist-tab-border-radius` in all places that this variable is used. If you have a suggestion for a different approach, I'm open to ideas :)

frankieyan:
Ah ok, I think I get it - if we want to be able to apply different border radius values to the track and to the selected tab, perhaps separating it into two variables could be a solution, i.e. `--reactist-tab-track-border-radius` and `--reactist-tab-selected-border-radius`. It'd make it more explicit that this is a customization we support than the `exceptionallySetClassName` path.

rfgamaral:
> if we want to be able to apply different border radius values to the track and to the selected tab, perhaps separating it into two variables could be a solution

The new design uses the same radius for both, but I'll split them for extended customizability :)

> It'd make it more explicit that this is a customization we support than the `exceptionallySetClassName` path.

It doesn't change the fact that we still need this, though.

frankieyan:
Okay sorry I am finally starting to get it 😅 This problem would exist for all of the other exposed custom properties and not just for the border radius variables though, correct? 

I'm not against having the `exceptionallySetClassName` prop available but I want to fully understand the use case. 

Is the border radius something we want to customize per instance, or app-wide? If it's the latter, we could always configure the new values based on the global class added in https://github.com/Doist/todoist-web/pull/13368, and if it's the former, and I'd be curious why we'd do this, then having the class name makes more sense.

src/tabs/tabs.tsx
```diff
@@ -189,23 +189,32 @@ function TabList({
...
+        }
 
-        const selectedTab = Array.from(tabs).find(
-            (tab) => tab.getAttribute('id') === selectedId,
-        ) as HTMLElement | undefined
+        updateSelectedTabStyle()
 
-        if (selectedTab) {
-            setSelectedTabElement(selectedTab)
+        window.addEventListener('resize', updateSelectedTabStyle)
```
frankieyan, line 214: Should this be debounced?

rfgamaral: I'm guessing we don't want that:

- If the user is in the process of resizing, with a debounce function, they will not immediately see the tab size being resized, which will look weird (like your screenshot)
- We have a couple more instances of similar code to this in Todoist, none of which uses a debounce function.

We can add one if it ends up being an issue.

rfgamaral:
> This problem would exist for all of the other exposed custom properties and not just for the border radius variables though, correct?

Not necessarily.

All the other properties are used under elements that either expose a `className` or a `exceptionallySetClassName` prop. The `TabList` component didn't expose either, that's why we need it, because the track element inside that component uses a variable that I need to customize.

> Is the border radius something we want to customize per instance, or app-wide?

It will be app-wide.

> If it's the latter, we could always configure the new values based on the global class added in https://github.com/Doist/todoist-web/pull/13368

Hum, this is a good point. I was approaching this problem differently, but now that you mention this, I see how the same result could be achieved without the need for `exceptionallySetClassName`.

Since I've already published a new Reactist release, I'll test both approaches and see how they look. If I end up not using `exceptionallySetClassName`, I can remove it. Or we can just keep it around "just in case" since it's in the name that it should be used "exceptionally" 😅 What do you think?

src/tabs/tabs.tsx
frankieyan: Yeah I think this will cause resizing to be (even more) janky because reading `offsetLeft` and `offsetWidth` will cause a [reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a), and the resize event gets called very often. Limiting the redraws during resize should help mitigate this and reduce the work that's being done.

Even on the storybook page which is fairly lightweight, we can alread see its impact during a resize: 
image

frankieyan:
> Or we can just keep it around "just in case" since it's in the name that it should be used "exceptionally" 😅 What do you think?

No harm in leaving it in place I think 👍

PR #909, opened 2025-04-18
fix(Tabs): Add resize observer for dynamic tab selection adjustments

rfgamaral:
## Short description

This PR replaces our viewport‐only resize listener with a `ResizeObserver` on the tabs container itself, ensuring the selected tab is always correctly positioned whenever its parent’s width changes. In particular, it fixes an edge case where removing an inline element (for example, a status badge or "unread" count) from a tab causes the container to shrink without firing a window `resize` event, leaving the selected tab indicator out of sync.

**What’s changed:**

- **Replace** the `window.addEventListener('resize', …)` logic for updating the selected tab with a `ResizeObserver` attached to the tabs list container.
- **On resize** of the container or **on manual tab change**, run the same recalculation & repositioning logic for the active tab.

This makes the tabs component resilient to any changes in its own dimensions—whether coming from viewport resizes, dynamic inline element changes, font changes, or other content shifts—so the active tab indicator stays perfectly in sync.

---

_@frankieyan Assigned you directly since you reviewed the original PR where the viewport resize was introduced, and I want your thoughts on this alternative solution that should be more robust._

### DIFF
--- a/src/tabs/tabs.tsx
+++ b/src/tabs/tabs.tsx
@@ -223,21 +223,35 @@ function TabList({
 
     React.useEffect(
         function observeTabListWidthChange() {
-            const tabList = new ResizeObserver(([entry]) => {
+            let animationFrameId: number | null = null
+
+            const tabListObserver = new ResizeObserver(([entry]) => {
                 const width = entry?.contentRect.width
 
                 if (width && tabListPrevWidthRef.current !== width) {
                     tabListPrevWidthRef.current = width
-                    updateSelectedTabPosition()
+
+                    if (animationFrameId !== null) {
+                        cancelAnimationFrame(animationFrameId)
+                    }
+
+                    animationFrameId = requestAnimationFrame(() => {
+                        updateSelectedTabPosition()
+                        animationFrameId = null
+                    })
                 }
             })
 
             if (tabListRef.current) {
-                tabList.observe(tabListRef.current)
+                tabListObserver.observe(tabListRef.current)
             }
 
             return function cleanupResizeObserver() {
-                tabList.disconnect()
+                if (animationFrameId) {
+                    cancelAnimationFrame(animationFrameId)
+                }
+
+                tabListObserver.disconnect()
             }
         },
         [updateSelectedTabPosition],
--- a/src/tabs/tabs.tsx
+++ b/src/tabs/tabs.tsx
@@ -180,16 +180,18 @@ function TabList({
     exceptionallySetClassName,
     ...props
 }: TabListProps): React.ReactElement | null {
+    const tabListRef = React.useRef<HTMLDivElement | null>(null)
+    const tabListPrevWidthRef = React.useRef(0)
+
     const tabContextValue = React.useContext(TabsContext)
 
     const [selectedTabElement, setSelectedTabElement] = React.useState<HTMLElement | null>(null)
     const [selectedTabStyle, setSelectedTabStyle] = React.useState<React.CSSProperties>({})
-    const tabListRef = React.useRef<HTMLDivElement>(null)
 
     const selectedId = tabContextValue?.tabStore.useState('selectedId')
 
-    React.useLayoutEffect(() => {
-        function updateSelectedTabStyle() {
+    const updateSelectedTabPosition = React.useCallback(
+        function updateSelectedTabPositionCallback() {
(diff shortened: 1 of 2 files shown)


src/tabs/tabs.tsx
```diff
@@ -207,16 +209,39 @@ function TabList({
...
                 })
             }
-        }
+        },
+        [selectedId],
+    )
+
+    React.useEffect(
+        function updateSelectedTabPositionOnTabChange() {
+            updateSelectedTabPosition()
```
frankieyan, line 218: It'd be better to have `updateSelectedTabPosition` accept `selectedId` as a parameter to make the dependency here explicit, and `updateSelectedTabPosition` itself can then have no dependency.

src/tabs/tabs.tsx
```diff
@@ -207,16 +209,39 @@ function TabList({
...
+        },
+        // `selectedId` is a dependency to ensure the effect runs when the selected tab changes
+        [selectedId, updateSelectedTabPosition],
+    )
 
-        updateSelectedTabStyle()
+    React.useEffect(
+        function observeTabListWidthChange() {
+            const tabList = new ResizeObserver(([entry]) => {
```
frankieyan, line 226: I think it's still worth having this be throttled (rather than debounced if we want to see transitional frames), for somewhere between 64ms (so it's capped at ~30fps, to 100ms for 10fps).

Also slight nit:

```suggestion
            const tabListObserver = new ResizeObserver(([entry]) => {
```

frankieyan (approved):

src/tabs/tabs.tsx
rfgamaral: > and `updateSelectedTabPosition` itself can then have no dependency.

If `updateSelectedTabPosition` accepts `selectedId` as parameter, every call to `updateSelectedTabPosition` needs to be updateSelectedTabPosition(selectedId)`. This will force every hook making this call to need the `selectedId` dependency, including `observeTabListWidthChange`, which we only want to be called once to initialize the observer.

src/tabs/tabs.tsx
rfgamaral: The problem with "throttling" is that we don't have lodash on Reactist, and I'd have to implement a custom function just for this (or copy some battle-tested throttle function from somewhere). Alternative, use lodash and add it as a peer dependency to Reactist. Any other options? How do you think we should handle this?

rfgamaral: How about using [`requestAnimationFrame`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame)? Feels like the most appropriate solution. I've pushed a commit with that change, let me know what you think.

rfgamaral:
@frankieyan I'm going to merge as-is, since you approved despite your suggestions (I did act on one of them, which I think is the solution we need). If you still think this should be improved somehow, feel free to open a PR yourself and assign me directly.

frankieyan:
> I'm going to merge as-is, since you approved despite your suggestions (I did act on one of them, which I think is the solution we need). If you still think this should be improved somehow, feel free to open a PR yourself and assign me directly.

@rfgamaral could you please do a quick benchmark for resizing the window before and after integrating this into the app? It'd be intersting to see what kind of impact it has 🙏

src/tabs/tabs.tsx
frankieyan: > This will force every hook making this call to need the selectedId dependency, including observeTabListWidthChange, which we only want to be called once to initialize the observer.

This is both true and false 🙂 While `observeTabListWidthChange`'s only explicit dependency is `updateSelectedTabPosition`, `updateSelectedTabPosition` depends on `selectedId` so a new reference would be created when that changes, making `observeTabListWidthChange` also rerun when `selectedId` changes. Extracting that into a param and making that an explicit dependency for consumers would make it clearer that this is happening.

rfgamaral:
@frankieyan Didn't notice anything out of the ordinary.

Please note that resizing a window is probably not something that users are constantly doing, and the resize observer should be very efficient. We use it across the app in more prominent features, and so far it's working great. I don't think there's much for us to worry about here.

frankieyan:
Ok, as long as we're not seeing excessive slow downs when testing (hopefully with the cpu throttled), I'm OK with leaving the observer unthrottled for now 👍

PR #912, opened 2025-04-29
fix: fix text area auto-expand when controlled

### LINKED ISSUE #911: TextArea with autoExpand does not update the height when the value is programmatically changed
### Current behavior

When the value is changed programmatically from a large text that expands the text area, to a short or empty text, the textarea remains expanded as if the long text is still there.

Press Enter on a textarea like the one below

```jsx
export function View() {
    const [value, setValue] = React.useState('')
    return (
        <TextArea
            label="Type something"
            autoExpand
            value={value}
            onChange={(event) => setValue(event.target.value)}
            onKeyDown={(event) => {
                if (event.key === 'Enter') {
                    event.preventDefault()
                    setValue('') // Clear the input programmatically
                }
            }}
        />
    )
}
```

### Expected behavior

The textarea, when turned empty, should shrink its height to how it normally appears when empty.

### Possible solutions

🤷‍♂️

### Environment

-   `@doist/reactist` version: latest as of today
-   `react` version: 18
-   Browser: any browser

gnapse:
## Short description

Fixes #911 

The main issue is that our current auto-expand implementation is more suitable for when the component is uncontrolled. When the component is controlled, it is better to use the `value` changes as the signal to run the effect.

This PR now changes the auto-expand to set up two different `useEffect` calls:

1. one that runs only when the component is uncontrolled (it uses the `oninput` event to update the auto-expand size)
2. one that runs only when the component is controlled (it has the `value` as dependency in the effect, to update the auto-expand size)

## PR Checklist

### DIFF
--- a/src/text-area/text-area.stories.mdx
+++ b/src/text-area/text-area.stories.mdx
@@ -185,12 +185,14 @@ Note that these variables are shared with other components such as `Textfield`, 
 export function AutoExpandStory(props) {
     const [value, setValue] = React.useState('')
     return (
-        <Box maxWidth="small">
+        <Stack space="large" dividers="secondary" maxWidth="medium">
             <TextArea
-                label="Type something"
+                label="Text area with auto-expand"
+                auxiliaryLabel="(controlled)"
                 autoExpand
                 value={value}
                 onChange={(event) => setValue(event.target.value)}
+                message="Write as much or as little as you want. The input area will auto-expand to fit what you've typed."
                 rows={1}
                 onKeyDown={(event) => {
                     if (event.key === 'Enter') {
@@ -199,7 +201,12 @@ export function AutoExpandStory(props) {
                     }
                 }}
             />
-        </Box>
+            <Text size="caption" tone="secondary">
+                If you press Enter, the input will be cleared. This allows you to test that
+                auto-expand works when the input is cleared programmatically, shrinking the textarea
+                to the new expected height.
+            </Text>
+        </Stack>
     )
 }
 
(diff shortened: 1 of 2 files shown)


src/text-area/text-area.stories.mdx
```diff
@@ -182,20 +183,29 @@ Note that these variables are shared with other components such as `Textfield`,
...
+                        event.preventDefault()
+                        setValue('') // Clear the input programmatically
+                    }
+                }}
             />
-            <TextArea
-                {...props}
-                label="What do you want to accomplish?"
-                message="This one will not auto-expand."
-                autoExpand={false}
```
engfragui, line 197: 💭 Was it intentional to remove the "What do you want to accomplish?" textarea that had `autoExpand={false}`? I'm ok without it (as anyway there are plenty of other stories with textareas with `autoExpand={false}`

src/text-area/text-area.stories.mdx
```diff
@@ -182,20 +183,29 @@ Note that these variables are shared with other components such as `Textfield`,
...
+                value={value}
+                onChange={(event) => setValue(event.target.value)}
+                message="Write as much or as little as you want. The input area will auto-expand to fit what you've typed."
+                rows={1}
+                onKeyDown={(event) => {
+                    if (event.key === 'Enter') {
+                        event.preventDefault()
+                        setValue('') // Clear the input programmatically
+                    }
+                }}
```
engfragui, line 202: 💭 I've noticed that this textarea doesn't have a max-width anymore in storybook, in the sense that it extends further than before. Not sure if it means anything 😅

<table>
<tbody>
<tr><td>

Screenshot 2025-04-30 at 11 18 47


</td><td>


Screenshot 2025-04-30 at 11 18 38




</td></tr>
</table>

engfragui: Mmm it seems to be because of the different `max-width` value compared to before (prod):

<table>
<tbody>
<tr><td>

Screenshot 2025-04-30 at 11 23 52

</td><td>


Screenshot 2025-04-30 at 11 23 17

</td></tr>
</table>

engfragui: Oh I guess it's because we're not passing `{...props}` anymore to the `textarea` in this story. Was that intentional?

src/text-area/text-area.tsx
```diff
@@ -139,5 +124,60 @@ const TextArea = React.forwardRef<HTMLTextAreaElement, TextAreaProps>(function T
...
+    value: string | undefined
+    autoExpand: boolean
+    containerRef: React.RefObject<HTMLDivElement>
+    internalRef: React.RefObject<HTMLTextAreaElement>
+}) {
+    const isControlled = value !== undefined
+
+    React.useEffect(
+        function setupAutoExpandWhenUncontrolled() {
+            const textAreaElement = internalRef.current
```
engfragui, line 142: 💭 I see that in `setupAutoExpandWhenControlled` you check for `!isControlled -> return`.
Would it make sense to check for `isControl -> return` in this case?

engfragui (approved):
Makes sense to be overall! I left one blocking comment (need to pass `{...props}` to the controlled textarea in storybook).

src/text-area/text-area.stories.mdx
gnapse: My mistake! Must have removed the `{...props}` inadvertently.

PR #914, opened 2025-05-01
feat: Expose closeToast in useToasts hook action object

nats12:
## Short description

Adds a new prop to the `useToasts` hook: `showStickyToast`. This works exactly as seen in Storybook when checking the sticky toast checkbox [here](https://doist.github.io/reactist/?path=/docs/design-system-toast--notification-toasts-story). Except that state is managed locally in the story. 

We're exposing via the hook to give consumers a chance to choose whether the action calls `onDismiss` (which aside from closing the toast, would also call any logic consumers are passing into `onDismiss`, which may be undesirable).

## PR Checklist

### DIFF
--- a/src/toast/toast.stories.tsx
+++ b/src/toast/toast.stories.tsx
@@ -88,10 +88,26 @@ export function NotificationToastsStory() {
                         >
                             Show toast with action
                         </Button>
+                        <Button
+                            variant="primary"
+                            onClick={() => {
+                                const actionLabel = getRandom(actions)
+                                showToast({
+                                    message: `${count.current++}: ${getRandom(message)}`,
+                                    action: {
+                                        label: actionLabel,
+                                        onClick: storybookAction(actionLabel),
+                                    },
+                                    showStickyToast: true,
+                                })
+                            }}
+                        >
+                            Show sticky toast
+                        </Button>
                     </Inline>
                 </Box>
                 <SwitchField
-                    label="Show sticky toast?"
+                    label="Show sticky toast without hook?"
                     checked={showSticky}
                     onChange={(event) => setShowSticky(event.target.checked)}
                     message={
--- a/src/toast/toast.stories.tsx
+++ b/src/toast/toast.stories.tsx
@@ -97,17 +97,17 @@ export function NotificationToastsStory() {
                                     action: {
                                         label: actionLabel,
                                         onClick: storybookAction(actionLabel),
+                                        closeToast: false,
                                     },
-                                    showStickyToast: true,
                                 })
                             }}
                         >
-                            Show sticky toast
+                            Show toast without closing on action
                         </Button>
                     </Inline>
                 </Box>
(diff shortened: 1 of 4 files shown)


nats12:
@gnapse I've assigned you based on your proximity to the design system, and the conversation we had [here](https://github.com/Doist/Issues/issues/16625#issuecomment-2826994527) regarding the issue.

Initially in that issue's comment, I spoke of totally decoupling the `onDismiss` from the action handler. However, having put more thought into it, it seems common practice for the confirm button to then also close the toast. Removing this behaviour would result in quite the breaking change as consumers would have to manually go through all toasts to close them, which may be the desired behaviour 90% of the time anyway.

Instead, I've opted for the reverse option, let's by default close the toast when hitting the confirm action **_unless_** this new `showStickyToast` prop is used in the hook.

Let me know your thoughts 👍🏼

src/toast/toast.stories.tsx
```diff
@@ -88,10 +88,26 @@ export function NotificationToastsStory() {
...
+                            onClick={() => {
+                                const actionLabel = getRandom(actions)
+                                showToast({
+                                    message: `${count.current++}: ${getRandom(message)}`,
+                                    action: {
+                                        label: actionLabel,
+                                        onClick: storybookAction(actionLabel),
+                                    },
+                                    showStickyToast: true,
+                                })
```
gnapse, line 102: I would suggest doing it like this:

```suggestion
                                showToast({
                                    message: `${count.current++}: ${getRandom(message)}`,
                                    action: {
                                        label: actionLabel,
                                        onClick: storybookAction(actionLabel),
                                        closeToast: false,
                                    },
                                })
```

A few things to note about this:

1. Since this is associated to the `action` no longer closing the toast, it feels more contextual to put this flag inside the action object.
2. Calling it `closeToast` is more straightforward, IMO, about what t does.
3. This new flag would, of course, be `true` by default, so that all our toasts keep the current functionality.

What do you think @nats12?

nats12: Thanks @gnapse! Great suggestion, I didn't think to add it to the `action` object. Makes sense.

I've done that now in the last commit. Let me know if there's anything else 👍🏼

src/toast/toast.test.tsx
```diff
@@ -124,6 +124,24 @@ describe('useToast', () => {
         expect(within(screen.getByRole('alert')).getByTestId('test-icon')).toBeInTheDocument()
     })
 
+    it('allows to keep the toast after the action is performed if closeToast is false', () => {
```
gnapse, line 127: I'd suggest putting this in a `describe` block to make it more readable:

```js
describe('if `closeToast` is false', () => {
    it('keeps the toast visible after the action button is clicked', () => {
        // …
    })
})
```

also, there should be a sister test to this one, testing that when `closeToast` is omitted or true, the toast is dismissed when clicking the action button.

PR #931, opened 2025-06-20
fix: Update `BaseField` styling to match the form field spec

rfgamaral:
## Short description

This PR updates the `BaseField` component styling to match the form field spec:

- https://www.figma.com/design/5gTX7MuUxhCIvL6WK87JVA/24Q3-Foundation?node-id=1947-125676&m=dev

## PR Checklist


rfgamaral:
@pawelgrimm Hover styles were actually missing, so I added them now. This should now match the Figma spec much more closely. It's ready to review. Let me know what you think.

rfgamaral:
> I don't think we should apply the hover border color when fields are disabled/read-only

Good call, let me fix that.

PR #939, opened 2025-08-06
fix: Enable banner with an Image to use container full width (with CSS vars)

FilipMam:
## Short description

For the purpose [of adding a QR code banner](https://app.todoist.com/app/task/web-triggered-sidebar-card-with-qr-code-6cWMwvjqRrf3WcM7) to the app, we'd like the banner with the image to use the full width of its container. This is currently not possible, because the banners with images get automatically narrowed to the image's width.

CleanShot 2025-08-06 at 12 04 59@2x

This PR enables the user to choose if the banner with an image should be narrowed (default behaviour) or if it should stretch to the full width.

CleanShot 2025-08-06 at 12 06 49@2x

## Implementation

This solution leverages CSS vars to set the banner's min-width. 

## Reference
- Required for https://app.todoist.com/app/task/web-triggered-sidebar-card-with-qr-code-6cWMwvjqRrf3WcM7

### DIFF
--- a/src/banner/banner.module.css
+++ b/src/banner/banner.module.css
@@ -62,6 +62,12 @@
     display: block;
 }
 
+/* The close button is invisible in this resolution, so we're hiding the container to prevent flex-gap from making extra space */
+.icon:empty,
+.icon:has(.closeButton:only-child) {
+    display: none;
+}
+
 .icon .closeButton {
     display: none;
 }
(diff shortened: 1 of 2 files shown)


engfragui:
I left my notes [here](https://github.com/Doist/reactist/pull/938#issuecomment-3163073148)

FilipMam:
Disclaimer: Technically, this PR is not approved, but it was opened alongside the other [PR](https://github.com/Doist/reactist/pull/938) that introduced a different solution to the same problem. Francesca approved the other PR already and suggested that, out of the two, she prefers the first one, but left me with the final choice-- I'm gonna roll with this one 🙈

PR #945, opened 2025-09-04
fix: Smaller loader in `<Button>` component

engfragui:
## Short description

This PR reduces the size of the loader icon that gets displayed inside a `<Button>` component, when the `loading` prop is true. More specifically, the loader size was `24px` prior to this PR. This PR changes it to `18px`.

## Reference

- [Todoist task](https://app.todoist.com/app/task/web-add-appropriately-sized-loader-spinner-to-upload-photo-flow-6cpWVf4jmRRpqFW7)
- Requested [in Figma here](https://www.figma.com/design/z4zytHrQew1cySsuoVVckt?node-id=1884-226828&m=dev#1411492158)
- Discover the need for this fix while implementing https://github.com/Doist/todoist-web/pull/14237
- Enables https://github.com/Doist/todoist-web/pull/14639

## Demo 

<table>
<tbody>
<tr><td>

https://github.com/user-attachments/assets/00f7b102-362e-44bf-8fa1-6f4887fa9e6a

</td><td>

https://github.com/user-attachments/assets/e4fae595-3f7c-4d84-8e3f-24a2de955091

</td></tr>
</tbody>
</table>

### DIFF
--- a/src/button/button.stories.mdx
+++ b/src/button/button.stories.mdx
@@ -26,14 +26,21 @@ export function Icon() {
     )
 }
 
-export function LoadingButton(props) {
+export function LoadingButton({ disabledWhenLoading = false, ...props }) {
     const [loading, setLoading] = useState(false)
     useEffect(() => {
         if (!loading) return undefined
         const timeout = setTimeout(() => setLoading(false), 3000)
         return () => clearTimeout(timeout)
     }, [loading])
-    return <Button {...props} loading={loading} onClick={() => setLoading(true)} />
+    return (
+        <Button
+            {...props}
+            loading={loading}
+            disabled={disabledWhenLoading ? loading : props.disabled}
+            onClick={() => setLoading(true)}
+        />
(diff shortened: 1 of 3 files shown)


alexmuench:
>  if our current implementation looks good in any variant (see my demo + screenshot in this message)

Only in the primary button, can we change the loading spinner color to “ondark” aka white-ish for now?  
image

frankieyan:
> I feel like I can still see the spinner's background in both the **primary** and **secondary** variants -- however it is true that I cannot in the **tertiary** and **quaternary** variants. Here's a (non-visually appealing) simulation:

Oh yeah, I wasn't sure if we always used the disabled state when `loading={true}`, since we don't in Reactist, so when they are not disabled, they look like this (from your video):

image

Just saw Alex's proposal though and that looks great! Once we implement the new specs we should probably simplify things on our end so that `loading` would automatically put the button in disabled state as well.

PR #1031, opened 2026-04-08
fix: use OIDC trusted publishing instead of npm token for releases

rfgamaral:
## Short description

The release workflow was using `registry-url` in `setup-node`, which overwrote the repo's `.npmrc` and required an `NPM_TOKEN` for authentication. This broke OIDC trusted publishing, which PR #1026 "fixed" by adding the token back. That was the wrong direction: the token is unnecessary when OIDC is configured correctly.

This PR removes `registry-url` from the initial `setup-node` step so the repo `.npmrc` is preserved and OIDC handles npm authentication without a token.

The workflow was also aligned with Typist's release workflow: `package-published` output from semantic-release instead of tag-diff detection, `node_modules` caching, explicit build step, consistent step naming, and `GH_PACKAGES_TOKEN` for GitHub Packages publishing. One difference from Typist: Reactist keeps the GitHub App token (`create-github-app-token`) instead of `GH_REPO_TOKEN`, because Reactist has a branch ruleset on `main` that requires the App as a bypass actor for semantic-release to push release commits. Typist has no such ruleset, but would break if one were added. This is something we should bring back to Typist.

## Pre-merge checklist

- [ ] Update npm trusted publisher config to reference `publish-package-release.yml` (was `release.yml`)


.github/workflows/publish-package-release.yml
```diff
@@ -0,0 +1,110 @@
+name: Package Release
```
doistbot, line 1: [P1] Renaming the workflow file breaks trusted publishing until npm is updated to trust this new path. The PR description already notes the trusted publisher still references `release.yml`, so merging this as-is will make the next `semantic-release` publish fail npm authentication. Either keep the existing filename or land the npm trusted-publisher change before this rename ships.

.github/workflows/publish-package-release.yml
```diff
@@ -0,0 +1,110 @@
...
+              uses: actions/setup-node@v6
+              with:
+                  cache: npm
+                  node-version-file: .node-version
+
+            - name: Cache project 'node_modules' directory
+              id: node-modules-cache
+              uses: actions/cache@v5
+              with:
+                  key: node-modules-cache-${{ hashFiles('**/package-lock.json', '**/.node-version') }}
```
doistbot: [P1] This cache key can reuse stale patched dependencies. The repo runs `patch-package` in `postinstall` and currently carries a patch under `patches/`, but `patches/**/*.patch` is not part of the key. If a patch changes without a lockfile change, the cache hits, `npm ci` is skipped, and the release is built/published with the previous patch contents. Include the patch files in the key, or always run `npm ci`.

release.config.js
```diff
@@ -17,5 +17,12 @@ export default {
             },
         ],
         '@semantic-release/github',
+        [
+            '@semantic-release/exec',
+            {
+                verifyConditionsCmd: 'echo "package-published=false" >> $GITHUB_OUTPUT',
```
doistbot: [P2] These commands make `release.config.js` depend on GitHub Actions internals. Outside Actions, `$GITHUB_OUTPUT` is unset, so a local or `--dry-run` `npx semantic-release` now fails in `verifyConditions` with a shell redirect error before it can compute the next release. Guard the write when `GITHUB_OUTPUT` is absent, or move the output handling into the workflow step instead of the shared release config.

doistbot (commented):
This PR updates the release workflow to rely on OIDC trusted publishing rather than an npm token while cleanly aligning the process with Typist's established setup. These adjustments improve the security and consistency of the deployment pipeline by eliminating unnecessary tokens and standardizing the steps. There are just a few details to address to ensure everything runs smoothly, including coordinating the workflow file rename with the npm trusted publisher configuration, incorporating patch files into the dependency cache key to avoid stale builds, and guarding `$GITHUB_OUTPUT` references in the release configuration so local dry runs do not break.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1031&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ae0a9d680-3316-11f1-9df0-3ba280fec0e9%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ae0a9d680-3316-11f1-9df0-3ba280fec0e9&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

.github/workflows/publish-package-release.yml
rfgamaral: Already covered by the pre-merge checklist. The npm trusted publisher config will be updated before merging.

.github/workflows/publish-package-release.yml
rfgamaral: Good catch. Fixed by adding `patches/**` to the cache key.

release.config.js
rfgamaral: Good point. While publishing should never happen outside CI (OIDC provenance wouldn't work anyway), `semantic-release --dry-run` is a legitimate local use case. Fixed by guarding the `$GITHUB_OUTPUT` writes.

scottlovegrove (commented):
👍🏻

henningmu:
Nice one, thank you 🙌

PR #1037, opened 2026-05-15
feat: add ControlPresentation primitive

pawelgrimm:
## Short description

Adds `ControlPresentation`, a presentational layout primitive for input-like form controls. 

It:
- draws the appropriate border based on state (idle / hover / focus / disabled / readonly / invalid) using html/aria props when appropriate (e.g. `aria-readonly`)
- wraps an arbitrary focusable control passed as `children`, forwards clicks to it
- exposes two slots (`startSlot`, `endSlot`) flanking the control
  - public API also includes `ControlActionButton`, a compact 24×24 button variant sized to fit inside the chrome alongside a 16px icon
- Spacing matches the design specs
It is Intended to be responsible for the presentation of the control in `TextField`, the forthcoming `SelectField`, `PasswordField`, `ComboboxField`, and similar — anything that wants the visual chrome of a text field around an arbitrary control.

## Out of scope

- A new `Button` / `IconButton` `xsmall` size to replace `ControlActionButton`'s dimensions (separate Button-taxonomy spec, out of scope).
- `TextField` / `SelectField` / `TextArea` migration onto this primitive (follow-up PR).
- `BorderedTextField` (the outlined-with-label-inside sibling that will also compose `OutlinedControlContainer`) — follow-up PR.

## PR Checklist


pawelgrimm:
@doistbot /review

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+
+            onClick?.(event)
+
+            const control = controlWrapperRef.current?.firstElementChild as HTMLElement | null
+            if (!control) return
+
+            // Don't re-fire when the user clicked directly on the control —
+            // it handled itself, and re-dispatching would double-activate.
+            if (event.target instanceof Node && control.contains(event.target)) return
+
```
doistbot: [P1] Clicks on interactive elements in slots (like a clear button) currently bubble up and synthetically focus/click the control. While a test expects focus to move to the control, this behavior also triggers `control.click()`/`showPicker()`. For example, clicking "Clear" on a composed `<SelectField>` would unintentionally pop open the dropdown.

Taking focus away from a clicked button also disrupts keyboard accessibility. Consider ignoring clicks that originate from interactive elements outside the control:
```tsx
if (event.target instanceof Node && control.contains(event.target)) return

// Ignore clicks on slot buttons/links so they can behave natively
if (event.target instanceof Element && event.target.closest('button, a, input, select, textarea, [role="button"]')) return
```

doistbot (commented):
This PR introduces the `ControlPresentation` and `ControlActionButton` primitives, establishing a unified, state-responsive visual wrapper for input-like form controls. Centralizing the border chrome, conditional spacing, and native state reflection is a great step toward more composable and consistent form fields across the library. A few refinements have been noted around click-forwarding logic and event bubbling, tightening up TypeScript types and CSS scoping, expanding test coverage, and addressing styling duplication with existing field components.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1037&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A3d589110-507b-11f1-9c5e-041b3218c637%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A3d589110-507b-11f1-9c5e-041b3218c637&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+                try {
+                    control.showPicker()
+                } catch {
+                    // showPicker can throw (e.g. focus state issues) — fall back
+                    // silently; the focus above is the minimum useful behavior.
+                }
+                return
+            }
+
+            isDispatchedReentryRef.current = true
```
doistbot: [P2] If `control.click()` is caught by a child element's handler that calls `e.stopPropagation()`, the event won't bubble back up to `handleWrapperClick`. `isDispatchedReentryRef` will remain `true`, causing the *next* wrapper click to be incorrectly ignored.

Since `.click()` dispatches synchronously, you can prevent this state leak using a `try/finally` block:
```tsx
isDispatchedReentryRef.current = true
try {
    control.click()
} finally {
    isDispatchedReentryRef.current = false
}
```
(You can then safely remove `isDispatchedReentryRef.current = false` from the early return at line 71).

src/control-presentation/control-presentation.module.css
```diff
@@ -0,0 +1,128 @@
...
+    ) {
+        background-color: var(--reactist-field-readonly-background);
+        color: var(--reactist-field-readonly-content);
+    }
+
+    /* Disabled: matches the native disabled attribute (form controls),
+     * the soft-disable ARIA convention (used by Reactist's own Button to
+     * keep elements focusable while announcing as disabled), and the
+     * data-attr convention used by Ariakit/Radix. */
+    &:has(
```
doistbot: [P2] These `:has()` state selectors are matching any descendant of the container, not just the wrapped control. That means slot content can flip the field chrome — e.g. a disabled `ControlActionButton` in `endSlot` will match `[aria-disabled]` here and make an otherwise editable input look disabled. Scope the state selectors to the control subtree so slots don't drive the wrapper state.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+            if (event.target instanceof Node && control.contains(event.target)) return
+
+            // .focus() handles inputs (cursor/keyboard focus). The activation
+            // hop differs by element type:
+            //  - Native <select>: synthetic .click() is blocked by browsers
+            //    for the native dropdown. .showPicker() is the correct API.
+            //  - Everything else (inputs, native <button>, Ariakit Select
+            //    triggers, anchors): .click() either activates the click
+            //    handler or is harmlessly redundant.
+            control.focus()
```
doistbot: [P2] The wrapper forwards activation unconditionally, even when the inner control advertises disabled state via `aria-disabled`/`data-disabled` (the same conventions the CSS treats as disabled). For custom triggers that don't prevent synthetic clicks themselves, clicking the chrome will still activate a disabled-looking control. Guard the forwarded `click()` with the same disabled-state checks before dispatching it.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+                ref={ref}
+                {...restWithoutClick}
+                className={[styles.container, exceptionallySetClassName]}
+                display="flex"
+                alignItems="center"
+                overflow="hidden"
+                borderRadius="standard"
+                onClick={handleWrapperClick}
+            >
+                {startSlot ? (
```
doistbot: [P2] `startSlot` and `endSlot` are typed to accept numbers, but these truthiness checks drop valid `0` content. `startSlot={0}` / `endSlot={0}` currently render nothing. Use an explicit nullish check instead of a truthy check so numeric slot content is preserved.

src/control-presentation/control-action-button.tsx
```diff
@@ -0,0 +1,48 @@
...
+      } & Omit<ComponentProps<typeof IconButton>, 'variant' | 'size'>)
+
+/**
+ * A compact action button intended for `ControlPresentation`'s `endSlot`. Wraps
+ * Reactist's `Button` / `IconButton` with a 24×24, 3px-radius variant sized to fit
+ * the field chrome alongside a 16px icon glyph.
+ */
+export const ControlActionButton = React.forwardRef<HTMLButtonElement, ControlActionButtonProps>(
+    function ControlActionButton({ exceptionallySetClassName, ...props }, ref) {
+        return 'children' in props ? (
```
doistbot, line 37: [P3] The shared props between `Button` and `IconButton` can be extracted to avoid duplicating the `variant` and `exceptionallySetClassName` setup.

```tsx
const sharedProps = {
    ref,
    variant: 'quaternary' as const,
    exceptionallySetClassName: classNames([
        styles.controlActionButton,
        exceptionallySetClassName,
    ]),
}

return 'children' in props ? (
    <Button {...props} {...sharedProps} />
) : (
    <IconButton {...props} {...sharedProps} />
)
```

src/control-presentation/control-presentation.module.css
```diff
@@ -0,0 +1,128 @@
+:root {
+    --reactist-field-height: 32px;
+}
+
+.container {
```
doistbot, line 5: [P2] This adds a second source of truth for the input chrome that still exists in `text-field.module.css` and `select-field.module.css` (32px field height, idle/hover/focus/error borders, read-only styling, slot spacing). Until those components are migrated, any visual/state fix will have to be made in multiple places. Please either switch the existing field components to consume `ControlPresentation` in this PR, or extract the shared chrome rules into one reused path before publishing the new primitive.

src/control-presentation/control-action-button.tsx
```diff
@@ -0,0 +1,48 @@
...
+    | ({
+          icon?: React.ReactElement
+      } & Omit<ComponentProps<typeof IconButton>, 'variant' | 'size'>)
+
+/**
+ * A compact action button intended for `ControlPresentation`'s `endSlot`. Wraps
+ * Reactist's `Button` / `IconButton` with a 24×24, 3px-radius variant sized to fit
+ * the field chrome alongside a 16px icon glyph.
+ */
+export const ControlActionButton = React.forwardRef<HTMLButtonElement, ControlActionButtonProps>(
```
doistbot, line 26: [P2] `ControlActionButton` creates a second compact field-action implementation alongside the existing `TextField`/`PasswordField` path: `PasswordField` still renders a raw `IconButton`, and `text-field.module.css` already shrinks slotted buttons to 24px. That leaves two parallel ways to express the same field affordance, with styling drift already possible. Please migrate the existing field action usage to this wrapper here, or move the 24px field-action treatment into a shared button size/token that both paths reuse.

src/control-presentation/control-presentation.module.css
```diff
@@ -0,0 +1,128 @@
...
+    background: var(--reactist-field-background);
+    color: var(--reactist-field-content);
+
+    /* Read-only: matches native :read-only on form controls (scoped to
+     * <input> / <textarea> to avoid catching unrelated elements that
+     * default to :read-only), the ARIA convention (aria-readonly="true"),
+     * and the data-attr convention used by Ariakit/Radix-style components.
+     * The :not(='false') exclusion treats aria-readonly="false" /
+     * data-readonly="false" as "not read-only" per the ARIA spec. */
+    &:has(
```
doistbot: [P2] The project guidelines explicitly state `No CSS nesting - keep selectors flat` for component styling. Please un-nest the selectors in this file (e.g., flatten `&:has(...)` to `.container:has(...)`, `& > *` to `.control > *`, and `&.controlActionButton` to `.controlActionButton.controlActionButton.controlActionButton`) to align with established conventions.

src/control-presentation/control-action-button.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import { Button, IconButton } from '../button'
+
+import styles from './control-presentation.module.css'
+
+import type { ComponentProps } from 'react'
+
+export type ControlActionButtonProps =
+    | ({
+          children: React.ReactElement
```
doistbot: [P2] This wrapper narrows the underlying button APIs more than intended: `Button` accepts any `ReactNode` label and `IconButton` accepts its own icon type, but this type only allows `ReactElement` for both. That means supported cases like `<ControlActionButton>Clear</ControlActionButton>` stop type-checking even though the inner component can render them. Reuse the underlying prop types (or widen `children`/`icon`) so the wrapper doesn't drop valid Button/IconButton inputs.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+    /**
+     * The control element (an `<input>`, an Ariakit Select trigger,
+     * `<select>`, etc.). Attributes belonging to the control — `type`,
+     * `value`, `onChange`, `readOnly`, `disabled`, `aria-invalid`, and so on —
+     * are set on this element directly. The wrapper chrome (read-only
+     * background, disabled state, error border) derives from those attributes.
+     *
+     * Click handlers belong on the control itself, not on this wrapper.
+     */
+    children: React.ReactNode
```
doistbot: [P2] `children` is typed as `ReactNode`, but the implementation later assumes exactly one element (`firstElementChild`) for focus/click forwarding and only styles direct element children. Tighten this to a single `ReactElement` (or add runtime validation) so the public API can't accept unsupported strings, multiple controls, or multi-node fragments.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+} & ObfuscatedClassName &
+    Omit<ComponentProps<typeof Box>, 'className'>
```
doistbot: [P2] Extending full `ComponentProps<typeof Box>` leaks Box's polymorphic `as` prop into this API, but the component is hard-coded around a `div` wrapper and forwards an `HTMLDivElement` ref. `as="label"` / `as="button"` will type-check today while changing semantics and making the ref type wrong. Omit `as` (and ideally narrow the forwarded Box props to the few supported layout/DOM props) to keep the abstraction stable.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,141 @@
...
+                    control.showPicker()
+                } catch {
+                    // showPicker can throw (e.g. focus state issues) — fall back
+                    // silently; the focus above is the minimum useful behavior.
+                }
+                return
+            }
+
+            isDispatchedReentryRef.current = true
+            control.click()
```
doistbot: [P2] This unconditionally dispatches a second click event for every wrapper-background click on any non-`<select>` control. For the main text-entry cases this component is meant to wrap (`TextField`, `PasswordField`, etc.), `focus()` already provides the needed behavior, so the extra `.click()` just adds another full event/bubble pass through React and any control-level `onClick` handlers. Please gate the synthetic click to controls that actually need activation (e.g. button-like/custom triggers or picker-like inputs) and skip it for plain text inputs/`textarea`.

src/control-presentation/control-presentation.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { ControlPresentation } from './control-presentation'
+
+describe('ControlPresentation', () => {
+    it('renders the control passed as children', () => {
+        render(
+            <ControlPresentation>
+                <input aria-label="Subject" data-testid="subject" />
+            </ControlPresentation>,
+        )
+        const control = screen.getByTestId('subject')
```
doistbot: [P3] Project conventions specify querying elements by role. Since the nested controls have an `aria-label`, consider replacing `getByTestId` with `screen.getByRole('textbox', { name: 'Subject' })` (and the equivalent `getByRole` for buttons) here and throughout the file.

src/control-presentation/control-presentation.test.tsx
```diff
@@ -0,0 +1,175 @@
...
+        expect(control).not.toHaveFocus()
+
+        userEvent.click(container.firstElementChild as Element)
+        expect(control).toHaveFocus()
+    })
```
doistbot: [P2] This test verifies that the inner control receives focus, but it leaves the synthetic `.click()` and the native `<select>` `.showPicker()` fallbacks untested. Consider adding tests to ensure these core activation mechanics trigger correctly when the wrapper is clicked.

src/control-presentation/control-action-button.tsx
doistbot, line 26: [P2] This new exported primitive ships without any dedicated test coverage. Please add tests for both branches of the union (`children` -> `Button`, `icon` -> `IconButton`) so regressions in the prop discrimination and the injected compact styling don't slip through unnoticed.

pawelgrimm: Migrations are not in scope here — we will integrate in a follow-up PR.

src/control-presentation/control-presentation.module.css
pawelgrimm: Migrations are not in scope here — we will integrate in a follow-up PR.

src/control-presentation/outlined-control-container.tsx
```diff
@@ -0,0 +1,110 @@
...
+            if (isDispatchedReentryRef.current) return
+
+            onClick?.(event)
+
+            // Find the first focusable form control descendant. More robust
+            // than firstElementChild — works whether the control is the
+            // direct child (ControlPresentation) or nested inside a column
+            // layout (BorderedTextField).
+            const wrapper = controlWrapperRef.current
+            const control = wrapper?.querySelector<HTMLElement>('input, select, textarea')
```
doistbot: <a href="#">P1</a>  This query searches the entire wrapper subtree, including slots. If `startSlot` contains a focusable element (like a country-code `<select>`), this will incorrectly target the slot element instead of the primary control. Furthermore, restricting the query to `'input, select, textarea'` prevents click forwarding from working on custom controls (like a `<button role="combobox">`). Consider having consumers pass a `controlRef`, or querying a specific marker (e.g., `[data-primary-control]`) so the main control is identified unambiguously and supports arbitrary focusable elements.

src/control-presentation/outlined-control-container.tsx
```diff
@@ -0,0 +1,110 @@
...
+            if (!control) return
+
+            if (event.target instanceof Node && control.contains(event.target)) return
```
doistbot, line 106: <a href="#">P1</a>  Clicks on interactive elements in the slots (like a "Clear" button) will bubble to the wrapper, bypassing this check because they are outside the targeted `control`. This causes the wrapper to synthetically focus and click the primary control, stealing focus from the slot button and potentially triggering unwanted actions. Ignore clicks that originate from interactive slot elements (e.g., `if (event.target instanceof Element && event.target.closest('button, a, [role="button"]')) return`).

src/control-presentation/outlined-control-container.module.css
```diff
@@ -0,0 +1,76 @@
...
+) {
+    background-color: var(--reactist-field-readonly-background);
+    color: var(--reactist-field-readonly-content);
+}
+
+/* Disabled: matches the native disabled attribute (form controls),
+ * the soft-disable ARIA convention (used by Reactist's own Button to
+ * keep elements focusable while announcing as disabled), and the
+ * data-attr convention used by Ariakit/Radix. */
+.container:has(
```
doistbot, line 40: <a href="#">P1</a>  These `:has()` state selectors match any descendant of the container. Because `OutlinedControlContainer` wraps both the slots and the main control, a disabled element in a slot (e.g., a disabled "Clear" button in `endSlot`) will incorrectly trigger this rule and apply disabled styling to the entire field wrapper. Consider scoping these selectors to target the main control specifically (e.g., using a specific wrapper class), or passing state props directly.

src/control-presentation/control-presentation.tsx
```diff
@@ -0,0 +1,72 @@
...
+        ref,
+    ) {
+        return (
+            <OutlinedControlContainer
+                ref={ref}
+                borderRadius="small"
+                onClick={onClick}
+                exceptionallySetClassName={classNames(styles.container, exceptionallySetClassName)}
+            >
+                {startSlot ? (
```
doistbot: <a href="#">P2</a>  `startSlot` and `endSlot` are typed to accept numbers via `SlotContent`, but these truthiness checks drop valid `0` content, rendering nothing instead of the number `0`. Use an explicit nullish check (e.g., `startSlot != null`) so numeric slot content is preserved.

src/control-presentation/control-action-button.tsx
doistbot: <a href="#">P2</a>  This wrapper narrows the underlying button APIs more than intended. `Button` accepts any `ReactNode` label and `IconButton` accepts `ReactElement | string` for `icon`, but this union requires `ReactElement` in both branches. Callers like `<ControlActionButton>Clear</ControlActionButton>` or `icon="×"` now fail type-check even though the wrapped components support them. Reuse the underlying `children`/`icon` prop types instead of redefining them here.

src/control-presentation/outlined-control-container.tsx
```diff
@@ -0,0 +1,110 @@
...
+                    // silently; the focus above is the minimum useful behavior.
+                }
+                return
+            }
+
+            // Guard against the synchronous re-entry from control.click() bubbling
+            // back to this handler. Always clear after the dispatch returns so the
+            // guard cannot latch if a consumer's onClick stops propagation.
+            isDispatchedReentryRef.current = true
+            control.click()
```
doistbot: <a href="#">P2</a>  This adds a second click dispatch for every wrapper click on non-`<select>` controls. For the main text-entry cases this primitive is meant to wrap (`input[type=text|email|password|search|number|... ]`, `textarea`), `focus()` already gives the needed behavior, so the extra `.click()` just runs another full event/bubble cycle through React and any control `onClick` handlers. Please gate the synthetic click to controls that actually need activation (for example picker-like inputs) and skip it for plain text-entry controls.

src/control-presentation/control-action-button.tsx
doistbot, line 26: <a href="#">P2</a>  `ControlActionButton` remains exported without any dedicated test coverage. Please add tests to verify both branches of its union API (rendering `Button` when passed `children`, and `IconButton` when passed `icon`) so the prop discrimination works correctly.

src/control-presentation/outlined-control-container.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+                    <input aria-label="control" />
+                </OutlinedControlContainer>,
+            )
+            expect(container.firstElementChild).toHaveClass(styles.borderRadiusLarge!)
+            expect(container.firstElementChild).not.toHaveClass(styles.borderRadiusSmall!)
+        })
+    })
+
+    describe('click-to-focus dispatch', () => {
+        it('focuses the inner control when the wrapper is clicked', async () => {
```
doistbot, line 78: <a href="#">P2</a>  This test confirms that the wrapper forwards focus, but the suite still lacks a test for the synthetic `.click()` forwarding. (For instance, the stop-propagation test on line 115 would still pass if `control.click()` were deleted from the implementation). Please add a test that asserts an `onClick` handler on the inner control is called when the wrapper is clicked.

src/control-presentation/control-presentation.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+        )
+        const control = screen.getByTestId('subject')
+        expect(control).toHaveAttribute('type', 'email')
+        expect(control).toHaveAttribute('placeholder', 'you@example.com')
+        expect(control).toHaveAttribute('data-custom', 'yes')
+        expect(control).toHaveAttribute('readonly')
+        expect(control).toBeDisabled()
+    })
+
+    it('focuses the control when a non-interactive startSlot is clicked', async () => {
```
doistbot, line 57: <a href="#">P1</a>  This only covers a decorative slot. The main new surface here is interactive slot content (`ControlActionButton`, clear buttons, chevrons), and a regression there will forward the wrapper activation into the inner control — for example, clicking a slotted button on a wrapped `<select>` can open the picker. Please add a behavior test with a slotted button asserting the button handles the click and the wrapped control does not get activated.

src/control-presentation/control-presentation.test.tsx
```diff
@@ -0,0 +1,120 @@
...
+    it('merges exceptionallySetClassName onto the wrapper', () => {
+        const { container } = render(
+            <ControlPresentation exceptionallySetClassName="custom-class">
+                <input aria-label="Subject" />
+            </ControlPresentation>,
+        )
+        expect(container.firstElementChild).toHaveClass('custom-class')
+    })
+
+    it('endSlot accepts multi-child composition', () => {
```
doistbot, line 102: <a href="#">P2</a>  The slot rendering tests never exercise the `number` branch of `SlotContent`. `0` is the edge case most likely to break here, and this API explicitly supports numeric slot content for things like counters/units. Add a case for `startSlot={0}` or `endSlot={0}` so valid numeric content can't be silently dropped.

doistbot (commented):
Thanks pawelgrimm for your contribution 😇. This introduces a solid foundation for unifying form control presentation and standardizing input layouts.

Few things worth tightening:
- Refine the wrapper's DOM queries and event routing so interactive slot elements don't steal focus, trigger unintended synthetic clicks, or accidentally apply disabled styling to the entire container via broad `:has()` selectors.
- Loosen `ControlActionButton` types to fully support the underlying button APIs, and update slot rendering checks to preserve valid numeric content like `0`.
- Gate the wrapper's synthetic `.click()` dispatch to only activate picker-style controls, avoiding redundant event bubbling for plain text entries.
- Expand test coverage for `ControlActionButton` branches, synthetic click forwarding, and edge cases like interactive or numeric slot content.

<details>

- **[P3] src/control-presentation/control-presentation.test.tsx:18**: Project conventions specify querying elements by role. Since the inner control has an `aria-label`, consider replacing `getByTestId('subject')` with `screen.getByRole('textbox', { name: 'Subject' })` here and throughout the file to align with established accessibility testing patterns.
- **[P3] src/control-presentation/control-action-button.tsx:11**: `ControlActionButton` is a new public export, but its branch-specific props are undocumented. The repo’s component conventions call for JSDoc on each public prop so consumers get proper hover/docs coverage. Please add prop comments for the `children` and `icon` branches here (or refactor this union into documented interfaces).

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1037&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Adcb62480-5ade-11f1-9174-08380da5af9e%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Adcb62480-5ade-11f1-9174-08380da5af9e&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

PR #1048, opened 2026-05-26
feat!: Revamped Avatar component

pawelgrimm:
## Summary

We are bringing the Avatar component up to spec with the design, including:
- Support for specific sizes
- Circle and rounded shapes
- Dynamic image URL selection (based on `srcset`) 
- Image load failure handling
- Wider array of initials/meta colors
- Accessibility handling for custom or decorative labels

We also added a new documentation page for the component, which walks through all the features it supports. The new and improved stories feature all the contributors to this repo from the past year 😁

We also removed some features that we don't really need:
- `media` query-based responsive sizing
- Deriving initials from email (you can just pass the user's email as `name`, if desired)

Sorry for the rather large PR — I'd split it up, but that would mean pushing accompanying stories, docs, and tests in a separate PR, which isn't great either. Lmk if you feel strongly about having a smaller PR in this case and I can do my best to rework it.

## References

- [🎨 Product Library - Global / Avatar](https://www.figma.com/design/xo9yAsH8PQUpi0eTJh9pmR/Product-Library---Global?node-id=7847-14757&m=dev)
- [💬 Avatar Component Revamp](https://twist.com/a/1585/ch/81455/t/5735955/)

## 📸 Demo

<table>
<tbody>
<tr><td>

CleanShot 2026-05-27 at 11 57 05

</td><td>

CleanShot 2026-05-27 at 11 57 19

</td></tr>
</tbody>
</table>

## PR Checklist


src/avatar/avatar.tsx
```diff
@@ -1,81 +1,187 @@
...
+     */
+    'data-testid'?: string
+}
+
+/**
+ * Displays an avatar from an image URL, a source map keyed by intrinsic
+ * image width, or initials derived from the provided name (with a background
+ * color).
+ */
+function Avatar({ image, ...props }: AvatarProps) {
```
doistbot: <a href="#">P2</a>  According to `CLAUDE.md`, components should use `React.forwardRef` with a named function matching the component name. Please wrap `Avatar` in `React.forwardRef` and pass the `ref` through to the underlying `Box` (you'll likely also need to forward it through `AvatarContent`).

src/avatar/avatar.test.tsx
```diff
@@ -1,64 +1,233 @@
...
+    function failCurrentAvatarImage(currentSrc: string) {
+        const image = screen.getByRole('img', { name: 'Jane Doe' })
+        Object.defineProperty(image, 'currentSrc', {
+            configurable: true,
+            value: currentSrc,
+        })
+        fireEvent.error(image)
+    }
+
+    it('renders a string image URL', () => {
```
doistbot, line 22: <a href="#">P1</a>  According to `CLAUDE.md`, every component should have an accessibility test using `jest-axe`. Please add a base a11y check for the `Avatar` component (e.g., `expect(await axe(container)).toHaveNoViolations()`).

src/avatar/avatar.tsx
```diff
@@ -1,81 +1,187 @@
...
-    const sizeClassName = getClassNames(styles, 'size', avatarSize)
+    'data-testid': testId,
+}: AvatarProps) {
+    const imageSources = getSources(image, size)
+    const [failedImageSources, setFailedImageSources] = React.useState<string[]>([])
+    const availableImageSources = getAvailableImageSources(imageSources, failedImageSources)
+
+    const initials = getInitials(name)
+    const hasInitials = initials !== ''
+    const label = alt ?? name
```
doistbot: <a href="#">P2</a>  `label` falls back to the raw `name`, but initials/color selection already normalize whitespace first. With `name="   "` (or other whitespace-only input), this renders as an empty avatar while still producing `alt="   "` for images or skipping `aria-hidden` for the empty state, so it stops being decorative and ends up with a blank accessible name. Please derive the default label from the normalized name instead of the raw prop.

src/avatar/avatar.tsx
```diff
@@ -1,81 +1,187 @@
...
-        ? {
-              backgroundImage: `url(${avatarUrl})`,
-              textIndent: '-9999px', // hide the initials
-          }
-        : {
-              backgroundColor: colorList[emailToIndex(user.email, colorList.length)],
-          }
-
-    const sizeClassName = getClassNames(styles, 'size', avatarSize)
+    'data-testid': testId,
```
doistbot: <a href="#">P2</a>  This closes over the root element props instead of forwarding them. The previous `Avatar` implementation spread `...props` to `Box`, and sibling components in this library still follow that pattern, so JS consumers could pass `id`, `title`, `aria-describedby`, `as`, etc. After this change only `data-testid` survives, which makes the API more brittle and will force more ad-hoc prop additions later. Please keep a `...props` rest object and type `AvatarProps` to include the root `Box`/HTML attributes rather than special-casing individual ones.

src/avatar/avatar.test.tsx
```diff
@@ -1,64 +1,233 @@
...
+        })
+    })
+
+    it('defaults to circle shape', () => {
+        render(<Avatar data-testid="avatar" size={36} name="Jane Doe" />)
+
+        expect(screen.getByTestId('avatar')).toHaveClass('shape-circle')
+    })
+
+    it('uses custom alt text as the accessible label', () => {
```
doistbot, line 257: <a href="#">P2</a>  The new accessibility tests only cover custom `alt` on the `` branch. This PR also applies `alt` to the initials fallback, and that path is easy to regress without noticing because it swaps from `alt` to `role="img"`/`aria-label`. Please add a case like `<Avatar size={36} name="Jane Doe" alt="Account avatar" />` and assert the initials fallback is exposed with that label.

doistbot (commented):
Thanks pawelgrimm for your contribution 😊. The new fallback handling and dynamic image selection look solid, and the updated documentation featuring contributor avatars is a great touch.

Few things worth tightening:
- Wrap the component in `React.forwardRef` and restore the `...props` rest spread to the root element to maintain a flexible, standard API.
- Derive the default fallback label from the normalized `name` instead of the raw prop so whitespace-only inputs don't yield a blank accessible name.
- Add a base `jest-axe` test to the suite per our guidelines, alongside a targeted test verifying that custom `alt` text is correctly exposed on the initials fallback.

<details>
<summary>Optional follow-up notes (6)</summary>

- **[P3] src/avatar/avatar.stories.tsx:9**: This duplicates the supported-size list that already exists as `AVATAR_SIZES` in `utils.ts`. If the component contract changes, the stories/playground can drift from the actual runtime/type surface. Import the shared constant here instead of maintaining a second copy.
- **[P3] src/avatar/utils.ts:74**: `normalizeAvatarName` already replaces all whitespace sequences with a single space. Using `.split(' ')` here produces identical results and avoids running the Unicode regex engine a second time.
- **[P3] src/avatar/utils.ts:147**: Consider adding a fast-path `if (failedSources.length === 0) return imageProps;` before checking `imageProps.sources`. This avoids re-filtering the array and reconstructing the image properties object on every render for the typical happy path where no image sources have failed.
- **[P3] src/avatar/avatar.tsx:102**: `getInitials(name)` and `getAvatarMetaColorIndex(name)` now run on every render, even when `availableImageSources` is truthy and the component never uses the initials/meta-color path. For image-heavy avatar lists, that adds avoidable normalization/segmentation/hash work to the hot path. Computing these only when we actually fall back to initials would avoid that per-render overhead.
- **[P3] src/avatar/avatar.mdx:25**: This docs page now renders all of the image-heavy canvases inline, and most of those stories use live GitHub avatar URLs. Opening `Avatar.mdx` will trigger roughly 26 external image requests up front, which makes Storybook/Chromatic slower and less deterministic than it needs to be. Using local fixtures (or trimming the number of inline canvases) would avoid that resource overhead.
- **[P3] src/avatar/avatar.tsx:113**: This new `metaColor-*` class family is camel-cased, but Reactist’s CSS naming convention is dash-separated (`variant-primary`, `size-small`, etc.). Please rename these lookups/classes to something like `meta-color-*` so Avatar stays consistent with the rest of the library.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1048&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A4b9b2a10-59ec-11f1-8004-9bdcecb5905d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A4b9b2a10-59ec-11f1-8004-9bdcecb5905d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/avatar/avatar.tsx
pawelgrimm: ✅ f18bcf8b

src/avatar/avatar.test.tsx
pawelgrimm: ✅ 748df881

src/avatar/avatar.tsx
pawelgrimm: ✅ 1f07bcc7

src/avatar/avatar.tsx
pawelgrimm: ✅ f18bcf8b

src/avatar/avatar.test.tsx
pawelgrimm: ✅ 1f07bcc7

src/avatar/avatar.tsx
```diff
@@ -1,81 +1,218 @@
...
-            {...props}
-        >
-            {userInitials}
-        </Box>
+        <AvatarContent
+            as={as}
+            ref={ref}
+            // Allows `AvatarContent` to remount when the image map changes,
+            // which resets error states
+            key={getAvatarImageIdentityKey(image)}
```
rmartins90: Small concern with putting the reset key on `AvatarContent`: this now remounts the actual root element whenever `image` changes.

That feels okay for a plain `div`, but with `as="button"` or `as="a"` it can drop focus or any DOM state on the interactive element. What do you think?

src/avatar/avatar.tsx
```diff
@@ -1,81 +1,218 @@
...
+                className={classNames(
+                    styles.avatar,
+                    styles[`shape-${shape}`],
+                    metaColorIndex !== undefined && styles[`meta-color-${metaColorIndex}`],
+                    !availableImageSources && !hasInitials && styles.empty,
+                    exceptionallySetClassName,
+                )}
+                style={getAvatarStyle(size)}
+                data-testid={testId}
+                aria-hidden={isDecorative || undefined}
```
rmartins90: That root can now be polymorphic because of as. So this is possible:

```tsx
<Avatar
    as="button"
    aria-label="Open profile"
    alt=""
    image="avatar.png"
    size={36}
/>
```

In this scenario the whole button gets `aria-hidden="true"`, so assistive tech won’t see the button. The decorative state should hide only the visual child, not the root element, unless the caller explicitly passes aria-hidden. Makes sense?

src/avatar/avatar.test.tsx
```diff
@@ -1,64 +1,305 @@
 import * as React from 'react'
 
-import { render, screen } from '@testing-library/react'
+import { fireEvent, render, screen } from '@testing-library/react'
```
rmartins90, line 3: NIT: our test guidance in `CLAUDE.md` says to use `userEvent` instead of `fireEvent`.

I realize image `error` is not really a user interaction, so this may be one of those awkward edge cases where `fireEvent` is the practical option. If we want to keep it, maybe worth adding a small helper/comment so it reads as intentional rather than drifting from the convention.

src/avatar/avatar.mdx
```diff
@@ -0,0 +1,189 @@
...
+
+<Title />
+
+<Subtitle>Image, initials, and empty-state avatar primitive.</Subtitle>
+
+## Basic usage
+
+Use `Avatar` for people by default. Pass `size`, `name`, and an optional
+`image`; `name` supplies the default accessible label, the initials fallback,
+and the deterministic meta color used when initials render.
```
rmartins90, line 24: NIT: I'm curious, are these forced line breaks intentional or coming from a formatter?

rmartins90 (approved):
Left 3 minor comments @pawelgrimm. Other than that, it looks good to me

src/avatar/avatar.tsx
pawelgrimm: Yeah, that's fair. I'll push the root element up to this component and just reset the image itself.

src/avatar/avatar.mdx
pawelgrimm: This is just how the agent decided to format it and it's not enforced by Prettier or anything like that.

src/avatar/avatar.test.tsx
pawelgrimm: Yeah, userEvent is kinda limited in the sense that it can't force an error, which to be fair, isn't something the user can do either 😅

We _could_ make the error happen in other ways (network mock + error response), but this was the most simple way.

src/avatar/avatar.tsx
pawelgrimm: ✅ [7f9421e](https://github.com/Doist/reactist/pull/1048/commits/7f9421ea01b9c1effaf122057b9fa57f714719d5)

src/avatar/avatar.tsx
pawelgrimm: Ah, good point!

✅ 30fca01

src/avatar/avatar.test.tsx
pawelgrimm: ✅ [79082f0](https://github.com/Doist/reactist/pull/1048/commits/79082f0e49bf71a918083f77d81d89a5283fd90c)

scottlovegrove:
@pawelgrimm Don't the changes in this now mean the Avatar component is no longer responsive? What if I'm looking at the web app on mobile? I'm now getting the same avatar size as on desktop, right? 

v32 drops the responsive `size` API. In v31 `size` accepted `ResponsiveProp<AvatarSize>` would render smaller avatars on mobile and larger on desktop from a single element. v32's `size` is a required single number and also drives `srcSet` selection, so there's no supported way to vary it per viewport short of duplicating the avatar behind media queries or hacking `--reactist-avatar-size` via `exceptionallySetClassName`. Was dropping responsive sizing intentional? If so, what's the recommended migration for responsive avatars?

pawelgrimm:
> @pawelgrimm Don't the changes in this now mean the Avatar component is no longer responsive? What if I'm looking at the web app on mobile? I'm now getting the same avatar size as on desktop, right?

@scottlovegrove Yes 😅

> v32 drops the responsive `size` API. In v31 `size` accepted `ResponsiveProp<AvatarSize>` would render smaller avatars on mobile and larger on desktop from a single element. v32's `size` is a required single number and also drives `srcSet` selection, so there's no supported way to vary it per viewport short of duplicating the avatar behind media queries or hacking `--reactist-avatar-size` via `exceptionallySetClassName`. Was dropping responsive sizing intentional? If so, what's the recommended migration for responsive avatars?

I dropped the responsive sizing intentionally since 1) we weren't using it internally anywhere, and 2) it would have made the `srcSet` selection and some of the CSS a lot more complicated, so I decided to drop it.

I'm not at all opposed to you adding it back in, but if I were you, I'd just duplicate the avatar behind media-query driven containers OR dynamically select a size in JS based on the screen size.

scottlovegrove:
> I dropped the responsive sizing intentionally since ... we weren't using it internally anywhere

_sigh_ 🙄 

> I'm not at all opposed to you adding it back in, but if I were you, I'd just duplicate the avatar behind media-query driven containers OR dynamically select a size in JS based on the screen size.

How does this not go against the principles of this library? It _should_ be responsive, everything was built for responsiveness, it's a web app UI package!

gnapse:
> We also removed some features that we don't really need:
> 
>  * `media` query-based responsive sizing

I'm also a bit concerned about removing functionality in this way.

Does this mean then that we're going to remove the `ResponsiveProp<T>` feature entirely? Or was it just for the Avatar? If it's the latter, then why just the avatar? Why is this feature not bothering in the other places where it is currently supported?

pawelgrimm:
> How does this not go against the principles of this library? It _should_ be responsive, everything was built for responsiveness, it's a web app UI package!
> Does this mean then that we're going to remove the ResponsiveProp<T> feature entirely? Or was it just for the Avatar? If it's the latter, then why just the avatar? Why is this feature not bothering in the other places where it is currently supported?

I’m sorry this blocked you from upgrading, @scottlovegrove. I prepared https://github.com/Doist/reactist/pull/1058 for you, which restores the previous Avatar implementation as `DeprecatedAvatar` while we figure out the stable API for the newer component. Feel free to review and merge it if it’s useful to you.

A robust component library should bias towards responsiveness, I won't disagree with you there, but I genuinely thought I was acting in line with the existing conventions in this library. Reactist supports responsive behavior today in that we have a responsive prop system, but it is applied selectively. From what I can see, responsive props are used exclusively in the layout primitives (`Box`, `Inline`, `Stack`, and `Columns`). There are plenty of components with `size`-type props that are not responsive: `Button`, `Loading`, `Spinner`, and even `Text` and `Heading`. I’d argue responsive typography sizing is at least as important to responsive UI as Avatar sizing, but `Text::size` and `Heading::size` are both static today.

Responsive sizing on Avatar seemed more similar to the latter set of component, so I dropped it given the new implementation complexity around numeric sizes and `srcSet`/`sizes`, not to mention the forthcoming avatar groups (https://github.com/Doist/reactist/pull/1051, https://github.com/Doist/reactist/pull/1052).

> I'm also a bit concerned about removing functionality in this way.

I think you raise a good point, @gnapse! Should we add and maintain features in Reactist that we don’t use internally? If yes, why and what criteria should we use for deciding that? How should we handle deprecating functionality?

Given the renewed interest in this library, I think it's a good time to codify our stance on these kinds of principles. It'll be much easier to make calls in a consistent and predictable manner. @frankieyan Is this something you'd be interested in taking point on as part of your component modernization work?

frankieyan:
Yeah, I can bring this up with the designers and see where responsiveness stands for us. I would prefer that we roll the responsive prop into the new Avatar if it doesn't overly complicate it, over having to keep both copies around for long, as I've just started to remove some of our older deprecated components. @pawelgrimm would this make the follow-up avatar groups difficult to implement? The last thing I'd want to do is to delay your timelines.

Baking responsiveness into our products would make better experiences. It's never something we spent alot of effort on beyond making it serviceable, and comparing Todoist's mobile app vs mobile web, our (web) avatars, which aren't even using Reactist components, are definitely undersized. That said, I'd like to get product and design's input first and ensure our apps can/will make use of them before we invest our time here.

Ultimately, I think we should cater to our products' requirements first and include a similar callout as Typist: 

> Whilst everyone is welcome to fork or use this package in their own products, development decisions are centered around Doist product requirements.

And if we end up deciding mobile web experiences are not something we want to improve in the near term, I'll make that explicit both here in Reactist and in the handbook.

pawelgrimm:
>  @pawelgrimm would this make the follow-up avatar groups difficult to implement? 

@frankieyan That's exactly what I'm afraid of. I'll try reworking the API to make responsive sizing fit in, but then I might have to abandon the approach I'm taking to clip avatars and just require consumers to pass in the background color.

I'm talking about the space between each avatar in the group, as you see here: 
CleanShot 2026-06-04 at 17 37 49@2x

One approach is to use a background-colored border around each avatar. The other is to clip the avatars so that the background content is shown through the gap. I thought the latter had much better ergonomics... at the cost of massively complicated CSS 🙈

> The last thing I'd want to do is to delay your timelines.

Oh don't worry, I've already de-prioritized those components for now 😅 I was planning on picking them back up when I have some more slack.

frankieyan:
> One approach is to use a background-colored border around each avatar. The other is to clip the avatars so that the background content is shown through the gap. I thought the latter had much better ergonomics... at the cost of massively complicated CSS 🙈

Do we know what the designers' intentions are? Currently in both Todoist and Twist, these stacked avatars are implemented with borders:

image

Not saying it looks better this way, but it'd be simpler, and are already implemented

pawelgrimm:
> Do we know what the designers' intentions are? Currently in both Todoist and Twist, these stacked avatars are implemented with borders:
> image
> 
> Not saying it looks better this way, but it'd be simpler, and are already implemented

@frankieyan I'm almost certain that the "border" is supposed to be the same exact color as the background under it. I remember fixing an issue for something like this once, where we had one of these on a button (with hover states) and the hover state looked bad (because it used the idle background color).

frankieyan:
@pawelgrimm go ahead and implement the avatar groups as you've originally planned. We'll figure it out if responsiveness becomes a requirement later on, there's no point blocking this work over it now. Worst case is that we might have to define all the clipping paths at each size, but it's encapsulated in one component, I don't think it would be too bad (famous last words!)

frankieyan:
@scottlovegrove @gnapse if you'd like to take part in the discussion,(and automations is in the centre of it) it's happening [here](https://comms.todoist.com/69/inbox/t/Cc8JVu65JArRFk49ehSsy/).

PR #1049, opened 2026-05-27
feat(tooltip): expose `--reactist-tooltip-font-size` custom property

pawelgrimm:
## Description

We are adding a CSS custom property so consumers can set the default font-size used in `<Tooltip>`.

## PR Checklist

### DIFF
--- a/src/tooltip/tooltip.module.css
+++ b/src/tooltip/tooltip.module.css
@@ -1,5 +1,10 @@
+:root {
+    --reactist-tooltip-font-size: inherit;
+}
+
 .tooltip {
     text-overflow: ellipsis;
     max-width: 300px;
     z-index: var(--reactist-stacking-order-tooltip);
+    font-size: var(--reactist-tooltip-font-size);
 }
--- a/src/tooltip/tooltip.mdx
+++ b/src/tooltip/tooltip.mdx
@@ -1,0 +1,54 @@
+import { Meta, Story, Canvas, Controls, Description } from '@storybook/addon-docs/blocks'
+import { Tooltip } from './tooltip'
+import * as TooltipStories from './tooltip.stories'
+
+<Meta of={TooltipStories} />
+
+# Tooltip
+
+<Description of={Tooltip} />
+
+<Canvas>
+    <Story of={TooltipStories.TooltipPlayground} />
+</Canvas>
+
+## Examples
+
+### Rich content
+
+<Canvas>
+    <Story of={TooltipStories.TooltipRichContent} />
+</Canvas>
+
+### Custom z-index
+
+<Canvas>
+    <Story of={TooltipStories.TooltipCustomZIndex} />
+</Canvas>
+
+### Global context
+
+<Canvas>
+    <Story of={TooltipStories.TooltipGlobalContext} />
+</Canvas>
+
+### Imperative control
+
+<Canvas>
+    <Story of={TooltipStories.TooltipImperativeControl} />
+</Canvas>
+
+## Props
+
+<Controls of={TooltipStories.TooltipPlayground} />
+
+## CSS custom properties
+
+The following CSS custom properties are available to customize the tooltip component appearance. The values appearing below are the default values.
+
+```css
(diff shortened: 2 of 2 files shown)


src/tooltip/tooltip.module.css
```diff
@@ -1,5 +1,10 @@
+:root {
+    --reactist-tooltip-font-size: inherit;
```
doistbot, line 2: <a href="#">P2</a>  Defining the default on `:root` makes global overrides order-dependent: if an app sets `:root { --reactist-tooltip-font-size: ... }` before Reactist’s CSS loads, this later `inherit` declaration wins and the override stops working. To keep consumer overrides reliable, drop the `:root` default and use a fallback at the usage site instead, e.g. `font-size: var(--reactist-tooltip-font-size, inherit)`.

doistbot, line 2: <a href="#">P2</a>  The PR description mentions overriding this variable "in a scope", but because `AriakitTooltip` renders in a portal attached to `document.body` by default, the tooltip DOM node is not a descendant of the trigger's wrapper elements. Scoped CSS overrides on a parent container (e.g., `.my-wrapper { --reactist-tooltip-font-size: 14px; }`) will be bypassed by standard CSS inheritance.

Global overrides on `body` or `:root` will work normally. For scoped overrides, consumers must apply the scoped class directly to the tooltip via the `exceptionallySetClassName` prop. Consider adding a brief CSS comment here documenting this portal limitation so consumers understand how to properly scope the override.

doistbot (commented):
Thanks pawelgrimm for your contribution 😎 👊. The addition of the `--reactist-tooltip-font-size` custom property is a nice, lightweight way to make tooltip styling more flexible.

Few things worth tightening:
- Drop the `:root` default definition and instead use a fallback at the usage site (e.g., `font-size: var(--reactist-tooltip-font-size, inherit)`) to ensure global consumer overrides aren't broken by CSS load order.
- Add a brief CSS comment noting that because tooltips render in a portal, consumers attempting scoped overrides must apply their class directly via `exceptionallySetClassName` rather than targeting a parent wrapper.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1049&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0600e640-5a03-11f1-9321-44eb3e5d82f5%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0600e640-5a03-11f1-9321-44eb3e5d82f5&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tooltip/tooltip.module.css
pawelgrimm: This is an established convention — consumers are expected to load Reactist CSS first or with CSS Cascade Layers.

pawelgrimm: We already expect consumers to override this by setting `exceptionallySetClassName` on Tooltip or rendering a custom child.

PR #1052, opened 2026-05-28
feat: Add AvatarPair

pawelgrimm:
## Short description

Adds `<AvatarPair>` for two overlapping avatars.

## PR Checklist

## References

- [🎨 Product Library - Web / Avatar](https://www.figma.com/design/LYlWNzvhMDh907l07mPPQk/Product-Library---Web?node-id=3249-113742&t=HbeIwSxR0GGEN4ob-0)
- Stacks on https://github.com/Doist/reactist/pull/1051

## 📸 Demo

CleanShot 2026-05-27 at 22 57 28@2x

### DIFF
--- a/src/avatar/avatar-pair.tsx
+++ b/src/avatar/avatar-pair.tsx
@@ -7,7 +7,7 @@ import { polymorphicComponent } from '../utils/polymorphism'
 
 import { ROUNDED_AVATAR_RADIUS_BY_SIZE } from './utils'
 
-import styles from './avatar.module.css'
+import styles from './avatar-pair.module.css'
 
 import type { ObfuscatedClassName } from '../utils/common-types'
 import type { PolymorphicComponentProps } from '../utils/polymorphism'
--- a/src/avatar/avatar-pair.tsx
+++ b/src/avatar/avatar-pair.tsx
@@ -1,0 +1,153 @@
+import * as React from 'react'
+
+import classNames from 'classnames'
+
+import { Box } from '../box'
+import { polymorphicComponent } from '../utils/polymorphism'
+
+import { ROUNDED_AVATAR_RADIUS_BY_SIZE } from './utils'
+
+import styles from './avatar.module.css'
+
+import type { ObfuscatedClassName } from '../utils/common-types'
+import type { PolymorphicComponentProps } from '../utils/polymorphism'
+import type { AvatarShape, AvatarSize } from './utils'
+
+type AvatarPairStyle = React.CSSProperties & {
+    '--reactist-avatar-pair-size': string
+    '--reactist-avatar-pair-spacing': string
+    '--reactist-avatar-pair-mask': string
+    '--reactist-avatar-pair-rounded-radius': string
+    '--reactist-avatar-pair-rounded-mask-radius': string
+}
+
+const AVATAR_PAIR_MASK_BY_SIZE: Record<AvatarSize, string> = {
+    80: '3px',
+    72: '3px',
+    62: '3px',
+    50: '3px',
+    40: '3px',
+    36: '2.5px',
+    30: '2.5px',
+    28: '2px',
+    24: '2px',
+    20: '2px',
+    18: '1.5px',
+    16: '1.25px',
+    12: '1px',
+}
+
+const AVATAR_PAIR_SPACING_BY_SIZE: Record<AvatarSize, string> = {
+    80: '36px',
+    72: '32px',
+    62: '28px',
+    50: '22px',
+    40: '18px',
+    36: '16px',
+    30: '14px',
+    28: '12px',
+    24: '12px',
+    20: '10px',
+    18: '10px',
+    16: '8px',
+    12: '6px',
+}
+
+/**
+ * Props for the `AvatarPair` component.
+ */
+type AvatarPairOwnProps = ObfuscatedClassName & {
+    /**
+     * The rendered avatar size, in CSS pixels.
+     *
+     * Direct child Avatar components should use the same size.
+     */
+    size: AvatarSize
+
+    /**
+     * The paired avatar shape.
+     *
+     * Direct child Avatar components should use the same shape.
+     *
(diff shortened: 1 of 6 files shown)


src/avatar/avatar-pair.tsx
```diff
@@ -0,0 +1,153 @@
...
+     * @default 'circle'
+     */
+    shape?: AvatarShape
+
+    /**
+     * Exactly two paired Avatar children. The first child is the foreground
+     * avatar (positioned bottom-right); the second is the diagonal overlay
+     * (positioned top-left, masked where it overlaps the first).
+     */
+    children: readonly [React.ReactElement, React.ReactElement]
```
doistbot, line 39: <a href="#">P2</a>  This tuple only guarantees “two React elements”, not “two direct Avatar children”. A fragment or wrapper component still type-checks here and can render 3+ DOM children, but the CSS only positions `:first-child` and `:last-child`, so the extra nodes stack incorrectly. Narrow the child type to Avatar elements or add a runtime guard for non-direct children.

src/avatar/avatar-pair.tsx
```diff
@@ -0,0 +1,153 @@
...
+            shape = 'circle',
+            children,
+            exceptionallySetClassName,
+            'data-testid': testId,
+            ...restProps
+        },
+        ref,
+    ) {
+        return (
+            <Box
```
doistbot, line 77: <a href="#">P2</a>  When this renders with `as="button"`, it gets the HTML default `type="submit"`. Using `AvatarPair` inside a form will submit the form on click unless every caller remembers to override it. Default the root to `type="button"` when `as === 'button'`, while still letting an explicit `type` prop win.

src/avatar/avatar-pair.tsx
```diff
@@ -0,0 +1,153 @@
...
+    ComponentType,
+    AvatarPairOwnProps,
+    'omitClassName'
+>
+
+/**
+ * Displays two Avatar children with the second avatar positioned diagonally
+ * above-left of the first avatar.
+ */
+const AvatarPair = polymorphicComponent<'div', AvatarPairOwnProps, 'omitClassName'>(
```
doistbot, line 63: <a href="#">P2</a>  `polymorphism.ts` and `polymorphicComponent` are marked as deprecated in the repository guidelines (`CLAUDE.md`). New components should avoid this utility and instead use `React.forwardRef` with a named function matching the component name (using `asChild` or standard DOM wrappers if custom tags are needed).

doistbot (commented):
Thanks pawelgrimm for your contribution 😊. The new `AvatarGroup` and `AvatarPair` components bring robust, size-responsive mask styling and solid test coverage.

Few things worth tightening:
- **Component architecture**: Migrate away from the deprecated `polymorphicComponent` utility in favor of standard `React.forwardRef` patterns as per the project guidelines.
- **Child constraints**: Narrow the child types or add a runtime guard to prevent fragments or wrappers from injecting 3+ DOM nodes and breaking the CSS `:first-child`/`:last-child` selectors.
- **Button defaults**: Explicitly default to `type="button"` when the component is rendered `as="button"` to prevent accidental form submissions.

<details>

- **[P3] src/avatar/avatar-pair.tsx:20**: `--reactist-avatar-pair-rounded-radius` is defined here and populated in `getAvatarPairStyle`, but it is never consumed in `avatar.module.css` (unlike its counterpart in `AvatarGroup`). It can be safely removed from `AvatarPairStyle`, the return object below, and the `.avatarPair` block in the CSS file.
- **[P3] src/avatar/avatar-pair.tsx:24**: `AVATAR_PAIR_MASK_BY_SIZE` is a byte-for-byte copy of `AVATAR_GROUP_MASK_BY_SIZE` in `avatar-group.tsx`. That gives us two sources of truth for the same avatar geometry, so any future size/mask tweak has to be updated in both files or the overlap components will drift. Please hoist the shared mask lookup into `src/avatar/utils.ts` and reuse it here and in AvatarGroup.
- **[P3] src/avatar/avatar-pair.test.tsx:56**: This only asserts the rounded branch. Since the default circle variant uses a different shape class and is the main rendering path, add a small sibling test for the default render (`avatarPairShape-circle`) so a regression there doesn’t slip through unnoticed.
- **[P3] src/avatar/avatar-pair.test.tsx:67**: The API promise here is “exactly two” children, but this test only covers the “not one child” side. Add a second `@ts-expect-error` case with three avatars; otherwise a typing regression that allows extra children would still pass while breaking the pair layout selectors.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1052&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A53ac2570-5a45-11f1-9345-663d913cb39a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A53ac2570-5a45-11f1-9345-663d913cb39a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

PR #1058, opened 2026-06-04
fix: Restore previous Avatar implementation as DeprecatedAvatar

pawelgrimm:
## Short description

Restores the previous Avatar implementation as `DeprecatedAvatar` while keeping the revamped `Avatar` export in place.

The deprecated component keeps legacy props such as `user`, `avatarUrl`, `colorList`, and responsive string sizes with scoped CSS module styles, and now lives in `src/deprecated-avatar`.

Restores the legacy Avatar stories under `Deprecated Components/Avatar` and moves the deprecated Input story into the same section.

## PR Checklist


src/deprecated-avatar/deprecated-avatar.tsx
```diff
@@ -0,0 +1,90 @@
...
+    colorList?: string[]
+    size?: ResponsiveProp<DeprecatedAvatarSize>
+    avatarUrl?: string
+    user: { name?: string; email: string }
+}
+
+/**
+ * @deprecated Use `Avatar` instead.
+ */
+function DeprecatedAvatar({
```
doistbot, line 49: <big><kbd>🟠 P1</kbd></big>  The component does not use `React.forwardRef` or extend standard HTML attributes in its props type (e.g., `Omit<React.ComponentPropsWithoutRef<'div'>, 'style'>`). This violates project conventions, prevents passing standard attributes like `onClick` or `id` without TypeScript errors, and prevents consumers from attaching a `ref`.

Consider wrapping the component in `React.forwardRef` and extending the props interface. Omitting `style` from the allowed props is recommended to prevent consumers from accidentally overriding the background styles set below when spreading `...props`.

src/deprecated-avatar/deprecated-avatar.tsx
```diff
@@ -0,0 +1,90 @@
...
+              textIndent: '-9999px', // hide the initials
+          }
+        : {
+              backgroundColor: colorList[emailToIndex(user.email, colorList.length)],
+          }
+
+    const sizeClassName = getClassNames(styles, 'deprecated-size', avatarSize)
+
+    return (
+        <Box
```
doistbot, line 73: <big><kbd>🟡 P2</kbd></big>  This root doesn't use any of `Box`'s layout features, but `Box` still runs its full responsive/class-name resolution path on every render. That makes each `DeprecatedAvatar` do more work than needed. Consider rendering a plain `div` here (or a lightweight wrapper that only applies the base reset class) and building the class list locally.

src/deprecated-avatar/deprecated-avatar.test.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react'
+
+import { render, screen } from '@testing-library/react'
+
+import { DeprecatedAvatar } from './deprecated-avatar'
+
+describe('DeprecatedAvatar', () => {
```
doistbot, line 7: <big><kbd>🟡 P2</kbd></big>  This suite never exercises the restored `colorList` prop, even though preserving that legacy API is one of the goals of this PR. Please add a case that passes a custom list and asserts the resulting `backgroundColor`, otherwise a regression here will slip through while the initials-only tests still pass.

src/deprecated-avatar/deprecated-avatar.tsx
```diff
@@ -0,0 +1,90 @@
...
+    size = 'l',
+    className,
+    colorList = AVATAR_COLORS,
+    exceptionallySetClassName,
+    ...props
+}: DeprecatedAvatarProps) {
+    const userInitials = getInitials(user.name) || getInitials(user.email)
+    const avatarSize = size ? size : 'l'
+
+    const style = avatarUrl
```
doistbot, line 61: <big><kbd>🟡 P2</kbd></big>  When `avatarUrl` is present, this restores the old pattern of rendering the avatar photo as a CSS background on a generic `<div>`. That gives the image no semantic markup and no accessible name beyond the hidden initials, which conflicts with the component accessibility guidance in `CLAUDE.md` (https://github.com/Doist/reactist/blob/main/CLAUDE.md). Please render an actual `` here, or at minimum add `role="img"` plus an `aria-label` derived from `user` for the image case.

doistbot (commented):
Thanks Paweł for safely restoring the legacy Avatar component to ease the migration path 😄 🙌.

Few things worth tightening:
- **API conventions**: Wrap the component in `React.forwardRef` and extend standard HTML `div` props (omitting `style`) so consumers can attach refs and standard attributes.
- **Accessibility**: Provide semantic markup for the image by either using an actual `` tag or adding `role="img"` with a derived `aria-label`.
- **Performance**: Swap the root `Box` component for a plain `div` to avoid unnecessary responsive resolution overhead.
- **Test coverage**: Add a test exercising the restored `colorList` prop to ensure the legacy behavior is fully protected.

<details>

- **[P3] src/deprecated-avatar/deprecated-avatar.test.tsx:7**: The test suite is missing an accessibility test (`jest-axe`). Project guidelines mandate that every component should have an a11y test, even if this is a restored deprecated component.
- **[P3] src/deprecated-avatar/deprecated-avatar.stories.tsx:103**: Spreading `...args` directly passes the story's `userName` and `email` keys as props to `DeprecatedAvatar`, which spreads them onto `Box` and ultimately the DOM `div`. This will trigger a React "does not recognize prop on a DOM element" warning in the console. Consider destructuring these properties out before spreading the rest of the arguments.
- **[P3] src/deprecated-avatar/deprecated-avatar.tsx:67**: `colorList` is currently typed as any `string[]`, so `colorList={[]}` is a valid call. In that case `emailToIndex(user.email, colorList.length)` becomes `... % 0`, which returns `NaN` and leaves `backgroundColor` unset, so the avatar can render white initials on no background. Guard the empty-array case here, or fall back to `AVATAR_COLORS` when `colorList.length === 0`.
- **[P3] src/deprecated-avatar/deprecated-avatar.test.tsx:13**: `toMatchSnapshot()` is locking the entire rendered node here, including Box's class composition, for a behavior that's really just “uses `avatarUrl`”. A focused assertion on `backgroundImage`/`textIndent` would cover the contract with less churn when unrelated markup or class details change.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1058&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A0e664050-6048-11f1-8f79-733f1b732a96%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A0e664050-6048-11f1-8f79-733f1b732a96&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

pawelgrimm:
Doistbot made some good points, but we're not reworking the avatar here, we're just bringing back the previous implementation under the "Deprecated" namespace.

scottlovegrove (approved):

PR #1059, opened 2026-06-06
feat: add readOnlyVariant to TextField and TextArea

lmjabreu:
_No linked issue. Component-library follow-up from building read-only description views in todoist-web._

## Short description

`TextField` and `TextArea` hardcode a grey fill on their read-only state, with no way to opt out. Consumers that want a read-only field to read as plain text override the fill with their own CSS.

This adds a `readOnlyVariant` prop to both (also inherited by `PasswordField`):

- `filled` (default) keeps today's grey fill, so existing usage is unchanged.
- `plain` suppresses the fill, so a read-only field matches its editable appearance.

The prop only applies when `readOnly` is set. `TextArea`'s CSS moves from the `textarea[readonly]` selector to a `.readOnly` class so the fill can be gated on the variant; the default render is identical to before. The shared types live in `base-field`.

This is the first of three styling-only PRs for the read-only state (#1060 subdues the border and drops hover; #1061 fixes the bordered fill). None of them removes part of the component.

## On the two variants

Review asked whether we need two ([discussion](https://github.com/Doist/reactist/pull/1059#issuecomment-4685235318)). Long term, no: the goal is one read-only style for genuine inputs (a copyable token, a feed URL), with the no-fill case moving to a separate `EditableText` component for inline-editable display text.

`plain` stays for now because two todoist-web surfaces still render that way (`goal-edit-view`, `project-details-sidebar`). Both are being removed but exist in code today, so dropping the variant now would regress them. The prop replaces their CSS overrides in the meantime. Removing `plain` is a separate PR once those surfaces are gone.

## PR Checklist

### DIFF
--- a/src/text-area/text-area.test.tsx
+++ b/src/text-area/text-area.test.tsx
@@ -230,6 +230,27 @@ describe('TextArea', () => {
         expect(textareaElement).toHaveValue('')
     })
 
+    it('applies the read-only fill by default', () => {
+        render(<TextArea label="Tell us a bit about you" readOnly />)
+        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).toHaveClass(
+            'readOnly',
+        )
+    })
+
+    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
+        render(<TextArea label="Tell us a bit about you" readOnly readOnlyVariant="plain" />)
+        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).not.toHaveClass(
+            'readOnly',
+        )
+    })
+
+    it('does not apply the read-only fill to an editable field', () => {
+        render(<TextArea label="Tell us a bit about you" />)
+        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).not.toHaveClass(
+            'readOnly',
+        )
+    })
+
     it('can be a controlled field', async () => {
         function TestCase() {
             const [value, setValue] = React.useState('')
--- a/src/text-area/text-area.test.tsx
+++ b/src/text-area/text-area.test.tsx
@@ -237,11 +237,13 @@ describe('TextArea', () => {
         )
     })
 
-    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
+    it('omits the read-only fill when readOnlyVariant is "plain", but stays read-only', async () => {
         render(<TextArea label="Tell us a bit about you" readOnly readOnlyVariant="plain" />)
-        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).not.toHaveClass(
-            'readOnly',
-        )
+        const user = userEvent.setup()
+        const textareaElement = screen.getByRole('textbox', { name: 'Tell us a bit about you' })
+        expect(textareaElement).not.toHaveClass('readOnly')
+        await user.type(textareaElement, 'I love to travel around the world')
+        expect(textareaElement).toHaveValue('')
     })
 
     it('does not apply the read-only fill to an editable field', () => {
--- a/src/text-field/text-field.test.tsx
+++ b/src/text-field/text-field.test.tsx
@@ -196,12 +196,13 @@ describe('TextField', () => {
         expect(wrapper).toHaveClass('readOnly')
     })
 
-    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
+    it('omits the read-only fill when readOnlyVariant is "plain", but stays read-only', async () => {
         render(<TextField label="Whatʼs your job title?" readOnly readOnlyVariant="plain" />)
-        const wrapper = screen
-            .getByRole('textbox', { name: 'Whatʼs your job title?' })
-            .closest('.inputWrapper')
-        expect(wrapper).not.toHaveClass('readOnly')
+        const user = userEvent.setup()
+        const inputElement = screen.getByRole('textbox', { name: 'Whatʼs your job title?' })
+        expect(inputElement.closest('.inputWrapper')).not.toHaveClass('readOnly')
+        await user.type(inputElement, 'Software developer')
+        expect(inputElement).toHaveValue('')
     })
 
     it('does not apply the read-only fill to an editable field', () => {
--- a/src/text-field/text-field.test.tsx
+++ b/src/text-field/text-field.test.tsx
@@ -188,6 +188,30 @@ describe('TextField', () => {
         expect(inputElement).toHaveValue('')
     })
 
+    it('applies the read-only fill by default', () => {
+        render(<TextField label="Whatʼs your job title?" readOnly />)
+        const wrapper = screen
+            .getByRole('textbox', { name: 'Whatʼs your job title?' })
+            .closest('.inputWrapper')
+        expect(wrapper).toHaveClass('readOnly')
+    })
+
+    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
+        render(<TextField label="Whatʼs your job title?" readOnly readOnlyVariant="plain" />)
+        const wrapper = screen
+            .getByRole('textbox', { name: 'Whatʼs your job title?' })
+            .closest('.inputWrapper')
+        expect(wrapper).not.toHaveClass('readOnly')
+    })
+
+    it('does not apply the read-only fill to an editable field', () => {
+        render(<TextField label="Whatʼs your job title?" />)
+        const wrapper = screen
+            .getByRole('textbox', { name: 'Whatʼs your job title?' })
+            .closest('.inputWrapper')
+        expect(wrapper).not.toHaveClass('readOnly')
+    })
+
     it('can be a controlled input field', async () => {
         function TestCase() {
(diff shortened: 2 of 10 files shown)


lmjabreu:
@doistbot /review

doistbot (commented):
Thanks Luis for cleanly implementing the new `readOnlyVariant` prop to give us more visual flexibility with read-only fields 😎 👊.

Few things worth tightening:
- Ensure the new tests for both `TextField` and `TextArea` verify that the components remain functionally non-editable when `readOnlyVariant="plain"` is passed, rather than just asserting on the styling classes.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1059&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Adfe10bd0-61c7-11f1-979f-a589e41a74f6%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Adfe10bd0-61c7-11f1-979f-a589e41a74f6&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/text-field/text-field.test.tsx
```diff
@@ -188,6 +188,30 @@ describe('TextField', () => {
...
 
+    it('applies the read-only fill by default', () => {
+        render(<TextField label="Whatʼs your job title?" readOnly />)
+        const wrapper = screen
+            .getByRole('textbox', { name: 'Whatʼs your job title?' })
+            .closest('.inputWrapper')
+        expect(wrapper).toHaveClass('readOnly')
+    })
+
+    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This new `plain`-variant case only checks the wrapper class. It should also exercise the existing read-only behavior (typing doesn't change the value), otherwise a regression that drops the `readOnly` attribute specifically on `readOnlyVariant="plain"` would still pass all the new tests.

src/text-area/text-area.test.tsx
```diff
@@ -230,6 +230,27 @@ describe('TextArea', () => {
...
     })
 
+    it('applies the read-only fill by default', () => {
+        render(<TextArea label="Tell us a bit about you" readOnly />)
+        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).toHaveClass(
+            'readOnly',
+        )
+    })
+
+    it('omits the read-only fill when readOnlyVariant is "plain"', () => {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This covers the styling branch but not the functional one. Please also assert that `readOnlyVariant="plain"` still keeps the textarea non-editable, so a future bug in the new prop path can't pass while making the field editable.

src/text-field/text-field.test.tsx
lmjabreu: Done. The plain-variant test now also types into the field and asserts the value stays empty, so a regression that drops `readOnly` on the `readOnlyVariant="plain"` path would fail here. (76b7d80a)

src/text-area/text-area.test.tsx
lmjabreu: Done. Added a typing assertion to the plain-variant case so it fails if the textarea becomes editable on the new prop path. (76b7d80a)

doistbot (commented):
Thank you Luis for the clean, well-tested addition of the `readOnlyVariant` to our text components 🤗.

No issues were flagged during the review.

<details>

- **[P3] src/text-area/text-area.module.css:32**: For `TextArea`, switching the default filled style from `textarea[readonly]` to a `.readOnly` class duplicates the same state in JS and CSS. The filled case can stay anchored to the native `readonly` attribute, and `readOnlyVariant="plain"` can add a small opt-out class/attribute instead. That keeps the visual default tied to the semantic readonly state and removes one sync point to maintain.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1059&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ac65b1840-623a-11f1-9a63-b0931e7b5ca7%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ac65b1840-623a-11f1-9a63-b0931e7b5ca7&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

lmjabreu:
@engfragui tagging you on this one for review – can't recall which user/group you recommended on the last meeting 😅 Feel free to re-assign.

henningmu:
@lmjabreu see the linked handbook docs in the template: https://handbook.doist.com/doc/code-reviews-THNOSl1QID#h-15-assign-a-random-reviewer (it's `@doist/frontend-product` 🙌)

gnapse:
I'm still not sure if I understand why we need two variants for read-only mode. Why wouldn't we want consistency instead? That is, that read-only input fields always show consistently in the same way no matter what.

I think we should not have the gray background variant at all, even if it's the default today. Graying out the field is more appropriate for disabled fields. What do you think?

pawelgrimm:
Are we talking about these components?
CleanShot 2026-06-08 at 13 50 42@2x

They doesn't read like a variant of `TextField` and `TextArea` to me — they're more like a `Text` that turns into an edit form when you click on it. Perhaps the right solution is to develop a new component for that behavior and call it something like `EditableText` or `InlineEditableText` or something along those lines?

lmjabreu:
> why we need two variants for read-only mode

We don't! :D The plain variant only existed because of Goals and the old Project Descriptions sidebars (both being dropped, but they still exist in code today).

The use-case for the plain variant is better served by a separate component (EditableText like Paweł says).

FWIW, EditableText has **a lot** of use-cases across TD: project name, section name/description, task name/description, and so on. Basically any text that can be edited inline, but it's usually styled as display text (so.. `contenteditable` 😅)

Regarding the grey 'disabled' styling: agree, that's addressed in subsequent PRs.

TLDR: one read-only style for inputs, and one editable-text component for editable (display/UI) text.

gnapse:
@lmjabreu if I read your comment correctly, then we can close this PR without merging, right?

frankieyan:
While working on the component inventory, I came across the component for this use case, called exactly `EditableText`, but it's only used for the editable project titles.

- https://github.com/Doist/todoist-web/tree/85aab3a9064601eb0981f597fd426ef30d76c977/src/03-features/project-view/components/editable-text
- https://www.figma.com/design/LYlWNzvhMDh907l07mPPQk/Product-Library---Web?node-id=17009-653383&t=Tv1bTqOsZDQgpDdn-4

So if anything, we should get that into Reactist and get a story written up for it.

Bloomca:
I am a bit confused on the state of this PR/use-cases. Is it for something like frozen/unjoined projects? Does the Frankie's comment change anything?

lmjabreu:
Going to close all three (#1059, #1060, #1061). They're not pressing, we descriptions work doesn't depend on them anymore, and there's better ways forward.

@gnapse The plain variant was used by Goals and the **old** Project Descriptions so the text field in the sidebar would look like task description field (not bordered).

We do have `EditableText` but it doesn't support rich text, it's used for project names (plain text).

We could extend `EditableText` but that's separate to these PRs.

PR #1060, opened 2026-06-06
feat: subdue read-only field border and drop hover

lmjabreu:
_No linked issue. Follow-up to #1059, stacked on `feat/readonly-variant-fields` (base of this PR). Review #1059 first; this targets that branch and will retarget `main` once #1059 merges._

## Short description

Read-only `TextField` and `TextArea` reused the editable idle border and still brightened it on hover and focus. On a field you cannot edit, that hover/focus reaction reads as interactive. This refines the read-only treatment:

- **Subdued border.** Read-only fields use `--reactist-divider-tertiary` (one step lighter than the editable idle border) so a locked field is visually quieter.
- **No hover reaction.** The hover border-brightening is gated out for read-only fields.
- **Focus is kept.** Read-only fields remain keyboard-focusable (you can tab in to read and copy), so removing the focus indicator would break [WCAG 2.4.7 Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html). The focus border stays.

This applies to the read-only state regardless of `readOnlyVariant`; the variant only toggles the grey fill. It covers the default and the `bordered` variant of both components.

## Implementation notes

- The read-only fill class is split out as `readOnlyFilled`, so the read-only **state** (subdued border, no hover) and the **fill** (grey background) are independent. `plain` read-only fields get the subdued border without the fill.
- `BaseField` gains a `readOnly` prop. The bordered border is drawn by `BaseField`, so it needs the read-only signal to gate hover and apply the subdued color. The prop is optional and additive.
- Error tone still wins: the alert border uses `!important`, so a read-only field in an error state keeps its red border.

## Scope

This is intentionally separate from #1059. #1059 is purely additive (a new prop, no change to existing appearance). This PR changes the **default** read-only appearance of every field across the app, so it carries the visual diffs and is reviewed on its own.

## PR Checklist

### DIFF
--- a/src/text-area/text-area.test.tsx
+++ b/src/text-area/text-area.test.tsx
@@ -256,6 +256,14 @@ describe('TextArea', () => {
         expect(textareaElement).not.toHaveClass('readOnlyFilled')
     })
 
+    it('applies the read-only class to the container when variant is "bordered"', () => {
+        render(<TextArea label="Tell us a bit about you" readOnly variant="bordered" />)
+        const container = screen
+            .getByRole('textbox', { name: 'Tell us a bit about you' })
+            .closest('.textAreaContainer')
+        expect(container).toHaveClass('readOnly')
+    })
+
     it('can be a controlled field', async () => {
         function TestCase() {
             const [value, setValue] = React.useState('')
--- a/src/text-area/text-area.test.tsx
+++ b/src/text-area/text-area.test.tsx
@@ -230,27 +230,30 @@ describe('TextArea', () => {
         expect(textareaElement).toHaveValue('')
     })
 
-    it('applies the read-only fill by default', () => {
+    it('applies the read-only state and fill by default', () => {
         render(<TextArea label="Tell us a bit about you" readOnly />)
         expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).toHaveClass(
             'readOnly',
+            'readOnlyFilled',
         )
     })
 
-    it('omits the read-only fill when readOnlyVariant is "plain", but stays read-only', async () => {
+    it('omits the fill when readOnlyVariant is "plain", but stays read-only', async () => {
         render(<TextArea label="Tell us a bit about you" readOnly readOnlyVariant="plain" />)
         const user = userEvent.setup()
         const textareaElement = screen.getByRole('textbox', { name: 'Tell us a bit about you' })
-        expect(textareaElement).not.toHaveClass('readOnly')
+        // Plain keeps the read-only state (subdued border, no hover) but drops the grey fill
+        expect(textareaElement).toHaveClass('readOnly')
+        expect(textareaElement).not.toHaveClass('readOnlyFilled')
         await user.type(textareaElement, 'I love to travel around the world')
         expect(textareaElement).toHaveValue('')
     })
 
-    it('does not apply the read-only fill to an editable field', () => {
+    it('does not apply any read-only styling to an editable field', () => {
         render(<TextArea label="Tell us a bit about you" />)
-        expect(screen.getByRole('textbox', { name: 'Tell us a bit about you' })).not.toHaveClass(
-            'readOnly',
-        )
+        const textareaElement = screen.getByRole('textbox', { name: 'Tell us a bit about you' })
+        expect(textareaElement).not.toHaveClass('readOnly')
+        expect(textareaElement).not.toHaveClass('readOnlyFilled')
     })
 
     it('can be a controlled field', async () => {
--- a/src/text-field/text-field.test.tsx
+++ b/src/text-field/text-field.test.tsx
@@ -188,29 +188,33 @@ describe('TextField', () => {
         expect(inputElement).toHaveValue('')
     })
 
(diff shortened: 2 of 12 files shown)


lmjabreu:
@doistbot /review

src/text-field/text-field.test.tsx
```diff
@@ -188,29 +188,33 @@ describe('TextField', () => {
         expect(inputElement).toHaveValue('')
     })
 
-    it('applies the read-only fill by default', () => {
+    it('applies the read-only state and fill by default', () => {
```
doistbot, line 191: <big><kbd>🟡 P2</kbd></big>  This only verifies classes on `.inputWrapper`. The bordered read-only behavior added in this PR lives on the outer `BaseField` container, so a broken `readOnly` forward to `BaseField` would still pass. Add a `variant="bordered"` case and assert the bordered container gets the read-only class.

src/text-area/text-area.test.tsx
```diff
@@ -230,27 +230,30 @@ describe('TextArea', () => {
         expect(textareaElement).toHaveValue('')
     })
 
-    it('applies the read-only fill by default', () => {
+    it('applies the read-only state and fill by default', () => {
```
doistbot, line 233: <big><kbd>🟡 P2</kbd></big>  This only verifies classes on the `<textarea>` itself. The new bordered behavior here is driven by the `.textAreaContainer.readOnly` class, so a regression in the bordered path would still pass. Add a `variant="bordered"` read-only case and assert the container carries `readOnly`.

doistbot (commented):
Thanks for refining the read-only visual behaviors and cleanly separating the state from the fill variant, Luis 🤗.

Few things worth tightening:
- Expand the tests for both `TextField` and `TextArea` to include a `variant="bordered"` case, ensuring the outer containers correctly receive the `readOnly` class rather than only checking the inner input elements.

<details>

- **[P3] src/text-area/text-area.tsx:112**: Since `BaseField` receives `readOnly` on line 107 and natively handles the subdued border and hover gating for the `.bordered` wrapper (via `base-field.module.css`), passing `styles.readOnly` here is redundant. You can remove this line, and simplify `text-area.module.css` (lines 54-60) to only target the `:not(.bordered)` variant, exactly like `text-field` does: ```css .textAreaContainer:not(.bordered) textarea.readOnly { border-color: var(--reactist-divider-tertiary); } .textAreaContainer:not(.bordered) textarea:not(.readOnly):hover { border-color: var(--reactist-inputs-hover); } ```
- **[P3] src/text-field/text-field.mdx:94**: This new note now conflicts with the sentence just above it that says `readOnlyVariant="plain"` makes the field look identical to its editable state. After this change, plain read-only fields still get the subdued border/no-hover treatment, so please update that earlier sentence here (and the matching copy in `text-area.mdx`) to keep the docs consistent.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1060&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ada1b3cb0-61e1-11f1-9f09-e9ea69a7c32a%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ada1b3cb0-61e1-11f1-9f09-e9ea69a7c32a&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/text-field/text-field.test.tsx
lmjabreu: Done. Added a `variant="bordered"` read-only case that walks up to `.closest('.bordered')` and asserts the container carries `readOnly` — so a broken `readOnly` forward to `BaseField` would now fail.

src/text-area/text-area.test.tsx
lmjabreu: Done. Added a `variant="bordered"` read-only case that walks up to `.closest('.textAreaContainer')` and asserts the container carries `readOnly` — so a regression in the bordered path would now fail.

lmjabreu:
Closing alongside #1059. Not using this on Project Descriptions anymore, see discussion in #1059.

PR #1061, opened 2026-06-07
fix: fill the bordered container for read-only filled fields

lmjabreu:
_No linked issue. Follow-up to #1060, stacked on `feat/readonly-field-subdued-border` (base of this PR). Review #1059 and #1060 first; this targets #1060's branch and will retarget `main` once those merge._

## Short description

In the `bordered` variant, the container owns the border and padding while the field sits inside it. A read-only `filled` field put its grey on the inner field, so the fill showed up as an inset band that did not reach the border:

Before: grey band floating inside a white padded box.
After: the whole bordered box is filled, so a read-only field reads as one cohesive grey card.

The fill now lives on the container for the bordered variant. The default variant is unchanged, since its field already fills edge to edge.

## Implementation notes

- `BaseField` gains an internal `readOnlyFilled` signal (not part of the public field props). The field components pass `readOnly && readOnlyVariant === 'filled'`.
- The container fill is scoped to `.container.bordered.readOnlyFilled`, so the default variant's container is never filled (only its inner field is, as before).
- `plain` bordered read-only fields are unaffected: subdued border, no fill.

## Scope

Separate from #1060 on purpose. #1060 refines the read-only border and hover for every field; this is a narrow appearance fix for one variant combination (bordered + filled + read-only).

## PR Checklist

### DIFF
--- a/src/base-field/base-field.module.css
+++ b/src/base-field/base-field.module.css
@@ -41,6 +41,12 @@
     border-color: var(--reactist-divider-tertiary);
 }
 
+/* In the bordered variant the container owns the border and padding, so a filled read-only field
+   is filled here (rather than on the inner field) to reach the border edge to edge */
+.container.bordered.readOnlyFilled {
+    background-color: var(--reactist-field-readonly-background);
+}
+
 .container.bordered:not(.readOnly):hover {
     border-color: var(--reactist-inputs-hover) !important;
 }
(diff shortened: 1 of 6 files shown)


lmjabreu:
Closing alongside #1059. Not using this on Project Descriptions anymore, see discussion in #1059.

PR #1062, opened 2026-06-10
fix(menu): Keep persistent elements interactive while a modal menu is open

engfragui:
⚠️ JUST AN EXPLORATION

## Summary

A modal `MenuList` marks the element tree outside it as `inert`, which removes those elements from hit-testing. On desktop apps (e.g. Electron) this breaks the native window drag region (`-webkit-app-region: drag`) of the title bar, since `inert` elements no longer receive the OS drag gesture.

This change defaults `MenuList`'s `getPersistentElements` to exempt any element marked with `data-reactist-persist` from the inert barrier, keeping it interactive while the menu stays modal in every other respect. A consumer-provided `getPersistentElements` is still honored and merged with the default.

## Reference 

- Helps with https://github.com/Doist/Issues/issues/20435
- Used in https://github.com/Doist/todoist-web/pull/18162

## Usage

Mark the element that should stay interactive (e.g. the draggable title bar):

```html
<div data-reactist-persist style="-webkit-app-region: drag" />
```

### DIFF
--- a/src/menu/menu.tsx
+++ b/src/menu/menu.tsx
@@ -149,10 +149,28 @@ interface MenuListProps
         ObfuscatedClassName {}
 
 /**
+ * Selector for elements that should remain interactive while a modal menu is open.
+ *
+ * A modal `MenuList` marks the entire element tree outside of it as `inert`, which removes those
+ * elements from hit-testing. On desktop apps (e.g. Electron) this breaks the native window
+ * drag region (`-webkit-app-region: drag`) of the title bar, since `inert` elements no longer
+ * receive the OS drag gesture. Marking such elements with `data-reactist-persist` keeps them
+ * interactive while the menu stays modal in every other respect.
+ */
+const PERSISTENT_ELEMENTS_SELECTOR = '[data-reactist-persist]'
+
+function getPersistentElementsDefault(): Element[] {
+    if (typeof document === 'undefined') {
+        return []
+    }
+    return Array.from(document.querySelectorAll(PERSISTENT_ELEMENTS_SELECTOR))
+}
+
+/**
  * The dropdown menu itself, containing a list of menu items.
  */
 const MenuList = React.forwardRef<HTMLDivElement, MenuListProps>(function MenuList(
-    { exceptionallySetClassName, modal = true, flip, ...props },
+    { exceptionallySetClassName, modal = true, flip, getPersistentElements, ...props },
     ref,
 ) {
     const { menuStore, getAnchorRect } = React.useContext(MenuContext)
@@ -163,6 +181,16 @@ const MenuList = React.forwardRef<HTMLDivElement, MenuListProps>(function MenuLi
     const { isSubMenu } = React.useContext(SubMenuContext)
 
     const isOpen = menuStore.useState('open')
+
+    const mergedGetPersistentElements = React.useCallback(
+        function mergedGetPersistentElements(): Element[] {
+            const consumerElements = getPersistentElements
+                ? Array.from(getPersistentElements())
+                : []
+            return [...getPersistentElementsDefault(), ...consumerElements]
+        },
+        [getPersistentElements],
+    )
 
     return isOpen ? (
         <Portal preserveTabOrder>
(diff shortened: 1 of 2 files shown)


src/menu/menu.tsx
```diff
@@ -164,6 +182,16 @@ const MenuList = React.forwardRef<HTMLDivElement, MenuListProps>(function MenuLi
 
     const isOpen = menuStore.useState('open')
 
+    const mergedGetPersistentElements = React.useCallback(
+        function mergedGetPersistentElements(): Element[] {
+            const consumerElements = getPersistentElements
+                ? Array.from(getPersistentElements())
+                : []
+            return [...getPersistentElementsDefault(), ...consumerElements]
```
doistbot, line 190: <big><kbd>🟡 P2</kbd></big>  `MenuListProps` inherits Ariakit’s `getPersistentElements`, so callers will expect this prop to define the complete persistent-element set. Merging the default selector here turns it into an append-only hook: once an app adds `data-reactist-persist` anywhere, every `MenuList` will keep those elements interactive and a caller can’t opt back into strict modal behavior for a specific menu. That’s a surprising contract change for a pass-through Ariakit prop. Consider applying the default only when `getPersistentElements` is unset, or introducing a separate Reactist prop for the title-bar escape hatch.

doistbot, line 190: <big><kbd>🟡 P2</kbd></big>  This makes every modal `MenuList` pay for `document.querySelectorAll('[data-reactist-persist]')` whenever Ariakit resolves persistent elements, even in apps that never use this feature. Since menus are a repeated interaction path, walking the whole document and allocating fresh arrays here is avoidable overhead. Please gate the default lookup behind an explicit opt-in or cache the selector results for the lifetime of the open menu instead of rescanning the full DOM on each callback.

doistbot (commented):
This PR enables elements like native OS title bars to remain interactive while a modal `MenuList` is open by exempting them from the inert barrier.

Few things worth tightening:
* Merging the default selector changes the Ariakit `getPersistentElements` contract to be append-only, preventing callers from opting back into strict modal behavior—consider applying the default only when unset or providing a custom Reactist prop.
* Running a document-wide `querySelectorAll` whenever persistent elements resolve adds unnecessary overhead to a common interaction path—this scan should ideally be cached or gated behind an explicit opt-in.

<details>

- **[P3] src/menu/menu.test.tsx:429**: The implementation successfully merges `getPersistentElements` with `getPersistentElementsDefault()`. However, this test doesn't verify the merge because it lacks a `data-reactist-persist` element. If the code mistakenly overwrote the defaults instead of merging them, this test would still pass. Consider adding a `<div data-testid="default-persist" data-reactist-persist />` to this DOM and asserting that it also avoids the `aria-hidden` attribute.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1062&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab73acb40-64c3-11f1-98bc-0d461fdd133b%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab73acb40-64c3-11f1-98bc-0d461fdd133b&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

PR #1063, opened 2026-06-11
feat(prose): add table styling support

amix:
Follow-up to https://github.com/Doist/comms-web/pull/395#discussion_r3395143218 — Comms now renders GFM tables in messages and should eventually use `<Prose>` instead of its forked prose CSS, but `<Prose>` had no table styling.

## Short description

- Adds bordered table styles to `<Prose>`, with a new public `--reactist-prose-table-border` custom property (defaults to `var(--reactist-divider-primary)`).
- Wide tables scroll horizontally (`display: block; width: max-content; max-width: 100%; overflow: auto`) instead of stretching the surrounding content, with `word-break: normal` so cell text doesn't collapse to one character per line under `.prose`'s `break-word`.
- Headers are bold and start-aligned, with a `th:not([align])` guard so explicit column alignment survives both renderer conventions: `marked` emits the presentational `align` attribute (which author CSS would override without the guard), while react-markdown emits inline `text-align` styles (which win regardless).
- Cells reset first/last-child block margins so raw HTML tables with block content (e.g. `<td><p>…</p></td>`) don't get awkward vertical spacing.
- Adds `--reactist-prose-table-header-fill`, `--reactist-prose-table-row-odd-fill`, and `--reactist-prose-table-row-even-fill` custom properties (all default to `transparent`) so consumers can color the header row and stripe body rows, demoed in a "Table colors" story.
- Adds a table (with column alignment) to the Storybook prose example, plus a dedicated width-constrained "Table overflow" story so Chromatic exercises the horizontal-scroll layout at any viewport.

Smoke tested in Storybook in light and dark playground modes and in the overflow story.

## Out of scope

- Keyboard-focusable scrolling for clipped tables in Safari: Prose is CSS-only over HTML it doesn't control, so it can't inject a focusable wrapper or `tabindex`. Chrome 130+/Firefox handle this natively for scrollers without focusable children; the wrapper belongs at the consumer's renderer level (tracked for the Comms migration). See the resolved review thread for details.

## PR Checklist

- [x] Added tests for bugs / new features (visual coverage via the new Chromatic stories; no unit-testable logic)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/prose/prose-example.ts
+++ b/src/prose/prose-example.ts
@@ -74,6 +74,14 @@ Now a nested list:
 
     Do not bump wooden spoon or it will fall.
 
+And here's a table, with column alignment:
+
+| Item    | Quantity | Price |
+| ------- | :------: | ----: |
+| Carrots |        3 | $1.20 |
+| Celery  |        1 | $0.95 |
+| Lentils |      500g | $2.40 |
+
 ---
 
 A horizontal rule follows.
--- a/src/prose/prose-example.ts
+++ b/src/prose/prose-example.ts
@@ -90,3 +90,13 @@ A horizontal rule follows.
 
 The End
 `)
+
+export const proseTableOverflowExample = marked(`
+Tables whose content cannot wrap any further (e.g. long unbreakable links) scroll horizontally
+instead of stretching the surrounding content:
+
+| Resource | Reference |
+| -------- | --------- |
+| Prose styles | https://github.com/Doist/reactist/blob/main/src/prose/prose.module.css |
+| Prose stories | https://github.com/Doist/reactist/blob/main/src/prose/prose.stories.tsx |
+`)
--- a/src/prose/prose-example.ts
+++ b/src/prose/prose-example.ts
@@ -100,3 +100,17 @@ instead of stretching the surrounding content:
 | Prose styles | https://github.com/Doist/reactist/blob/main/src/prose/prose.module.css |
 | Prose stories | https://github.com/Doist/reactist/blob/main/src/prose/prose.stories.tsx |
 `)
+
+export const proseTableColorsExample = marked(`
+Consumers can give the header row and odd/even body rows their own background colors via the
+\`--reactist-prose-table-header-fill\`, \`--reactist-prose-table-row-odd-fill\`, and
+\`--reactist-prose-table-row-even-fill\` custom properties (all default to \`transparent\`):
+
+| Item    | Quantity | Price |
+| ------- | :------: | ----: |
+| Carrots |        3 | $1.20 |
+| Celery  |        1 | $0.95 |
+| Lentils |     500g | $2.40 |
+| Onions  |        2 | $0.80 |
+| Stock   |       1L | $3.10 |
+`)
--- a/src/prose/prose.module.css
+++ b/src/prose/prose.module.css
@@ -12,6 +12,13 @@
     --reactist-prose-horizontal-rule-color: var(--reactist-divider-primary);
 
     --reactist-prose-table-border: var(--reactist-divider-primary);
+
+    /* `transparent` (the initial value of `background-color`) rather than `initial`, which is
+       special-cased in custom property values to mean "guaranteed invalid", or `inherit`, which
+       backgrounds achieve through transparency rather than inheritance. */
+    --reactist-prose-table-header-fill: transparent;
+    --reactist-prose-table-row-odd-fill: transparent;
+    --reactist-prose-table-row-even-fill: transparent;
 }
 
 /* Internals for spacing. These are not considered public API. */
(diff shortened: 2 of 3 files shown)


doistbot (commented):
This PR introduces robust table styling support to the Prose component, including borders, explicit alignments, and horizontal scrolling for wide tables.

Few things worth tightening:
* Wrapping the table in a focusable overflow container to ensure horizontal scrolling is accessible to keyboard users.
* Adding a wider table to the Storybook fixture so Chromatic exercises the new overflow layout path.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1063&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A693649d0-6584-11f1-8e34-92367412be08%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A693649d0-6584-11f1-8e34-92367412be08&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/prose/prose.module.css
```diff
@@ -212,6 +214,52 @@
...
 }
 
+/* tables */
+
+.prose table {
+    /* Let wide tables scroll horizontally instead of stretching the surrounding content. */
+    display: block;
+    width: max-content;
+    max-width: 100%;
+    overflow: auto;
```
doistbot, line 231: <big><kbd>🟡 P2</kbd></big>  Making the `<table>` itself the scroll container leaves wide text-only tables inaccessible to keyboard users: once the page no longer overflows horizontally, there’s no focusable element they can use to scroll to the hidden columns. To keep clipped columns reachable, put the horizontal overflow on a focusable wrapper (or otherwise make the scroller focusable) instead of the table element.

src/prose/prose-example.ts
```diff
@@ -74,6 +74,14 @@ Now a nested list:
 
     Do not bump wooden spoon or it will fall.
 
+And here's a table, with column alignment:
+
+| Item    | Quantity | Price |
```
doistbot, line 79: <big><kbd>🟡 P2</kbd></big>  This new fixture only renders a narrow table, so the Chromatic coverage added by this PR won't catch regressions in the new overflow behavior (`display: block`, `width: max-content`, `overflow: auto`, `word-break: normal`). Please make one table wide enough to overflow, or add a dedicated wide-table example, so the main regression path is actually exercised.

amix: Posted by Claude Code on behalf of Amir:

Fixed in b723fe8e. Widening the playground table turned out to be viewport-dependent: cells wrap at spaces/hyphens under `max-width: 100%`, so the table only scrolls when its min-content width exceeds the capture width — easy to silently lose if Chromatic's viewport changes. Added a dedicated "Table overflow" story instead, which renders a table with unbreakable links inside a 400px container, guaranteeing the scroll layout is snapshotted at any viewport (verified locally: 587px content in a 352px scroller).

src/prose/prose.module.css
amix: Posted by Claude Code on behalf of Amir:

Valid concern, but it can't be solved in Prose itself: Prose is a CSS-only layer over HTML it doesn't control (`children` / `dangerouslySetInnerHTML`), so it has no way to inject a focusable wrapper or set `tabindex` on the table.

Two mitigations apply:

- Chrome 130+ and Firefox make scroll containers without focusable children keyboard-focusable natively, which covers exactly the text-only-table case described here. Safari is the remaining gap ([WebKit bug 277290](https://bugs.webkit.org/show_bug.cgi?id=277290), no implementation signal yet). Tables containing links are reachable everywhere, since tabbing through cell links scrolls clipped columns into view.
- The focusable wrapper belongs at the renderer level, where the HTML is generated — e.g. a `table` component mapping in react-markdown (Comms) or a renderer override in `marked` (Todoist), same as GitHub's `<markdown-accessibility-table>` approach. I'll carry this into the planned Comms migration to `<Prose>`.

This also matches what Comms ships today (the table CSS this ports from Doist/comms-web#395), so it's not a regression there. Resolving as a documented deferral.

rfgamaral (approved):
Looks good to me 🚀

But here's one suggestion: add more `--reactist-prose-*` variables that allow consumers to configure the background color of the header row, and the background color of odd and even rows. We can probably leave the Reactist defaults as `transparent`/`inherit`/`initial` (ask your clanker which one is most suited).

amix:
@rfgamaral great suggestion, I've implemented it. Could you please re-check 🙏

rfgamaral:
Looks great, merged it for you, @amix. We should get an automated release of this very soon.

PR #1074, opened 2026-06-16
feat: support React 19

frankieyan:
## Short description

This PR adds React 19 support. Storybook/Chromatic now runs v19 but tests and type checks will be run using both 18 and 19.

We attempted to run Storybook in v18 in parallel as well, but no luck so far.

## Test plan

- Introduce a React 19 only API, such as [`useActionState`](https://react.dev/reference/react/useActionState)
- [ ] It passes the normal `type-check` npm script
- [ ] It fails the `type-check:react18` npm script

## PR Checklist


src/menu/menu.test.tsx
```diff
@@ -286,8 +286,10 @@ describe('Menu', () => {
             expect(screen.getByRole('menuitem', { name: 'Save' })).toBeVisible()
         })
 
+        // Select via keyboard: in React 19 the hover-opened submenu closes synchronously when the pointer moves onto an item, unmounting it before a click can land.
```
doistbot: <big><kbd>🟠 P1</kbd></big>  This change papers over a real React 19 regression instead of covering it. The new comment says a hover-opened submenu closes as soon as the pointer moves onto an item, which means mouse users can't click submenu actions like "Save" anymore. Since this PR expands support to React 19, please fix that interaction and keep this test exercising the click path rather than switching it to keyboard-only selection.

doistbot (commented):
This PR updates feat: support React 19.

Few things worth tightening:
- See the inline comments for the issues found in the review.

<details>

- **[P3] package.json:41**: This changes the default `validate` path from one Jest run to two full runs. Because `setup` shells through `validate`, fresh local installs now pay that extra pass too. If the React 18 suite only needs broad coverage in CI, move it to a dedicated full/CI script instead of doubling the common local check.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1074&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A38fec5b0-695a-11f1-93c9-6f2f7b4bbc57%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A38fec5b0-695a-11f1-93c9-6f2f7b4bbc57&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

package.json
```diff
@@ -38,7 +38,7 @@
     "scripts": {
         "postinstall": "patch-package",
         "setup": "npm install && npm run validate",
-        "validate": "npm run lint && npm run type-check && npm run test",
+        "validate": "npm run lint && npm run type-check && npm run test && npm run test:react18",
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This adds a second runtime test pass, but `validate` still runs only one `type-check` against the React 19 typings from `devDependencies`. Because this package publishes declarations and now claims React 18 + 19 support, we should also add a React 18 typings/declaration check here; otherwise React 18-only type regressions can slip through CI.

src/utils/polymorphism.ts
```diff
@@ -190,11 +198,9 @@ function polymorphicComponent<
     OwnProps = EmptyObject,
     ShouldObfuscateClassName extends ObfuscateClassNameMode = 'obfuscateClassName',
 >(render: ForwardRefFunction<ComponentType, OwnProps, ShouldObfuscateClassName>) {
-    return React.forwardRef(render) as PolymorphicComponent<
-        ComponentType,
-        OwnProps,
-        ShouldObfuscateClassName
-    >
+    return React.forwardRef(
+        render as PolymorphicRenderFunction<ComponentType, OwnProps, ShouldObfuscateClassName>,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This cast papers over a real contract mismatch in the polymorphic helper. `PolymorphicComponentProps` is still built from `React.ComponentProps`, which includes `ref` under React 19, but a `forwardRef` render never receives `props.ref`. With the cast in place, future components can type-check while depending on a prop React strips at runtime. Please move the `PropsWithoutRef`/`ForwardedRef` change into the helper types themselves instead of masking it here.

src/components/icons/ThreeDotsIcon.svg.tsx
```diff
@@ -11,9 +9,5 @@ const ThreeDotsIcon = ({ color = '#A5A5A5' }) => (
         </g>
     </svg>
 )
-ThreeDotsIcon.propTypes = {
```
doistbot, line 14: <big><kbd>🟡 P2</kbd></big>  Removing `propTypes` here also removes the only prop contract this TSX component had. Since the function parameter is still untyped, `color` falls back to `any`, so consumers lose both runtime validation and compile-time checking. Please add a typed props annotation/interface before dropping `propTypes`.

.github/workflows/check-ci-validation.yml
```diff
@@ -129,10 +129,14 @@ jobs:
               run: |
                   npm ci
 
-            - name: Test codebase correctness
+            - name: Test codebase correctness (React 19)
               run: |
                   npm run test
 
+            - name: Test codebase correctness (React 18)
```
doistbot, line 140: <big><kbd>🟡 P2</kbd></big>  This adds a second full Jest pass in the same job, so every PR now waits for the React 19 and React 18 suites serially. These runs are independent; a matrix or separate jobs would keep wall-clock time close to one test run while still checking both runtimes.

package.json
frankieyan: The hope is that we don't keep React 18 support for very long 🤞Both Todoist and Twist have plans in place to migrate, and Automations is already on 19. CI is also reasonably fast as we have a small test suite (1-2min), so I wouldn't want to complicate the set up for now.

.github/workflows/check-ci-validation.yml
frankieyan: Same as https://github.com/Doist/reactist/pull/1074#discussion_r3423113924

src/menu/menu.test.tsx
frankieyan: This only happens because jsdom doesn't provide proper DOMRects, and Ariakit triggers a hide because of it: https://github.com/ariakit/ariakit/blob/200aa56676b8373fe10ec6c5fac61003ae82c1de/packages/ariakit-react-core/src/hovercard/hovercard.tsx#L201-L222. In React 18, the test had a short window for the click event to be handled, and with React 19, this queued render happens earlier, eliminating this window. 

Claude found https://github.com/facebook/react/pull/26512 after bisecting between different React commits and running this test, and this new behaviour is more correct.to me  We could mock the DOMRects for Ariakit to pick up, but that feels like it's testing too much into Ariakit's internals, and I'd rather stick with the keyboard flow.

src/components/icons/ThreeDotsIcon.svg.tsx
frankieyan: ✅ [a79132d](https://github.com/Doist/reactist/pull/1074/commits/a79132d149a398b4bc254c26523a2e3249eb2516)

src/utils/polymorphism.ts
frankieyan: ✅ [06f5c40](https://github.com/Doist/reactist/pull/1074/commits/06f5c402fd2f21c14284268f15df7a2ba52b0d7c)

package.json
frankieyan: Additional type check: [17c8ba2](https://github.com/Doist/reactist/pull/1074/commits/17c8ba2c03c29c7c2af7447b1440ff83f5fbeb78)

.github/workflows/check-ci-validation.yml
frankieyan: Still around a minute for the two checks: 

image

tsconfig.react18.json
```diff
@@ -0,0 +1,17 @@
+{
+    "extends": "./tsconfig.json",
+    "compilerOptions": {
+        "paths": {
+            "@": ["./"],
+            "@storybook/react": ["node_modules/@storybook/react/dist/index.d.ts"],
+            "@storybook/react-vite": ["node_modules/@storybook/react-vite/dist/index.d.ts"],
+            "storybook/actions": ["node_modules/storybook/dist/actions/index.d.ts"],
+            "storybook/test": ["node_modules/storybook/dist/test/index.d.ts"],
+            "react": ["node_modules/@types/react-18"],
```
doistbot, line 10: <big><kbd>🟡 P2</kbd></big>  This only remaps module imports. Because this repo now installs both `@types/react` 19 and `@types/react-18`, the React 18 compile can still pick up the 19 ambient declarations unless this config also narrows `types`/`typeRoots`. In that mixed setup, `npm run type-check:react18` can still let React 19-only type usage slip through. Please isolate the React 18 config from the default `@types/react*` packages instead of relying on `paths` alone.

src/tooltip/tooltip.tsx
```diff
@@ -144,13 +145,15 @@ const Tooltip = React.forwardRef<TooltipStore, TooltipProps>(
         }
 
         /* eslint-disable react-hooks/refs */
-        if (typeof child.ref === 'string') {
-            throw new Error('Tooltip: String refs cannot be used as they cannot be forwarded')
-        }
+        const childRef =
```
doistbot: <big><kbd>🟡 P2</kbd></big>  This ref fallback removed the previous fast-fail for string refs. React 18's public ref contract still includes `LegacyRef` strings, so JS or class-component consumers can now pass a string ref and we'll forward it into `TooltipAnchor` instead of throwing the clear unsupported-usage error we had before. Please keep the string-ref rejection after resolving `childRef`.

doistbot (commented):
This PR adds React 19 support across the library, with Storybook/Chromatic running on v19 and tests/type-checks running against both 18 and 19.

Few things worth tightening:
- The React 18 type-check config only remaps module paths via `paths`, but since both `@types/react` 19 and 18 are installed, ambient React 19 declarations can still leak in. Narrowing `types`/`typeRoots` (or otherwise isolating the packages) would make the React 18 check actually catch v19-only type usage.
- The Tooltip ref fallback removed the previous fast-fail for string refs. React 18 still allows `LegacyRef` strings, so JS or class-component consumers could now pass a string ref that gets silently forwarded into `TooltipAnchor` instead of getting a clear unsupported-usage error — worth re-adding the guard after resolving `childRef`.
- Expanding `validate` (and by extension `setup`) to run both test and type-check matrices serially doubles the cost of every fresh install and routine local check. Consider keeping the dual-matrix run in a dedicated CI script rather than the default local path.

<details>

- **[P3] package.json:41**: `validate` is the project's default local verification path, and `setup` shells through it. Expanding it to `type-check + type-check:react18 + test + test:react18` means every fresh install and routine local validation now pays two extra full passes. If the React 18 matrix is mainly for compatibility coverage, keep it in a dedicated CI/full-validation script instead of doubling the common local path.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1074&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A66005fb0-6a04-11f1-9433-bcf6c9a565c0%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A66005fb0-6a04-11f1-9433-bcf6c9a565c0&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tooltip/tooltip.tsx
frankieyan: ✅ [f33bb87](https://github.com/Doist/reactist/pull/1074/commits/f33bb87e6ffc728542f81c6c348300b719d4d993)

tsconfig.react18.json
frankieyan: React's types should be module scoped so I don't think this will be worth it. As it is, it should warn us when we are using React 19-only APIs:

image

rfgamaral (approved):
Looks good to me 🚀

PR #1075, opened 2026-06-16
feat: add ExpansionPanel component

benbreckler:
## Overview

The aim is to introduce a **new expansion panel component in Reactist** so we can expand and collapse sidebar sections in Comms and later implement the same component in Todoist. Demo [here](https://supercut.ai/share/doist/J7KprVnie1P4KjpEHQc1bR). 

This is related to the exploratory Expand Collapse PR [here](https://github.com/Doist/comms-web/pull/423). Once we have shipped the Reactist component, I can go back and update the Comms sidebar PR to use the Reactist component. 

## Preview

image


src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+        initiallyExpanded !== undefined &&
+        (props.isExpanded !== undefined || props.onToggleExpand !== undefined)
+    ) {
+        // eslint-disable-next-line no-console
+        console.warn('[ExpansionPanel]: cannot use initiallyExpanded in controlled mode')
+        initiallyExpanded = undefined
+    }
+
+    const [isExpanded, onToggleExpand] = useControlledState(
+        props.isExpanded || initiallyExpanded,
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Using `||` causes `isExpanded={false}` to fall through to `initiallyExpanded` (which is `undefined` in controlled mode). This only works correctly right now because the internal `useControlledState` hook happens to use `false` as its default parameter. Change this to `props.isExpanded ?? initiallyExpanded` to safely preserve the explicit `false` state.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+/**
+ * Button that toggles the panel's expand/collapse state. Wires up
+ * `aria-expanded` and `aria-controls` automatically from the parent
+ * `ExpansionPanel`'s context.
+ *
+ * Renders an `IconButton` when no children are passed, or a full-width `Button`
+ * with a chevron start icon otherwise. Remaining props forward to the underlying
+ * button.
+ */
+function ExpansionPanelToggle({
```
doistbot, line 178: <big><kbd>🟡 P2</kbd></big>  Repository guidelines require wrapping components in `React.forwardRef`. This is critical for `ExpansionPanelToggle` so consumers can attach a `ref` for focus management, tooltips, or popovers. `ExpansionPanelHeader` and `ExpansionPanelContent` should also forward refs to their underlying `Box` components.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+        )
+    }
+
+    return (
+        <Button
+            variant="quaternary"
+            width="full"
+            align="start"
+            {...props}
+            onClick={onToggleExpand}
```
doistbot, line 212: <big><kbd>🟡 P2</kbd></big>  Explicit props like `onClick` and `startIcon` are placed after `{...props}`, which silently overwrites any values provided by the consumer. Either omit these from `ExpansionPanelToggleProps` (as you did for `icon` and `size`) so they cause a type error, or compose them (e.g., chain `onClick` handlers).

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+            <AnimatedExpansionPanelContent isExpanded={isExpanded} onEntered={onEntered}>
+                <Box {...props} id={id}>
+                    {children}
+                </Box>
+            </AnimatedExpansionPanelContent>
+        </Box>
+    )
+}
+
+export { ExpansionPanel, ExpansionPanelContent, ExpansionPanelHeader, ExpansionPanelToggle }
```
doistbot, line 247: <big><kbd>🟡 P2</kbd></big>  Repository guidelines require exporting types explicitly (`export type { ButtonProps }`). `ExpansionPanelProps` and `ExpansionPanelToggleProps` should be exported here (and subsequently from the folder's `index.ts`) so consumers can strongly type their implementations.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+    id: '',
+    onToggleExpand: () => undefined,
+})
+
+type ExpansionPanelProps = {
+    /**
+     * The id to apply to the expanded content, used by the toggle's
+     * aria-controls attribute
+     */
+    id: string
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Requiring `id` here leaks the panel's internal `aria-controls` wiring into every call site. We already auto-generate stable IDs for similar relationships elsewhere, and React 18 gives us `useId()` for this. Making `id` optional and generating a fallback would avoid stringly-typed setup and accidental duplicate IDs across panels.

doistbot (commented):
This adds a well-animated `ExpansionPanel` disclosure primitive and consolidates duplicated logic from our apps.

Few things worth tightening:
- **API surface:** Make `id` optional with an internal fallback; switch `isExpanded` fallback from `||` to `??`; stop silently overwriting consumer `onClick`/`startIcon`; and explicitly export `ExpansionPanelProps` and `ExpansionPanelToggleProps`.
- **Refs & docs:** Wrap `ExpansionPanelToggle`, `ExpansionPanelHeader`, and `ExpansionPanelContent` in `React.forwardRef`; remove the redundant outer `Box`; and add missing JSDoc for `children` and `aria-describedby`.
- **Performance:** Add `mountOnEnter`/`unmountOnExit` so collapsed panels unmount their subtrees, and use a stable functional updater for the uncontrolled toggle to avoid unnecessary re-renders.
- **Tests:** Add a `jest-axe` a11y check, cover the `Button` toggle branch with children, and assert the `onEntered` callback fires after the animation.

<details>
<summary>Optional follow-up notes (3)</summary>

- **[P3] src/expansion-panel/expansion-panel.tsx:39**: Repository guidelines state: "Add JSDoc comments to every prop". The `children` prop here, as well as `aria-describedby` in `ExpansionPanelToggleProps`, are missing documentation.
- **[P3] src/expansion-panel/expansion-panel.tsx:196**: This outer `<Box>` wrapper is redundant, as `AnimatedExpansionPanelContent` already returns a `Box` at its root. Removing it will flatten the DOM tree.
- **[P3] src/expansion-panel/expansion-panel.tsx:18**: The inline function `() => setState(!state)` captures `state` and creates a new reference on every render in uncontrolled mode (unless automatically cached by the React Compiler). Because `onToggleExpand` is included in the `ExpansionPanel` context value's dependency array, an unstable reference can defeat `useMemo` and force all context consumers to re-render on unrelated parent updates. Using a stable updater callback (`const toggle = React.useCallback(() => setState(s => !s), [])`) guarantees the reference never changes and prevents this unnecessary invalidation.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1075&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A75de54d0-6993-11f1-8f16-54722535616d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A75de54d0-6993-11f1-8f16-54722535616d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/expansion-panel/animated-expansion-panel-content.tsx
```diff
@@ -0,0 +1,148 @@
...
+        // Reading this property is important, even if we do not consume the value.
+        // Without this, the expanded -> collapsed animation will not work.
+        // eslint-disable-next-line @typescript-eslint/no-unused-expressions
+        wrapperRef.current?.clientHeight
+        setElementHeight(transitionElementRef.current, 0)
+        setContentMotion(contentRef.current, 0, 0)
+    }, [])
+
+    return (
+        <Transition
```
doistbot, line 120: <big><kbd>🟡 P2</kbd></big>  `Transition` is left at its default mount behavior here, so a collapsed panel still renders and keeps its entire content subtree mounted; the `display: 'none'` on the container only skips paint. In a list/sidebar with many collapsed panels, that eagerly runs all child renders/effects up front and keeps hidden content alive after collapse. Add `mountOnEnter` at minimum, and `unmountOnExit` too if preserving hidden state isn't required.

src/expansion-panel/expansion-panel.test.tsx
```diff
@@ -0,0 +1,64 @@
+import * as React from 'react'
```
doistbot, line 1: <big><kbd>🟡 P2</kbd></big>  Missing `jest-axe` accessibility test. The project's CLAUDE.md requires every component to have an a11y test, and every other component in the repo follows this convention. Add:

```ts
import { axe } from 'jest-axe'

it('renders with no a11y violations', async () => {
    const { container } = renderPanel()
    expect(await axe(container)).toHaveNoViolations()
})
```

src/expansion-panel/expansion-panel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    ExpansionPanelContent,
+    ExpansionPanelHeader,
+    ExpansionPanelToggle,
+} from './expansion-panel'
+
+type PanelOverrides =
+    | { initiallyExpanded?: boolean }
+    | { isExpanded: boolean; onToggleExpand: () => void }
+
+function renderPanel(overrides: PanelOverrides = {}) {
```
doistbot, line 18: <big><kbd>🟡 P2</kbd></big>  The `renderPanel` helper and all three tests only exercise the icon-only (`IconButton`) toggle path. The `Button` variant (when `ExpansionPanelToggle` receives children) is a materially different rendering branch — it renders a `<Button>` with `startIcon` instead of an `<IconButton>` with `icon`, has distinct CSS, and is one of the two documented usage patterns in the Storybook stories. No test covers it. A simple additional test that renders `renderPanel` with an `ExpansionPanelToggle` that has children would verify the toggle still functions and wires `aria-expanded`/`aria-controls` correctly in that variant.

src/expansion-panel/expansion-panel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    return render(
+        <ExpansionPanel id="panel-content" {...overrides}>
+            <ExpansionPanelHeader>
+                <ExpansionPanelToggle aria-label="Toggle section" />
+            </ExpansionPanelHeader>
+            <ExpansionPanelContent>
```
doistbot, line 24: <big><kbd>🟡 P2</kbd></big>  The `onEntered` callback prop on `ExpansionPanelContent` is part of the public API (fired when the expand animation completes) but is never exercised. Use `jest.useFakeTimers()` to advance past the 200 ms transition timeout and assert the callback fires after expanding.

src/expansion-panel/expansion-panel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <span>Body content</span>
+            </ExpansionPanelContent>
+        </ExpansionPanel>,
+    )
+}
+
+describe('ExpansionPanel', () => {
```
doistbot, line 44: <big><kbd>🟡 P2</kbd></big>  Per project conventions, every component should have an accessibility test. Please add a `jest-axe` check to this suite (e.g. `expect(await axe(container)).toHaveNoViolations()`).

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Animated expand/collapse primitive. Compose with `ExpansionPanelHeader`,
+ * `ExpansionPanelToggle`, and `ExpansionPanelContent` as children.
+ *
+ * Supports both controlled (`isExpanded` + `onToggleExpand`) and uncontrolled
+ * (`initiallyExpanded`) modes; mixing the two logs a warning and falls back to
+ * controlled.
+ */
+function ExpansionPanel({ children, initiallyExpanded, id, ...props }: ExpansionPanelProps) {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Per project conventions (`CLAUDE.md`), design system components should use `React.forwardRef`. Please wrap `ExpansionPanel`, `ExpansionPanelHeader`, `ExpansionPanelToggle`, and `ExpansionPanelContent` in `React.forwardRef` so consumers can attach refs to the underlying DOM elements if needed.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+    const ChevronIcon = size === '24' ? ChevronDownIcon : ChevronDownSmallIcon
+
+    if (!children) {
+        return (
+            <IconButton
+                variant="quaternary"
+                size="small"
+                {...props}
+                onClick={onToggleExpand}
```
doistbot, line 195: <big><kbd>🟡 P2</kbd></big>  Placing `onClick={onToggleExpand}` after `{...props}` silently overwrites any `onClick` handler provided by the consumer. Consider merging them so the consumer's handler still fires (e.g. `onClick={(e) => { props.onClick?.(e); onToggleExpand(); }}`). This also applies to the `Button` on line 168.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+import { ChevronDownIcon, ChevronDownSmallIcon } from './chevron-down-icon'
+
+import styles from './expansion-panel.module.css'
+
+/** Make the listed keys of `T` optional. */
+type SetOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+
+function useControlledState(value = false, onToggle?: () => void): [boolean, () => void] {
+    const [state, setState] = React.useState(value)
+    return onToggle === undefined ? [state, () => setState(!state)] : [value, onToggle]
```
doistbot: <big><kbd>🟡 P2</kbd></big>  In uncontrolled mode, returning `() => setState(!state)` creates a new function reference on every render. This defeats the `React.useMemo` for `contextValue` below, causing the context to update and re-render all consumers on every re-render of `ExpansionPanel`. Consider using a stable callback: `React.useCallback(() => setState(s => !s), [])`.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+}: React.ComponentProps<typeof Box>): React.ReactElement {
+    return <Box {...props}>{children}</Box>
+}
+
+type ExpansionPanelToggleProps = {
+    /**
+     * Required accessible label for the toggle button. The consumer owns
+     * localization — typically something like `"Toggle list of Projects"`.
+     */
+    'aria-label': string
```
doistbot: <big><kbd>🟡 P2</kbd></big>  The toggle prop surface is inaccurate here. This top-level `aria-label` requirement forces text-button callers to keep a second label string even though visible children already name the button, and the inherited `Button`/`IconButton` props still allow `onClick`/`aria-expanded`/`aria-controls` even though those are always overwritten below. Please split the branches so only the icon-only variant requires `aria-label`, and omit the internally owned props from both branches.

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+import { AnimatedExpansionPanelContent } from './animated-expansion-panel-content'
+import { ChevronDownIcon, ChevronDownSmallIcon } from './chevron-down-icon'
+
+import styles from './expansion-panel.module.css'
+
+/** Make the listed keys of `T` optional. */
+type SetOptional<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>
+
+function useControlledState(value = false, onToggle?: () => void): [boolean, () => void] {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Controlled mode is derived from `onToggle` being present instead of from `isExpanded`. In JS consumers, passing `isExpanded` without `onToggleExpand` produces a hybrid component that only seeds internal state from the prop once, so later prop updates are ignored. It would be safer to decide control from `isExpanded !== undefined` and warn or throw when the controlled pair is incomplete.

src/expansion-panel/expansion-panel.tsx
doistbot: <big><kbd>🟡 P2</kbd></big>  Requiring `id` on the root exposes an internal implementation detail to every caller. The only current use is wiring `aria-controls` to the content element, so callers now have to manufacture globally unique ids just to render a panel. Consider making this optional and generating a stable fallback internally, with an override only when a consumer needs a specific id.

src/expansion-panel/expansion-panel.test.tsx
doistbot, line 18: <big><kbd>🟡 P2</kbd></big>  The `renderPanel` helper and all three test cases only exercise the icon-only toggle path (no children). The `ExpansionPanelToggle` component has a second rendering variant — when children are passed it renders a full-width `Button` with `startIcon` instead of an `IconButton`. This distinct code path with different prop forwarding and DOM structure is untested.

src/expansion-panel/expansion-panel.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+    it('wires the toggle to the content via aria-controls', () => {
+        renderPanel()
+
+        const toggle = screen.getByRole('button', { name: 'Toggle section' })
+        expect(toggle).toHaveAttribute('aria-controls', 'panel-content')
+        expect(document.getElementById('panel-content')).toBeInTheDocument()
+    })
+
+    it('toggles its expanded state in uncontrolled mode', async () => {
```
doistbot, line 73: <big><kbd>🟡 P2</kbd></big>  Tests verify `aria-expanded` but never assert whether `ExpansionPanelContent` is actually shown or hidden. Add visibility assertions (e.g., `toBeVisible()` / `not.toBeVisible()`) for the initial expanded and collapsed states to catch animation regressions.

src/expansion-panel/expansion-panel.tsx
doistbot: <big><kbd>🟠 P1</kbd></big>  `props.isExpanded || initiallyExpanded` evaluates to `undefined` when `isExpanded` is explicitly `false` (since `false || undefined` → `undefined`). This breaks `aria-expanded` and leaves the panel in an invalid state in controlled mode. Use `props.isExpanded ?? initiallyExpanded` instead.

doistbot (commented):
Adds a clean, animated `ExpansionPanel` disclosure primitive that consolidates duplicate expand/collapse behavior across projects.

Few things worth tightening:

- Fix the controlled-mode state logic: `props.isExpanded || initiallyExpanded` breaks when `isExpanded` is explicitly `false`, so use `??` instead. Derive controlled mode from `isExpanded !== undefined` rather than the presence of `onToggleExpand`, and memoize the uncontrolled toggle callback with `useCallback` so context stability isn't lost.
- Tighten the toggle API so `aria-label` is required only for the icon-only variant, internally owned props (`onClick`, `aria-expanded`, `aria-controls`) are omitted from the public type, consumer `onClick` handlers are preserved, and the root `id` is optional with an internal fallback.
- Align with Reactist conventions: wrap all subcomponents in `React.forwardRef`, add a `jest-axe` accessibility check, explicitly export the component prop types, and move the story file from `.jsx` to `.tsx`.
- Expand tests to cover the full-width `Button` toggle variant and assert that `ExpansionPanelContent` is actually shown or hidden in expanded and collapsed states.

<details>

- **[P3] src/expansion-panel/expansion-panel.tsx:33**: Per project conventions ("Export types explicitly"), please export `ExpansionPanelProps` (and `ExpansionPanelToggleProps`, etc.) so consumers can safely type their own wrappers.
- **[P3] src/expansion-panel/expansion-panel.stories.jsx:1**: Storybook file uses `.jsx` but the documented convention prefers `.stories.mdx` or `.stories.tsx`. Consider renaming to `.stories.tsx` for consistency with the rest of the TypeScript codebase.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1075&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A776a9a20-6993-11f1-9848-63cc165b622f%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A776a9a20-6993-11f1-9848-63cc165b622f&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

frankieyan:
Hey @benbreckler, thanks for tackling this. Could you please address Doistbot's comments and then re-request a review from me?

benbreckler:
Sure, I asked AI to address the Doistbot comments. Let me know if you have feedback @frankieyan 🙏

src/expansion-panel/expansion-panel.test.tsx
benbreckler: Done in 8eb36694 — added a `jest-axe` check to the suite: `const { container } = renderPanel({ initiallyExpanded: true }); expect(await axe(container)).toHaveNoViolations()`.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — wrapped all four subcomponents (`ExpansionPanel`, `ExpansionPanelHeader`, `ExpansionPanelToggle`, `ExpansionPanelContent`) in `React.forwardRef` with named functions, forwarding the ref to the underlying DOM element.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — resolved by omitting the internally-owned props (`onClick`, `aria-expanded`, `aria-controls`) from the public toggle type. Consumers can no longer pass `onClick`, so there's nothing to silently overwrite or merge.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — the uncontrolled toggle is now a stable callback: `React.useCallback(() => setInternalState((expanded) => !expanded), [])`, so `contextValue` no longer changes on every render.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — split the union so `aria-label` is required only for the icon-only variant (text buttons are named by their children), and the internally-owned props (`onClick`/`aria-expanded`/`aria-controls`) are omitted from both branches.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — controlled mode is now derived from `isExpanded !== undefined` rather than the presence of `onToggleExpand`, so later prop updates are honored. Also added a `console.warn` when `isExpanded` is passed without `onToggleExpand`.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — `id` is now optional with a `React.useId()` fallback; consumers only pass one when they need a specific, stable DOM id.

— Claude

src/expansion-panel/expansion-panel.test.tsx
benbreckler: Done in 8eb36694 — added `renderButtonTogglePanel` and a test covering the full-width `Button` variant (children): it asserts the button is named by its children, wires `aria-controls`, and toggles `aria-expanded`.

— Claude

src/expansion-panel/expansion-panel.test.tsx
benbreckler: Done in 8eb36694 — added a visibility test: content `toBeVisible()` when `initiallyExpanded`, and `not.toBeVisible()` when collapsed (a collapsed panel starts in the `exited` / `display:none` state, so this is assertable without advancing timers).

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 8eb36694 — switched to `props.isExpanded ?? initiallyExpanded`. (The `value = false` default parameter happened to coerce the `false` case back to `false`, so it wasn't an observable bug — but `??` is the correct intent.)

— Claude

src/expansion-panel/animated-expansion-panel-content.module.css
```diff
@@ -0,0 +1,12 @@
+.container {
+    min-height: 0;
+    height: 0;
+    transition-duration: 200ms;
```
frankieyan, line 4: Does the design team have tokens assigned to transition timings and easing functions? This could be a good opportunity to introduce them here.

While we have [variables defined for them in Todoist](https://github.com/Doist/todoist-web/blob/main/src/05-common/styles/variables/animations.css), we haven't done a good job extracting these into variables here in Reactist.

src/expansion-panel/animated-expansion-panel-content.tsx
```diff
@@ -0,0 +1,148 @@
+import * as React from 'react'
+import { Transition } from 'react-transition-group'
+
+import { Box } from '../box'
+
+import styles from './animated-expansion-panel-content.module.css'
+
+const HEIGHT_TRANSITION_DURATION = 200 /* milliseconds */
```
frankieyan: Instead of adding a comment for the unit, it is more common that we include it in the variable name:


```suggestion
const HEIGHT_TRANSITION_DURATION_MS = 200
```

src/expansion-panel/animated-expansion-panel-content.tsx
```diff
@@ -0,0 +1,148 @@
...
+        setElementHeight(transitionElementRef.current, 0)
+        setContentMotion(contentRef.current, 0, CONTENT_ENTER_OFFSET, OPEN_TRANSITION)
+    }, [])
+
+    const handleEntering = React.useCallback(() => {
+        if (!transitionElementRef.current) {
+            return
+        }
+
+        setElementHeight(transitionElementRef.current, wrapperRef.current?.clientHeight ?? 0)
```
frankieyan, line 86: 💭 Nothing to action on here, but hopefully, [`interpolate-size`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/interpolate-size) comes to Firefox and Safari soon so this can be done in CSS.

src/expansion-panel/animated-expansion-panel-content.tsx
```diff
@@ -0,0 +1,148 @@
...
+            onEntered={handleEntered}
+            onExiting={handleExiting}
+            onExit={handleExit}
+            timeout={HEIGHT_TRANSITION_DURATION}
+            in={isExpanded}
+        >
+            {(state) => (
+                <Box
+                    ref={transitionElementRef}
+                    overflow="hidden"
```
frankieyan: Having overflow: hidden here could be confusing, because we're splitting up the logic by also setting overflow: visible via the `.entered` class. I'd consolidate it in one place:


```suggestion
                    overflow={state === 'entered' ? 'visible' : 'hidden}
```

Then, remove `overflow: visible;` from the `.entered` selector.

src/expansion-panel/animated-expansion-panel-content.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    className={[styles.container, state === 'entered' ? styles.entered : null]}
+                    style={state === 'exited' ? { display: 'none' } : undefined}
```
frankieyan: `<Box>` exposes `display` so we can express this without the style attribute:


```suggestion
                    display={state === 'exited' ? 'none' : 'block'}
```

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,268 @@
...
+const ExpansionPanelToggle = React.forwardRef<HTMLButtonElement, ExpansionPanelToggleProps>(
+    function ExpansionPanelToggle(
+        { exceptionallySetClassName, children, 'aria-label': ariaLabel, size = '24', ...props },
+        ref,
+    ) {
+        const { id, isExpanded, onToggleExpand } = React.useContext(ExpansionPanelStateContext)
+
+        const ChevronIcon = size === '24' ? ChevronDownIcon : ChevronDownSmallIcon
+
+        // `props` is the toggle's discriminated union minus the destructured keys.
```
frankieyan: TypeScript is fine with `props` below (on line 199 and 220) being spread without the `as` assertions, so I'd remove those + this comment.  It works because TypeScript is a little loose when it comes to spreading objects as props, but if we want to properly discriminate between the Icon and Button prop types, we'll have to do the prop descructuring within and outside the if branch here, and IMO that's not worth the verbosity.

So three things:

- Remove `// `props` is the toggle's....`
- Replace `{...(props as ....)}` in the two instances below with `{...props}`

src/expansion-panel/expansion-panel.tsx
```diff
@@ -0,0 +1,268 @@
...
+    /** Called once the expand animation finishes (not fired on collapse). */
+    onEntered?: () => void
+}
+
+/**
+ * Container for the panel's expandable content. Animates height on
+ * expand/collapse driven by the parent `ExpansionPanel`'s state. Extra props
+ * forward to the inner `Box`.
+ */
+const ExpansionPanelContent = React.forwardRef<HTMLDivElement, ExpansionPanelContentProps>(
```
frankieyan: I don't think we need to expose `ref` from these components. What ref does is expose the HTML element for further manipulation beyond the exposed/documented API (the props). I don't think we should allow this flexibility unless we find a good reason to.

So, please remove the wrapp `const ExpansionPanelContent = React.forwardRef<HTMLDivElement, ExpansionPanelContentProps>` and just leave ExpansionPanelContent as a function definition.

Same applies to the other components like `ExpansionPanel`, `ExpansionPanelToggle`, and `ExpansionPanelHeader`

src/expansion-panel/expansion-panel.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import {
+    ExpansionPanel,
+    ExpansionPanelContent,
+    ExpansionPanelHeader,
+    ExpansionPanelToggle,
+} from './expansion-panel'
+
+export default {
+    title: '📐 Layout/ExpansionPanel',
```
frankieyan: Is this really a layout component? In https://github.com/Doist/frontend-issues/issues/1350 I've grouped this under "🧭 Navigation & structure", which we don't have in Reactist yet. Would you agree that's a more appropriate category?

src/expansion-panel/expansion-panel.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    component: ExpansionPanel,
+}
+
+export const IconToggle = {
+    name: 'Icon toggle',
+    render: () => (
+        <Box maxWidth="small">
```
frankieyan: In Todoist's Storybook, the panel is given a different colour than the background so it's easier to see where the panel is:

Image

Here I actually missed seeing the chevron at first and tried to click on "Fruit", because it's so far away and appears disconnected

Image

Could we make the demo more obvious?

src/expansion-panel/expansion-panel.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+            <ExpansionPanel id="expansion-panel-fruit" initiallyExpanded>
+                <ExpansionPanelHeader
+                    display="flex"
+                    alignItems="center"
+                    justifyContent="spaceBetween"
+                >
+                    <Text size="caption" weight="semibold" tone="secondary">
```
frankieyan: Does it make sense that these are smaller than the content below? Caption = 12px and body is 14px.

src/expansion-panel/animated-expansion-panel-content.tsx
frankieyan: 💭 Also, we do end up defining 200ms and the curves twice - once here and again in `animated-expansion-panel-content.module.css`. Nothing to action on for now but we should implement a way for these to be de-duped.

frankieyan (commented):
Looking mostly good, Ben. Left some comments for you, once addressed I think it'd be good to go, but please ping me again when ready.

src/expansion-panel/animated-expansion-panel-content.module.css
benbreckler: We discussed an [animation framework recently](https://comms.todoist.com/69/ch/AKJwGSH15XQha796WrGq7/t/Cb9rVjE1CVz7wzVg7LFkx/c/CbwfpcX7BcYUBwX1qV1VK) but we don't have a final plan and values yet. 

I would suggest we go with this approach for now and then use the variables to replace all current values later on?

src/expansion-panel/animated-expansion-panel-content.tsx
benbreckler: Done in 20a0d64d — renamed to `HEIGHT_TRANSITION_DURATION_MS`. On the duplicated `200ms`/curves across the TS and CSS: agreed, left as-is for now per your note; worth a follow-up to dedupe (and tokenize timings/easings more broadly).

— Claude

src/expansion-panel/animated-expansion-panel-content.tsx
benbreckler: Agreed, nothing to change here — would be great to move this into CSS once `interpolate-size` lands across browsers.

— Claude

src/expansion-panel/animated-expansion-panel-content.tsx
benbreckler: Done in 20a0d64d — `overflow` is now driven by the Box prop (`overflow={state === 'entered' ? 'visible' : 'hidden'}`) and removed from the `.entered` selector.

— Claude

src/expansion-panel/animated-expansion-panel-content.tsx
benbreckler: Done in 20a0d64d — switched to `display={state === 'exited' ? 'none' : 'block'}` via the Box prop, dropping the inline `style`.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 20a0d64d — removed the `as` casts and the comment; plain `{...props}` type-checks, as you said.

— Claude

src/expansion-panel/expansion-panel.tsx
benbreckler: Done in 20a0d64d — dropped `React.forwardRef` from all four subcomponents (plain function definitions).

Heads-up: `CLAUDE.md` currently lists `React.forwardRef` as the design-system convention (that's what the Doistbot review cited), so you may want to nuance that doc. Happy to follow your call on it here.

— Claude

src/expansion-panel/expansion-panel.stories.tsx
benbreckler: Done in 20a0d64d — moved to `🧭 Navigation & structure/ExpansionPanel`. Agreed that's a better fit than Layout.

— Claude

src/expansion-panel/expansion-panel.stories.tsx
benbreckler: Done in 20a0d64d — the demo panel now has a background (`background="aside"` + border radius + padding) so it reads as a distinct panel and the chevron no longer feels disconnected from the title.

— Claude

src/expansion-panel/expansion-panel.stories.tsx
benbreckler: Yes, I aligned in e3737de0: the title now uses `size="body"` (14px, matching the content) with `tone="secondary"` + `weight="semibold"`, matching todoist-web's sidebar section header.

src/expansion-panel/expansion-panel.stories.tsx
```diff
@@ -12,27 +12,27 @@ import {
 } from './expansion-panel'
 
 export default {
-    title: '📐 Layout/ExpansionPanel',
+    title: '🧭 Navigation & structure/ExpansionPanel',
```
frankieyan, line 15: One thing we missed is that the Storybook hierarchy is specifically defined:

https://github.com/Doist/reactist/blob/53777ff133389d10200d021a353a4731a98bfc1b/.storybook/preview.ts#L22-L38

Without adding this new category there, it ends up at the bottom of the nav:

Image

Can we insert it into the list in alphabetical order?

frankieyan (changes requested):
One last change to fix the story nav. After that it's good to go. Also, you'll need to rebase the branch against `origin/main` before you can do so to resolve conflicts with the latest changes on main.

src/expansion-panel/expansion-panel.stories.tsx
benbreckler: Done in c3ddcd57 — added `🧭 Navigation & structure` to the `storySort` `order` in `.storybook/preview.ts`, alphabetically between `📑 Menus & tabs` and `🪟 Overlays`, so the section sorts in place instead of falling to the bottom.

Also rebased onto `origin/main` (ee032d15) to resolve the conflicts — that pulled in the React 19 support, so I took main's React 18/19 setup and layered the `react-transition-group` peer dep on top, and fixed the chevron's typing for React 19 (`import type { JSX } from 'react'`). Green across lint, both type-checks, both test suites, and build.

— Claude

frankieyan (approved):
Good to go, Ben. Once you merge it, a release will be created automatically (takes a few minutes) for you to add to Comms 👍

PR #1076, opened 2026-06-18
docs: add Figma design links and real Badges to Storybook toolbar

frankieyan:
## Short description

This PR adds Figma references to most of our component stories in the Storybook toolbar, sourced from https://github.com/Doist/frontend-issues/issues/1350.

We also switch the a11y/deprecated badges in the toolbar to use the real Reactist `<Badge`>.

## PR Checklist

## Demo

<table>
  <thead>
    <tr>
      <th width="50%">Before</th>
      <th width="50%">After</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>
image

</td>
      <td>image</td>
    </tr>
  </tbody>
</table>


.storybook/figma/figma-tool.tsx
```diff
@@ -0,0 +1,35 @@
...
+import { FigmaIcon } from './figma-icon'
+
+export const FigmaTool = React.memo(function FigmaTool() {
+    const figmaParameter = useParameter(FIGMA_PARAMETER, undefined)
+    const links = React.useMemo(() => resolveFigmaLinks(figmaParameter), [figmaParameter])
+
+    return (
+        <Inline space="xsmall">
+            {links.length === 0 ? (
+                <Badge tone="info" label="No Figma link" />
```
doistbot, line 24: <big><kbd>🟡 P2</kbd></big>  This renders a false negative on docs pages that define their own MDX `<Meta>` instead of inheriting story meta. For example, `src/modal/modal-docs.mdx` has no `figma` parameter, so the modal docs page will now show `No Figma link` even though `src/modal/modal-examples.stories.tsx` adds one. Hiding the tool when `links.length === 0`, or wiring `figma` into those docs pages, would avoid that misleading state.

.storybook/badges/badges.ts
```diff
@@ -1,28 +1,21 @@
...
     id: string
     title: string
-    styles: CSSProperties | undefined
+    tone: BadgeTone
 }
 
-function isRecord(value: unknown): value is UnknownRecord {
-    return typeof value === 'object' && value !== null && !Array.isArray(value)
-}
+const VALID_TONES: readonly BadgeTone[] = ['info', 'positive', 'promote', 'attention', 'warning']
```
doistbot: <big><kbd>🟡 P2</kbd></big>  `VALID_TONES` is now a second source of truth for the `Badge` tones. Since `badgesConfig` isn't statically checked, adding or renaming a `Badge` tone later will silently make valid toolbar badges disappear until this array is updated too. Please export a runtime tone list or validator from the badge module and reuse it here.

.storybook/figma/figma.ts
```diff
@@ -0,0 +1,33 @@
...
+    label: string
+    url: string
+}
+
+function isRecord(value: unknown): value is Record<string, unknown> {
+    return typeof value === 'object' && value !== null && !Array.isArray(value)
+}
+
+function resolveEntry(entry: unknown): ResolvedFigmaLink[] {
+    if (typeof entry === 'string') {
```
doistbot: <big><kbd>🟡 P2</kbd></big>  Treating any non-empty string as both `label` and `url` makes `figma` a stringly-typed parameter: a copied label or typo becomes a broken `href` instead of being rejected. All current usages in this PR already use `{ label, url }`, so I'd drop the bare-string branch and keep the parameter shape explicit.

doistbot (commented):
This PR adds Figma design links to the Storybook toolbar for most component stories and switches the a11y/deprecated badges to use the real Reactist `<Badge>` component.

Few things worth tightening:
- The Figma tool shows a "No Figma link" message on docs pages that define their own MDX `<Meta>` (e.g. `modal-docs.mdx`) instead of inheriting story meta—consider hiding the tool when there are no links, or wiring `figma` into those docs pages.
- `VALID_TONES` duplicates the Badge component's tone list as a second source of truth; exporting a runtime tone list or validator from the badge module and reusing it here would prevent valid badges from silently disappearing if tones change.
- The bare-string branch in the Figma parameter resolver treats any non-empty string as both label and URL, turning typos into broken links—since all current usages already use `{ label, url }`, dropping the string shorthand would keep the parameter shape explicit.
- `IconButton` (which always wraps in a `Tooltip` and pulls `@ariakit/react` into the manager bundle) is heavier than needed for a simple external link in the Storybook manager; a lightweight `<a>` or button with a native `title` would reduce startup overhead on every page.

<details>

- **[P3] .storybook/figma/figma-tool.tsx:23**: `IconButton` is pretty expensive in manager UI. In this code path it always wraps itself in `Tooltip` (`tooltip={link.label}`), which pulls `@ariakit/react` into the Storybook manager bundle and creates a tooltip store for each Figma link. Because `manager.ts` loads on every Storybook page, this adds avoidable startup and render overhead for a simple external link. A lightweight `<a>`/button here with a native `title` would keep the manager much smaller.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1076&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Ab824e400-6ae0-11f1-9be2-76c7c284abf3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Ab824e400-6ae0-11f1-9be2-76c7c284abf3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

.storybook/figma/figma-tool.tsx
frankieyan: We do want new stories to show "No Figma" as a reminder, so I don't think we should default to not showing anything. So in [2a63b20](https://github.com/Doist/reactist/pull/1076/commits/2a63b2073e0a66e081dc19158348a0f713efe6e6), we add the entry to modal-docs.mdx, and add `figma: false` as an explicit escape hatch.

.storybook/badges/badges.ts
frankieyan: ✅ It's now type checked: [517cf27](https://github.com/Doist/reactist/pull/1076/commits/517cf27772fd4e85d8b3ee89b69ebe447aee1651) + [a74d315](https://github.com/Doist/reactist/pull/1076/commits/a74d315817d3f615c25ad480c8d316cf7e955e25)

.storybook/figma/figma.ts
frankieyan: ✅ [69595fc](https://github.com/Doist/reactist/pull/1076/commits/69595fcc3b991c62d1d5d14346fcfc926284f65f)

nats12 (approved):

PR #1079, opened 2026-06-19
fix(tooltip): Let ariakit merge the child ref to avoid React ref warning

frankieyan:
## Short description

In #1074, we tried to match `Tooltip`'s previous implementation where we extract and pass along its child's ref to Ariakit, while keeping React 18 and 19 compatibility.. However, this causes a dev mode warning in React 18 when we try to test `props.ref`:

> Warning: `ref` is not a prop. Trying to access it will result in `undefined` being returned.
> If you need to access the same value within the child component, you should pass it as a different
> prop. (https://reactjs.org/link/special-props)

Turns out, Ariakit already merges the child element's ref and check for `props.ref` through its `render` prop:

* https://github.com/ariakit/ariakit/blob/f6b6376e57f910c9163b89a36e717a1e857fe43c/packages/ariakit-react-core/src/utils/system.tsx#L39
* https://github.com/ariakit/ariakit/blob/f6b6376e57f910c9163b89a36e717a1e857fe43c/packages/ariakit-react-core/src/utils/misc.ts#L41-L44

So we can remove this logic and let Ariakit handle it: `<TooltipAnchor render={child} store={tooltip} />`

## PR Checklist

### DIFF
--- a/src/tooltip/tooltip.test.tsx
+++ b/src/tooltip/tooltip.test.tsx
@@ -159,7 +159,7 @@ describe('Tooltip', () => {
         jest.useRealTimers()
     })
 
-    it("does not interfere with the trigger element's ref forwarding", () => {
+    it("does not interfere with the trigger element's ref forwarding", async () => {
         const buttonRef = React.createRef<HTMLButtonElement>()
 
         render(
@@ -168,7 +168,14 @@ describe('Tooltip', () => {
             </Tooltip>,
         )
 
-        expect(buttonRef.current).toBe(screen.getByRole('button'))
+        const button = screen.getByRole('button', { name: 'Click me' })
+        expect(buttonRef.current).toBe(button)
+
+        const user = userEvent.setup()
+        await user.hover(button)
+        expect(
+            await screen.findByRole('tooltip', { name: 'tooltip content here' }),
+        ).toBeInTheDocument()
     })
 
     it('supports components with an "as" prop as the anchor', async () => {
--- a/src/tooltip/tooltip.tsx
+++ b/src/tooltip/tooltip.tsx
@@ -144,22 +144,9 @@ const Tooltip = React.forwardRef<TooltipStore, TooltipProps>(
             return child
         }
 
-        /* eslint-disable react-hooks/refs */
-        const rawRef =
-            'ref' in child.props
-                ? child.props.ref
-                : // child.ref access kept for React 18 compatibility
-                  (child as { ref?: unknown }).ref
-
-        if (typeof rawRef === 'string') {
-            throw new Error('Tooltip: String refs cannot be used as they cannot be forwarded')
-        }
-
-        const childRef = rawRef as React.Ref<HTMLDivElement>
-
         return (
             <>
-                <TooltipAnchor render={child} store={tooltip} ref={childRef} />
+                <TooltipAnchor render={child} store={tooltip} />
                 {isOpen && content ? (
                     <AriakitTooltip
                         store={tooltip}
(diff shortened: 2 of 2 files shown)


doistbot (commented):
This PR simplifies Tooltip by removing manual child ref extraction and letting Ariakit handle ref merging via its `render` prop, eliminating a React 18 dev warning about accessing `props.ref`.

Few things worth tightening:
- The new test doesn't actually lock in the regression this PR fixes—consider spying on `console.error` during render and asserting the special-props warning is never logged, so the test would catch it if the warning comes back.
- The hover interaction in the test relies on real timers and blocks for the full 500ms `showTimeout`; switching to fake timers with `flushMicrotasks()` would keep it in line with repo conventions and speed up the suite.

<details>

- **[P3] src/tooltip/tooltip.test.tsx:175**: This hover interaction relies on real timers, causing `findByRole` to block for the duration of the tooltip's 500ms `showTimeout`. To avoid slowing down the test suite and to align with the repository guidelines for Ariakit interactions, consider using fake timers and `flushMicrotasks()`: ```tsx jest.useFakeTimers() const user = userEvent.setup({ advanceTimers: jest.advanceTimersByTime }) await user.hover(button) act(() => { jest.advanceTimersByTime(500) }) await flushMicrotasks() expect(screen.getByRole('tooltip', { name: 'tooltip content here' })).toBeInTheDocument() jest.useRealTimers() ```

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1079&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A18ed2750-6c1e-11f1-8663-6cc03937f531%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A18ed2750-6c1e-11f1-8663-6cc03937f531&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/tooltip/tooltip.test.tsx
```diff
@@ -168,7 +168,14 @@ describe('Tooltip', () => {
             </Tooltip>,
         )
 
-        expect(buttonRef.current).toBe(screen.getByRole('button'))
+        const button = screen.getByRole('button', { name: 'Click me' })
+        expect(buttonRef.current).toBe(button)
+
+        const user = userEvent.setup()
+        await user.hover(button)
```
doistbot, line 175: <big><kbd>🟡 P2</kbd></big>  This still doesn't exercise the bug this PR fixes. The failure mode was a React 18 dev warning during render when `props.ref` is accessed, but these assertions only verify ref forwarding and hover behavior—both can still pass if that warning comes back. Please spy on `console.error` around the render and assert that the special-props warning is not logged.

frankieyan: Honestly, I'm not sure if it warrants an extra test 🤔

gnapse (commented):
✅

PR #1083, opened 2026-06-23
feat(menu): style icon-less link items like text links

rfgamaral:
## Short description

Menu items that are links (rendered via `render`) and have no icon now use the text-link color, instead of the default menu item color. Link items that include an icon are left untouched.

This came out of Quick Add 2.0, where the "Edit task actions" entry is a link shown in the theme's accent color. That color was a one-off override on the menu item, with no real pattern behind it. Rather than keep it as an exception, Panos and I agreed to make it a proper, reusable pattern here in Reactist, so todoist-web (and anything else) can drop the override.

It's CSS-only (`a[role='menuitem']:not(:has(svg))`), so it applies automatically with no new prop or API. Added a story and a short note to the docs.

Smoke tested in todoist-web across all themes.

## Reference

- [Discussion with Panos](https://comms.todoist.com/69/ch/CW8EkaRYvfSKaKSF3D8Nq/t/CbiRrT9BKv3tzbtwMrZJo/)

## PR Checklist

### DIFF
--- a/src/menu/menu.mdx
+++ b/src/menu/menu.mdx
@@ -1,4 +1,4 @@
-import { Meta, Story, Canvas, Controls, Description } from '@storybook/addon-docs/blocks'
+import { Meta, Canvas, Controls, Description } from '@storybook/addon-docs/blocks'
 import KeyboardShortcut from '../components/keyboard-shortcut'
 import { Button, IconButton } from '../button'
 import { Inline } from '../inline'
@@ -15,26 +15,20 @@ import * as MenuStories from './menu.stories'
 A set of components that provide the means to construct a menu that can be triggered
 via an element or as a context menu.
 
-<Canvas>
-    <Story of={MenuStories.SimpleMenuStory} />
-</Canvas>
+<Canvas of={MenuStories.SimpleMenuStory} />
 
 ### `tooltip`
 
 The `tooltip` prop allows a tooltip to be shown over the menu button.
 
-<Canvas>
-    <Story of={MenuStories.TooltipStory} />
-</Canvas>
+<Canvas of={MenuStories.TooltipStory} />
 
 ### `modal`
 
 By default, interaction with elements below the menu is disallowed when the menu is open. This behaviour can
 be changed by passing `modal={false}` to `<MenuList>`
 
-<Canvas>
-    <Story of={MenuStories.OutsideInteractionStory} />
-</Canvas>
+<Canvas of={MenuStories.OutsideInteractionStory} />
 
 ### `<SubMenu>`
 
@@ -42,17 +36,13 @@ You may nest the `<SubMenu>` component within `<Menu>` to create submenus.
 
 On smaller viewports where there isn't enough space to render the submenu on either side, it will be rendered under its parent menu item instead.
 
-<Canvas>
-    <Story of={MenuStories.SubMenuStory} />
-</Canvas>
+<Canvas of={MenuStories.SubMenuStory} />
 
 ### Link menu items
 
 A `MenuItem` rendered as a link (via `render`) that has no icon is automatically styled like a text link. Link items that include an icon keep the default menu item styling.
 
-<Canvas>
-    <Story of={MenuStories.LinkMenuItemStory} />
-</Canvas>
+<Canvas of={MenuStories.LinkMenuItemStory} />
 
 ### `<ContextMenuTrigger>`
 
@@ -61,15 +51,11 @@ for the menu.
 
 In the demo below, right click on the Settings button, or click on the more button:
 
-<Canvas>
-    <Story of={MenuStories.ContextMenuTriggerStory} />
-</Canvas>
+<Canvas of={MenuStories.ContextMenuTriggerStory} />
 
 ### Extra Features
 
 You may also use `<MenuGroup>` to group menu items, `disable` a menu item, or
 return `false` from `onSelect` to prevent the menu from closing when an item is selected.
 
-<Canvas>
-    <Story of={MenuStories.ExtraFeaturesStory} />
-</Canvas>
+<Canvas of={MenuStories.ExtraFeaturesStory} />
--- a/src/menu/menu.mdx
+++ b/src/menu/menu.mdx
@@ -46,6 +46,14 @@ On smaller viewports where there isn't enough space to render the submenu on eit
     <Story of={MenuStories.SubMenuStory} />
 </Canvas>
 
+### Link menu items
+
+A `MenuItem` rendered as a link (via `render`) that has no icon is automatically styled like a text link. Link items that include an icon keep the default menu item styling.
+
+<Canvas>
+    <Story of={MenuStories.LinkMenuItemStory} />
+</Canvas>
+
 ### `<ContextMenuTrigger>`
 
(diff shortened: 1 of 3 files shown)


src/menu/menu.module.css
```diff
@@ -63,6 +63,24 @@
...
+/* Link menu items without an icon are styled like a text link */
+:global(.reactist_menulist) a[role='menuitem']:not(:has(svg)) {
+    color: var(--reactist-text-link-idle-tint, var(--reactist-actionable-tertiary-idle-tint));
+    --reactist-content-primary: var(
+        --reactist-text-link-idle-tint,
+        var(--reactist-actionable-tertiary-idle-tint)
+    );
+}
+
+:global(.reactist_menulist) a[role='menuitem']:not(:has(svg)):hover,
```
doistbot: <a href="#">P2</a>  These `:hover` and `:focus` selectors have a specificity of `(0,3,2)`, which overrides the existing disabled rule `[role='menuitem'][aria-disabled='true']` `(0,3,0)`.

Since `aria-disabled` menu items can still receive keyboard focus, navigating to a disabled link item will incorrectly change its color to the active hover tint. You can prevent this by excluding disabled items from these selectors, for example: `a[role='menuitem']:not(:has(svg)):not([aria-disabled='true']):focus`.

src/menu/menu.module.css
```diff
@@ -63,6 +63,24 @@
     text-decoration: none;
 }
 
+/* Link menu items without an icon are styled like a text link */
+:global(.reactist_menulist) a[role='menuitem']:not(:has(svg)) {
```
doistbot: <a href="#">P2</a>  This only excludes items with a descendant `<svg>`, but menu items here already use non-SVG icons too (for example emoji/string icons in the existing `StructuredMenuItem` examples). A link item with an emoji, image, or avatar icon would still match `:not(:has(svg))` and get the text-link color even though it isn't icon-less. Please key this off an explicit icon wrapper/marker instead of assuming every icon renders as SVG.

doistbot (commented):
Menu items that are links with no icon now render in the text-link color via a CSS-only `:has(svg)` check, with a new story and docs note.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1083&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A589e5060-6ef1-11f1-9ce2-256c5d3b9970%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A589e5060-6ef1-11f1-9ce2-256c5d3b9970&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/menu/menu.module.css
doistbot: <a href="#">P2</a>  The new link styles do not account for the disabled state, causing two visual bugs when a link menu item is disabled (`aria-disabled="true"`):

1. **Hover specificity:** The new hover/focus selectors have higher specificity (`0,3,2`) than the existing generic disabled rule (`0,3,0`). Hovering over a disabled link item overrides its gray color and flashes the blue hover tint.
2. **Idle tint leak:** The idle rule redefines `--reactist-content-primary`. Since the generic disabled rule only forces `color` and leaves custom properties untouched, any inner content relying on this token (like the `<Text>` component in the newly added story) will stubbornly render as blue instead of gray.

To fix this, exclude disabled items from these new styles by appending `:not([aria-disabled='true'])` to both the idle and hover/focus selectors.

doistbot: <a href="#">P2</a>  This makes the menu item's visual variant depend on whether its child subtree happens to contain an `svg`, which breaks the `MenuItem` abstraction boundary. A link with a non-SVG leading visual (`img`, emoji, custom icon wrapper) will still be treated as "icon-less", while a text-only label that includes an incidental inline SVG will lose the text-link styling. Since `MenuItem` accepts freeform children, this is a brittle contract; prefer an explicit class/data attribute/prop for the text-link variant instead of inferring it from `:has(svg)`.

src/menu/menu.module.css
rfgamaral: Good catch. Fixed in a057f979 — excluded `aria-disabled` items from the idle and hover/focus selectors with `:not([aria-disabled='true'])`, and added a disabled link to the story.

src/menu/menu.module.css
rfgamaral: Fixed in a057f979. Excluding `aria-disabled` items from the new styles covers both points: the focus tint and the idle `--reactist-content-primary` leak. Added a disabled link to the story too.

rfgamaral: Intentional. All our icons are SVGs — the emoji/string icons in the examples are just illustrative, not real usage. An explicit prop/attribute would add API surface and undo the automatic, no-API behavior agreed with design, so we're keeping the `:has(svg)` check rather than solving a case that doesn't exist in practice.

rfgamaral: By design, same reasoning as the other thread. Every icon we use is an SVG, so the `:has(svg)` heuristic holds for our cases. Switching to an explicit variant prop adds API we deliberately avoided to keep this automatic.

src/menu/menu.mdx
```diff
@@ -46,6 +46,14 @@ On smaller viewports where there isn't enough space to render the submenu on eit
     <Story of={MenuStories.SubMenuStory} />
 </Canvas>
 
+### Link menu items
+
+A `MenuItem` rendered as a link (via `render`) that has no icon is automatically styled like a text link. Link items that include an icon keep the default menu item styling.
+
+<Canvas>
+    <Story of={MenuStories.LinkMenuItemStory} />
```
frankieyan: I noticed this actually causes all stories on the docs page to show the first one:

Image


So basically `<Canvas>` no longer accepts `<Story>`  as children: https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#canvas-block

Pushing up a fix:

Image

frankieyan (approved):
The selector is pretty narrow, and I can't think of a situation where we have a link without an icon and wouldn't want this applied. Let's ship it 👍

src/menu/menu.mdx
rfgamaral: Thank you, Frankie!

PR #1087, opened 2026-06-25
ci: deploy storybook on merges to main instead of tags

frankieyan:
## Short description

The public Storybook hasn't auto-deployed since the `v30.1.0` release after migrating to semantic-release (#1024), as each release commit uses `[skip ci]` in its commit message. This updates the workflow to be based off of each commit to `main` instead, which while fixes the issue, also keeps the docs updated after `docs:` commits as those don't generate releases.

## PR Checklist

- [ ] Updated docs (storybooks, readme)

### DIFF
--- a/.github/workflows/deploy-storybook.yml
+++ b/.github/workflows/deploy-storybook.yml
@@ -2,8 +2,8 @@ name: Deploy Storybook to Pages
 
 on:
     push:
-        tags:
-            - 'v*'
+        branches:
+            - main
     workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
@@ -13,11 +13,9 @@ permissions:
     id-token: write
     actions: read
 
-# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
-# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
     group: 'pages'
-    cancel-in-progress: false
+    cancel-in-progress: true
 
 jobs:
     # Single deploy job since we're just deploying


.github/workflows/deploy-storybook.yml
```diff
@@ -13,11 +13,9 @@ permissions:
     id-token: write
     actions: read
 
-# Allow only one concurrent deployment, skipping runs queued between the run in-progress and latest queued.
-# However, do NOT cancel in-progress runs as we want to allow these production deployments to complete.
 concurrency:
     group: 'pages'
-    cancel-in-progress: false
+    cancel-in-progress: true
```
doistbot, line 18: <a href="#">P2</a>  With `group: 'pages'` still shared across all runs, setting `cancel-in-progress: true` lets a later `workflow_dispatch` run from an older or non-`main` ref cancel the in-flight `main` deployment and publish stale Storybook content. Keep this `false`, or scope the concurrency/dispatch so only `main` runs can preempt each other.

doistbot (commented):
The workflow now deploys Storybook on pushes to `main` instead of tags, sidestepping the `[skip ci]` messages from semantic-release that had silently broken auto-deployment.

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1087&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253Aafbb0010-7062-11f1-9bc6-4153d0775fc3%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3Aafbb0010-7062-11f1-9bc6-4153d0775fc3&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

.github/workflows/deploy-storybook.yml
```diff
@@ -2,8 +2,8 @@ name: Deploy Storybook to Pages
 
 on:
     push:
-        tags:
-            - 'v*'
+        branches:
```
doistbot, line 5: <a href="#">P2</a>  Triggering this on every push to `main` means we now pay for a full `storybook build` and Pages deploy on merges that can't affect the published Storybook, like CI-only changes or release/changelog updates. Adding a `paths:` filter here for Storybook-relevant files (`src/**`, `.storybook/**`, `package.json`, lockfile, etc.) would avoid that extra runner time.

.github/workflows/deploy-storybook.yml
frankieyan: @doistbot I don't think this is an issue? We either deploy from main automatically, or someone manually runs it, which are served from the same URL. Whatever is triggered later should always "win".

.github/workflows/deploy-storybook.yml
frankieyan: We don't have much activity (especially non-source changes) + the run time is only around a minute.

gnapse (commented):
✅

PR #1101, opened 2026-07-22
feat: add segmented control radio and tabs

nats12:
## Short description

Adds the Todoist segmented-control patterns to Reactist while preserving the existing compound Tabs API.

- Adds options-based `SegmentedControlRadio` and `SegmentedControlTabs` components with controlled and uncontrolled selection.
- Supports disabled options, icons, rich labels, tooltips, extra labels/icons, link-rendered tabs, panel render modes, alignment, and full-width layouts.
- Moves the existing compound Tabs component onto the shared segmented-control styling and removes its old sliding indicator implementation.
- Adds React 18/19 tests, accessibility coverage, and Storybook examples, organized as Segmented Control → Radio/Tabs and Tabs → Compound API.
- Handles equal-height wrapped labels for narrow full-width consumers such as Todoist Developer Options.

Locally verified through Reactist Storybook and in Todoist Web via `yalc`.

## References
- https://github.com/Doist/frontend-issues/issues/1350

## PR Checklist


nats12:
@frankieyan before I send this off to review, it has been a while so wanted to check our process on this. I haven't removed the current Reactist Tabs from in here yet. So currently both exist:

Screenshot 2026-07-30 at 10 25 33

This is largely to not break consumers.

Would we mark as deprecated, then once this is merged, and OUR consumers are on the latest component, then completely remove?

frankieyan:
> @frankieyan before I send this off to review, it has been a while so wanted to check our process on this. I haven't removed the current Reactist Tabs from in here yet. So currently both exist:
> 
> Screenshot 2026-07-30 at 10 25 33
> This is largely to not break consumers.
> 
> Would we mark as deprecated, then once this is merged, and OUR consumers are on the latest component, then completely remove?

Yeah! I think marking the old tabs as deprecated first would be the way to go. 

I'd go with adding a badge to the stories: https://github.com/Doist/reactist/blob/eeac516b9eb297d9e55304580ec63be3e2c724d6/.storybook/badges/README.md?plain=1#L22

And marking the components themselves as deprecated with jsdoc: https://jsdoc.app/tags-deprecated

But not rename the button to avoid a breaking change for now (but of course if you'd like to offload the renaming in consumers or even migrate right way to an agent, that's fine too!)

nats12:
> Yeah! I think marking the old tabs as deprecated first would be the way to go.

> I'd go with adding a badge to the stories:


Thanks @frankieyan, ended up doing this. And I've added you as second reviewer on this one, hope that's OK.

src/tabs/tabs.tsx
```diff
@@ -102,26 +99,37 @@ interface TabProps
...
         <BaseTab
             id={id}
             ref={ref}
             disabled={disabled}
             store={tabStore}
-            render={render}
             className={className}
-            onClick={onClick}
+            accessibleWhenDisabled={accessibleWhenDisabled}
+            render={render ?? (({ style, ...renderProps }) => <button {...renderProps} />)}
```
doistbot: <a href="#">P2</a>  Preserve the `style` prop in this fallback render. Destructuring it away means existing `<Tab style={{ ... }}>` consumers silently lose their styles after this change. Forward it to the button (for example, `style={style}`) or use a fallback element that retains it.

src/segmented-control/segmented-control-tabs.test.tsx
```diff
@@ -0,0 +1,148 @@
...
+        )
+
+        expect(await screen.findByRole('tab', { name: 'Installed' })).toHaveAttribute(
+            'href',
+            '#installed',
+        )
+        expect(screen.getByTitle('Wrapped tab')).toBeVisible()
+    })
+
+    it('removes disabled tabs from keyboard navigation', async () => {
```
doistbot, line 93: <a href="#">P2</a>  No test verifies that clicking a tab in uncontrolled mode (using `initialSelectedOptionId`) actually switches the visible panel. The existing "removes disabled tabs from keyboard navigation" test uses uncontrolled mode but only asserts focus movement and disability — it never checks that the panel content changed (e.g., that "Mentions" panel became visible after `ArrowRight`). The old `Tabs` component tests cover this, and `SegmentedControlRadio` tests cover it for radio buttons. Add an assertion (or a dedicated uncontrolled test) that the correct `tabpanel` or `tabContent` text becomes visible after interaction.

src/tabs/tabs.tsx
nats12: Fixed in f8ee8fed. The fallback button now forwards the merged `style` value, and `tabs.test.tsx` has a regression test proving an inline style supplied to `Tab` reaches the rendered tab button.

src/segmented-control/segmented-control-tabs.test.tsx
nats12: Fixed in f8ee8fed. The uncontrolled keyboard-navigation test now asserts that moving past the disabled tab selects and displays the `Mentions` tab panel.

nats12:
Addressed the Doistbot review feedback in f8ee8fed:

- Preserved `Tab` inline styles in the fallback render and added a regression test.
- Verified uncontrolled `SegmentedControlTabs` keyboard navigation switches the visible panel.
- Reworded the temporary `SegmentedControl` alias deprecation note so it does not imply released consumers.
- Moved the shared animation-frame flush helper into `test-helpers`.
- Reused the controlled/uncontrolled selection prop types across radio and tabs.
- Shared the alignment type and justify-content mapping between the radio and compound-tab implementations.

The PR-size suggestion was not applied: the radio/tabs APIs and their shared styling are one cohesive migration, while splitting now would make their dependency relationship harder to review and land.

Validation: full React 19 and React 18 suites pass (46 suites, 631 passed and 1 skipped per version), along with lint, both type-checks, and React Compiler tracking.

src/segmented-control/segmented-control-tabs.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    render: () => (
+        <Box width="small">
+            <ControlledExample width="full" />
+        </Box>
+    ),
+}
+
+export const FullWidthWithWrappingLabels: Story = {
+    render: () => (
+        <Box width="small">
```
Bloomca, line 68: This could be limiting the full width

src/segmented-control/segmented-control-tabs.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    )
+}
+
+export const Default: Story = {
+    render: () => <ControlledExample />,
+}
+
+export const FullWidth: Story = {
+    render: () => (
+        <Box width="small">
```
Bloomca, line 60: Same here, potentially it limits the full width

src/segmented-control/index.ts
```diff
@@ -0,0 +1,14 @@
+export type {
+    AriaDescriptionProps,
```
Bloomca, line 2: My original reaction was that the name is too generic, but then I checked and it actually describes it very well. But I feel like we should move it somewhere else, maybe in `src/utils/common-types.ts`?

src/segmented-control/segmented-control.module.css
```diff
@@ -33,123 +33,162 @@
...
+.radio-item:has(input:focus-visible) {
+    outline: 5px auto Highlight;
+    outline: 5px auto -webkit-focus-ring-color;
 }
 
-.tab-themed[aria-disabled='true'] {
-    background-color: var(--reactist-tab-themed-disabled-fill);
-    color: var(--reactist-tab-themed-disabled-tint);
+.radio-item > input {
+    clip: rect(0 0 0 0);
```
Bloomca, line 158: I think we should preserve the original comment:

```css
/* ref: https://www.a11yproject.com/posts/how-to-hide-content/ */
```

frankieyan:
Something I didn't realize with the SegmentedControl is that it's still using Ariakit 😅 so this would be a lateral move for us. On the bleeding issue Seva pointed out, it also happens in the developer options, but I thought that was still using the old Tabs - which isn't actually better because it does this:

image

Would it be worth pivoting to the BaseUI [Tabs](https://base-ui.com/react/components/tabs) and [RadioGroup](https://base-ui.com/react/components/radio) instead?

In small viewports, I wonder if we could adopt the pattern that the template gallery uses so that the control is scrollable/draggable:

image

nats12:
> Something I didn't realize with the SegmentedControl is that it's still using Ariakit 😅

@frankieyan Good point. I checked and Ariakit is still used by several Reactist components, including the existing Tabs, Menu, Modal and Tooltip. Moving these segmented controls to Base UI would therefore introduce Base UI to Reactist for the first time, alongside Ariakit.

Is Base UI the direction we want new Reactist components to take, with the remaining Ariakit components migrated gradually? If so, I’m happy to investigate pivoting these two implementations since they're new anyway.

I agree with addressing the narrow viewport behaviour. That seems independent of the underlying primitive, so we could make the control horizontally scrollable when its tabs no longer fit.

frankieyan:
> Is Base UI the direction we want new Reactist components to take, with the remaining Ariakit components migrated gradually? If so, I’m happy to investigate pivoting these two implementations since they're new anyway.

I kinda thought so based on our discussions at the retreat, but we never made it official. Seems @pawelgrimm will do so with an ADR (not a spec or RFC, so it feels pretty final to me!): https://comms.todoist.com/69/ch/B6UxGRat2UEtmfWkdhZ1h/t/CeCwh9nSDpFjyVttUBMwu/

PR #1092, opened 2026-07-07
chore: set displayName on forwardRef components for readable Storybook source

frankieyan:
## Short description

This issue was discovered while implementing stories for https://github.com/Doist/reactist/pull/1089.

Our layout and typography primitives (`Box`, `Stack`, `Inline`, `Columns`, `Text`, etc.) are `forwardRef` components built through `polymorphicComponent`. In Storybook, docgen can't infer the correct component names through that wrapper, so with dynamic stories (i.e. ones that pass args), their code snippets would show `<React.ForwardRef>` instead of the real component names.

This PR solves it by setting an explicit `displayName` property on the component. I explored a different approach by having `polymorphicComponent` forward the underlying component names, however, they are stripped in prod mode so they won't work in the deployed storybook.

### Before / after

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>columns-before</td>
    <td>columns-after</td>
  </tr>
</table>

## PR Checklist

- [ ] Reviewed and approved Chromatic visual regression tests in CI

PR #1102, opened 2026-07-23
feat: add provisional Table component

rmartins90:
Related: https://github.com/Doist/frontend-issues/issues/1369

## Short description

Adds a provisional Reactist `Table` prototype based on Todoist's current flat data table for review during the Reactist session.

- Preserves native table markup, typed columns and data, stable row IDs, controlled manual sorting, empty states, custom rows and cells, and selected-row styling.
- Exposes `TableRow` and `TableCell` building blocks without making the CSS module public.
- Adds presentation-ready Data Display stories for default, sorting, selection, empty, placeholder-row, and narrow viewport states.
- Pins and externalizes `@tanstack/react-table` 8.21.3.

The API is intentionally provisional. This does not add hierarchy, drag and drop, multi-select, persistence, responsive redesign, or Todoist integration.

To review in Storybook:

1. Run `npm run storybook`.
2. Open **Data display / Table / Default** and try sorting with pointer and keyboard.
3. Open **Selected and clickable rows**, then activate rows with pointer, Enter, and Space.
4. Review the empty, custom placeholder, and narrow viewport stories.

## PR Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -155,6 +155,7 @@
     },
     "dependencies": {
         "@babel/runtime": "^7.28.4",
+        "@tanstack/react-table": "8.21.3",
         "aria-hidden": "^1.2.1",
         "dayjs": "^1.8.10",
         "patch-package": "^6.4.6",
--- a/src/table/table.module.css
+++ b/src/table/table.module.css
@@ -1,0 +1,117 @@
+.table {
+    width: 100%;
+    border-collapse: collapse;
+    color: var(--reactist-table-content, var(--reactist-content-primary));
+    font-family: var(--reactist-font-family);
+    font-size: var(--reactist-font-size-body);
+}
+
+.header {
+    border-bottom: 1px solid var(--reactist-table-divider, var(--reactist-divider-secondary));
+}
+
+.headerCell {
+    padding: var(--reactist-spacing-medium);
+    text-align: left;
+    user-select: none;
+}
+
+.headerCellSortable {
+    padding: 0;
+}
+
+.sortButton {
+    display: flex;
+    gap: var(--reactist-spacing-xsmall);
+    align-items: center;
+    width: 100%;
+    padding: var(--reactist-spacing-medium);
+    border: 0;
+    color: inherit;
+    font: inherit;
+    font-weight: var(--reactist-font-weight-medium);
+    text-align: left;
+    background: transparent;
+    cursor: pointer;
+}
+
+.sortButton:hover {
+    background-color: var(--reactist-table-header-hover, var(--reactist-bg-highlight));
+}
+
+.sortButton:focus-visible {
+    outline: 2px solid
+        var(--reactist-table-focus-ring, var(--reactist-actionable-primary-idle-fill));
+    outline-offset: -2px;
+}
+
+.headerCellContent {
+    display: inline-flex;
+    align-items: center;
+}
+
+.row {
+    border-bottom: 1px solid var(--reactist-table-divider, var(--reactist-divider-secondary));
+}
+
+.row.clickable {
+    cursor: pointer;
+}
+
+.header:has(+ tbody > .row[aria-selected='true']:first-child) {
+    border-bottom-color: transparent;
+}
+
+.row[aria-selected='true'],
+.row:has(+ .row[aria-selected='true']) {
+    border-bottom-color: transparent;
+}
+
+.row[aria-selected='true'] {
+    background-color: var(--reactist-table-selected-fill, var(--reactist-framework-fill-selected));
+}
+
+.row[aria-selected='true'] .cell:first-child {
+    border-start-start-radius: var(--reactist-border-radius-large);
+    border-end-start-radius: var(--reactist-border-radius-large);
+}
+
+.row[aria-selected='true'] .cell:last-child {
+    border-start-end-radius: var(--reactist-border-radius-large);
+    border-end-end-radius: var(--reactist-border-radius-large);
+}
+
+.row.clickable:focus-visible {
+    outline: 2px solid
+        var(--reactist-table-focus-ring, var(--reactist-actionable-primary-idle-fill));
+    outline-offset: -2px;
+}
+
+.cell {
+    padding: var(--reactist-spacing-small) var(--reactist-spacing-medium);
+    vertical-align: middle;
+}
+
+.emptyCell {
+    padding-block: var(--reactist-spacing-xxlarge);
+    color: var(--reactist-table-empty-content, var(--reactist-content-secondary));
+    text-align: center;
+}
+
+.sortIndicator {
+    display: inline-flex;
+    align-items: center;
+    width: 16px;
+    height: 16px;
+}
+
+.sortIndicatorIcon {
+    width: 16px;
+    height: 16px;
+    color: var(--reactist-table-sort-indicator, var(--reactist-content-secondary));
+    fill: none;
+    stroke: currentcolor;
+    stroke-width: 1.5;
+    stroke-linecap: round;
+    stroke-linejoin: round;
+}
--- a/src/table/table.test.tsx
+++ b/src/table/table.test.tsx
@@ -1,0 +1,194 @@
+import * as React from 'react'
+
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import { axe } from 'jest-axe'
+
+import { Table, TableCell, TableRow } from './'
+
+import type { TableColumn } from './'
+
+type Person = {
+    id: string
+    name: string
+    role: string
+}
+
+const people: Person[] = [
+    { id: 'person-1', name: 'Avery Morgan', role: 'Product designer' },
+    { id: 'person-2', name: 'Sam Rivera', role: 'Engineer' },
+]
+
+const columns: TableColumn<Person>[] = [
+    { accessorKey: 'name', header: 'Name' },
+    { accessorKey: 'role', header: 'Role' },
+]
+
+describe('Table', () => {
+    it('renders typed columns and data with native table semantics', () => {
+        render(<Table data={people} columns={columns} getRowId={(person) => person.id} />)
+
+        expect(screen.getByRole('table')).toBeInTheDocument()
+        expect(screen.getAllByRole('columnheader')).toHaveLength(2)
+        expect(screen.getAllByRole('row')).toHaveLength(3)
+        expect(screen.getByRole('cell', { name: 'Avery Morgan' })).toBeInTheDocument()
+        expect(screen.getByRole('cell', { name: 'Engineer' })).toBeInTheDocument()
+    })
+
+    it('exposes controlled sorting through labeled header buttons', async () => {
+        const user = userEvent.setup()
+        const onSort = jest.fn()
+
+        render(
+            <Table
+                data={people}
+                columns={columns}
+                sorting={{ columnId: 'name', direction: 'asc' }}
+                onSort={onSort}
+                getSortAriaLabel={({ columnId, direction }) =>
+                    direction ? `${columnId}, sorted ${direction}` : `${columnId}, activate to sort`
+                }
+            />,
+        )
+
+        expect(screen.getByRole('columnheader', { name: 'Name' })).toHaveAttribute(
+            'aria-sort',
+            'ascending',
+        )
+        expect(screen.getByRole('columnheader', { name: 'Role' })).toHaveAttribute(
+            'aria-sort',
+            'none',
+        )
+
+        await user.click(screen.getByRole('button', { name: 'name, sorted asc' }))
+        screen.getByRole('button', { name: 'role, activate to sort' }).focus()
+        await user.keyboard('{Enter}')
+
+        expect(onSort).toHaveBeenNthCalledWith(1, 'name')
+        expect(onSort).toHaveBeenNthCalledWith(2, 'role')
+        expect(onSort).toHaveBeenCalledTimes(2)
+    })
+
+    it('renders empty-state content across all visible columns', () => {
+        render(<Table data={[]} columns={columns} emptyState="No people yet" />)
+
+        expect(screen.getByRole('cell', { name: 'No people yet' })).toHaveAttribute('colspan', '2')
+    })
+
+    it('supports consumer-wired row activation, selection, and custom cells', async () => {
+        const user = userEvent.setup()
+
+        function SelectablePeopleTable() {
+            const [selectedId, setSelectedId] = React.useState<string | null>(null)
+
+            return (
+                <Table
+                    data={people}
+                    columns={columns}
+                    getRowId={(person) => person.id}
+                    renderRow={(row) => (
+                        <TableRow
+                            key={row.id}
+                            aria-selected={selectedId === row.id}
+                            tabIndex={0}
+                            onClick={() => setSelectedId(row.id)}
+                            onKeyDown={(event) => {
+                                if (event.key === 'Enter' || event.key === ' ') {
+                                    event.preventDefault()
+                                    setSelectedId(row.id)
+                                }
+                            }}
+                        >
+                            {row.getVisibleCells().map((cell) => (
+                                <TableCell key={cell.id} cell={cell}>
+                                    {cell.column.id === 'name' ? (
+                                        <strong>{row.original.name}</strong>
+                                    ) : undefined}
+                                </TableCell>
+                            ))}
+                        </TableRow>
+                    )}
+                />
+            )
+        }
+
+        render(<SelectablePeopleTable />)
+
+        const averyRow = screen.getByRole('row', { name: 'Avery Morgan Product designer' })
+        const samRow = screen.getByRole('row', { name: 'Sam Rivera Engineer' })
+
+        await user.click(averyRow)
+        expect(averyRow).toHaveAttribute('aria-selected', 'true')
+
+        samRow.focus()
+        await user.keyboard(' ')
+        expect(samRow).toHaveAttribute('aria-selected', 'true')
+        expect(screen.getByText('Avery Morgan', { selector: 'strong' })).toBeInTheDocument()
+    })
+
+    it('forwards refs and styling escape hatches from the public building blocks', () => {
+        const tableRef = React.createRef<HTMLTableElement>()
+        const rowRef = React.createRef<HTMLTableRowElement>()
+        const cellRef = React.createRef<HTMLTableCellElement>()
+
+        render(
+            <Table
+                ref={tableRef}
+                data={people.slice(0, 1)}
+                columns={columns}
+                exceptionallySetClassName="custom-table"
+                renderRow={(row) => (
+                    <TableRow key={row.id} ref={rowRef} exceptionallySetClassName="custom-row">
+                        {row.getVisibleCells().map((cell, index) => (
+                            <TableCell
+                                key={cell.id}
+                                ref={index === 0 ? cellRef : undefined}
+                                cell={cell}
+                                exceptionallySetClassName="custom-cell"
+                            />
+                        ))}
+                    </TableRow>
+                )}
+            />,
+        )
+
+        expect(tableRef.current).toHaveClass('custom-table')
+        expect(rowRef.current).toHaveClass('custom-row')
+        expect(cellRef.current).toHaveClass('custom-cell')
+    })
+
+    it('applies inert to custom cells without relying on React version support', () => {
+        render(
+            <Table
+                data={people.slice(0, 1)}
+                columns={columns}
+                renderRow={(row) => (
+                    <TableRow key={row.id}>
+                        {row.getVisibleCells().map((cell) => (
+                            <TableCell key={cell.id} cell={cell} inert />
+                        ))}
+                    </TableRow>
+                )}
+            />,
+        )
+
+        expect(screen.getByRole('cell', { name: 'Avery Morgan' })).toHaveAttribute('inert')
+        expect(screen.getByRole('cell', { name: 'Product designer' })).toHaveAttribute('inert')
+    })
+
+    it('has no automated accessibility violations in its sortable state', async () => {
+        const { container } = render(
+            <Table
+                data={people}
+                columns={columns}
+                sorting={{ columnId: 'name', direction: 'desc' }}
+                onSort={jest.fn()}
+                getSortAriaLabel={({ columnId, direction }) =>
+                    direction ? `${columnId}, sorted ${direction}` : `${columnId}, sortable`
+                }
+            />,
+        )
+
+        expect(await axe(container)).toHaveNoViolations()
+    })
+})
--- a/src/table/table.tsx
+++ b/src/table/table.tsx
@@ -1,0 +1,316 @@
+'use no memo'
+
+import * as React from 'react'
+
+import { flexRender, getCoreRowModel, useReactTable } from '@tanstack/react-table'
+import classNames from 'classnames'
+import { useMergeRefs } from 'use-callback-ref'
+
+import styles from './table.module.css'
+
+import type { Cell, ColumnDef, Row } from '@tanstack/react-table'
+import type { ObfuscatedClassName } from '../utils/common-types'
+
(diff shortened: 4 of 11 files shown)


src/table/table.tsx
```diff
@@ -0,0 +1,316 @@
...
+        <table
+            {...tableProps}
+            ref={ref}
+            className={classNames(styles.table, exceptionallySetClassName)}
+        >
+            <thead className={styles.header}>
+                {table.getHeaderGroups().map((headerGroup) => (
+                    <tr key={headerGroup.id} className={styles.row}>
+                        {headerGroup.headers.map((header) => {
+                            const canSort = Boolean(onSort && getSortAriaLabel)
```
doistbot: <a href="#">P2</a>  Don't make every header sortable. `TableColumn` permits TanStack group columns, which generate group and placeholder headers. When sorting props are supplied, this renders sort buttons (and `aria-sort`) for empty placeholders and group labels, and calls `onSort` with group IDs rather than data-column IDs. Gate sorting to non-placeholder leaf headers, or explicitly disallow grouped column definitions.

src/table/table.tsx
```diff
@@ -0,0 +1,316 @@
...
+                                    {canSort && onSort && getSortAriaLabel ? (
+                                        <button
+                                            type="button"
+                                            className={styles.sortButton}
+                                            aria-label={getSortAriaLabel({
+                                                columnId: header.column.id,
+                                                direction,
+                                            })}
+                                            onClick={() => onSort(header.column.id)}
+                                            onKeyDown={(event) => {
```
doistbot: <a href="#">P2</a>  Native `<button>` elements already trigger `onClick` on `Enter` (keydown) and `Space` (keyup). Intercepting `onKeyDown` and calling `onSort` immediately on `Space` keydown bypasses standard button keyboard behavior, where activation happens on keyup and can be cancelled. Removing `onKeyDown` lets native button activation handle both pointer and keyboard events cleanly.

src/table/table.module.css
```diff
@@ -0,0 +1,117 @@
...
+
+.row {
+    border-bottom: 1px solid var(--reactist-table-divider, var(--reactist-divider-secondary));
+}
+
+.row.clickable {
+    cursor: pointer;
+}
+
+.header:has(+ tbody > .row[aria-selected='true']:first-child) {
```
doistbot: <a href="#">P2</a>  The `thead` element has `.header` and its child `tr` has `.row`, both of which define `border-bottom: 1px solid ...`. When the first row is selected, this selector sets `border-bottom-color: transparent` only on `.header` (`thead`). In CSS `border-collapse`, the row's (`tr.row`) border takes precedence over the row group's (`thead`), so the divider between the header and the first row remains visible. To fix this, remove `.row` from the header `tr` (since `.header` already supplies the border) or include `.header:has(...) .row` in this rule.

src/table/table.module.css
```diff
@@ -0,0 +1,117 @@
...
+.row[aria-selected='true'],
+.row:has(+ .row[aria-selected='true']) {
+    border-bottom-color: transparent;
+}
+
+.row[aria-selected='true'] {
+    background-color: var(--reactist-table-selected-fill, var(--reactist-framework-fill-selected));
+}
+
+.row[aria-selected='true'] .cell:first-child {
```
doistbot: <a href="#">P2</a>  Under `border-collapse: collapse` (set in `.table`), browsers ignore `border-radius` on internal table elements (`td`/`tr`), which prevents the selected row's start and end cells from rendering with rounded corners. Using `border-collapse: separate; border-spacing: 0;` allows `border-radius` on `.cell:first-child` and `.cell:last-child` to render properly.

package.json
```diff
@@ -155,6 +155,7 @@
     },
     "dependencies": {
         "@babel/runtime": "^7.28.4",
+        "@tanstack/react-table": "8.21.3",
```
doistbot: <a href="#">P2</a>  `@tanstack/react-table` is in `dependencies` (exact-pinned to `8.21.3`), but the public API re-exports TanStack types (`TableColumn = ColumnDef`, `TableRowModel = Row`, `TableCellModel = Cell`) and the PR explicitly invites consumers to use TanStack directly. As a regular dependency, consumers who also install `@tanstack/react-table` get a second nested copy, and the re-exported types originate from reactist's nested instance — leading to bundle duplication and type friction across versions. Mirror the `@ariakit/react` precedent (a peerDependency whose types also surface in the public API) and move this to `peerDependencies` with a range such as `^8.21.3`.

src/table/table.tsx
```diff
@@ -0,0 +1,316 @@
...
+        renderRow,
+        sorting,
+        onSort,
+        getSortAriaLabel,
+        exceptionallySetClassName,
+        ...tableProps
+    }: TableProps<TData>,
+    ref: React.ForwardedRef<HTMLTableElement>,
+) {
+    const table = useReactTable({
```
doistbot: <a href="#">P1</a>  Keep the TanStack model out of the core Table API. Constructing it here means consumers cannot use another modelling layer: `renderRow` receives only this internally-created `Row`, and `TableCell` requires its `Cell`. This contradicts the stated compositional API and permanently exposes `@tanstack/react-table` types/dependency from Reactist. Make the core table compositional native-table primitives, and keep TanStack conversion in an integration adapter or story.

src/table/table.tsx
```diff
@@ -0,0 +1,316 @@
...
+        inert = false,
+        exceptionallySetClassName,
+        ...cellProps
+    }: TableCellProps<TData>,
+    ref: React.ForwardedRef<HTMLTableCellElement>,
+) {
+    const internalRef = React.useRef<HTMLTableCellElement>(null)
+    const combinedRef = useMergeRefs([internalRef, ref])
+
+    React.useEffect(
```
doistbot: <a href="#">P2</a>  This effect runs for every `TableCell`, including the default `inert={false}` path, so mounting a large table schedules one effect and a `toggleAttribute('inert', false)` DOM operation per cell. Split the inert DOM-sync behavior into an inert-only helper/component so ordinary cells remain plain `<td>` elements and only opted-in cells pay for the ref/effect.

src/table/table.test.tsx
```diff
@@ -0,0 +1,194 @@
...
+                    <TableRow key={row.id}>
+                        {row.getVisibleCells().map((cell) => (
+                            <TableCell key={cell.id} cell={cell} inert />
+                        ))}
+                    </TableRow>
+                )}
+            />,
+        )
+
+        expect(screen.getByRole('cell', { name: 'Avery Morgan' })).toHaveAttribute('inert')
```
doistbot: <a href="#">P2</a>  `getByRole` excludes `inert` subtrees by default (same mechanism as `aria-hidden`, cf. `src/banner/banner.test.tsx` using `{ hidden: true }`). Once `TableCell`'s `useEffect` sets the `inert` attribute, these cells are no longer exposed by role queries, so `getByRole('cell', ...)` will throw rather than return the element. Query the DOM directly (`container.querySelectorAll('td')`) or pass `{ hidden: true }`. Better: assert `queryByRole('cell', { name: 'Avery Morgan' })` is null and then check the attribute on the DOM node, so the test verifies the cell is actually hidden from assistive tech.

doistbot (commented):
This PR adds a compositional `Table` component with TanStack-powered sorting, selection, and clickable rows, plus Storybook docs and tests.

Few things worth tightening:
- **TanStack coupling:** The core API passes TanStack `Row`/`Cell` types through `renderRow` and `TableCell`, which locks consumers into one modelling layer and contradicts the compositional design. Keep the core as native-table primitives and move TanStack to an integration adapter or story. Also move `@tanstack/react-table` from `dependencies` to `peerDependencies` to avoid duplicate nested copies for consumers who install it directly.
- **Sort header issues:** Sorting currently activates on TanStack group and placeholder headers, rendering sort buttons and `aria-sort` on empty headers and calling `onSort` with group IDs — gate it to non-placeholder leaf headers. Additionally, the `onKeyDown` handler on the native `<button>` duplicates built-in Enter/Space activation and double-fires `onSort` on Space; remove it and let the browser handle keyboard activation.
- **CSS fixes:** Under `border-collapse: collapse`, `border-radius` on selected-row edge cells is ignored — switch to `border-collapse: separate; border-spacing: 0` so rounded corners render. The `:has` rule that clears the header divider for a selected first row only targets `thead.header`, not the child `tr.row` whose border takes precedence, so the divider stays visible — either drop `.row` from the header `tr` or extend the selector.
- **Inert handling and tests:** The `useEffect` that syncs the `inert` attribute runs on every `TableCell`, including the default non-inert path, adding a per-cell effect and DOM mutation to large tables — split it into an opt-in wrapper so ordinary cells stay plain `<td>`. Tests also query inert cells with `getByRole('cell')`, which excludes inert subtrees and will throw; use `{ hidden: true }` or direct DOM queries, or assert the cell is absent from the accessibility tree and then check the attribute on the node.

<details>

- **<a href="#">P3</a> src/table/table.tsx:1**: The `'use no memo'` directive opts the entire file out of React Compiler optimization and records a `CompileError` in `.react-compiler.rec.json`. Per repository guidelines, new components should be compiler-compatible. If there are no compiler violations when removing the directive, remove `'use no memo'` so the compiler can optimize the component.
- **<a href="#">P3</a> src/table/table.module.css:4**: The `--reactist-table-*` custom properties (content, divider, header-hover, focus-ring, selected-fill, empty-content, sort-indicator) are consumed with fallbacks but never defined in a `:root` block. AGENTS.md states component-specific tokens should go in the component's own CSS file under `:root`, and every other component follows this (e.g. `src/badge/badge.module.css`, `src/banner/banner.module.css`). Defining them in `:root` with their fallback defaults makes the customization surface discoverable in one place and matches the established pattern.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1102&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A941afa10-9b9e-11f1-9643-4a1a85688552%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A941afa10-9b9e-11f1-9643-4a1a85688552&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/table/table.tsx
```diff
@@ -0,0 +1,266 @@
...
+TableRow.displayName = 'TableRow'
+TableCell.displayName = 'TableCell'
+TableColumnHeader.displayName = 'TableColumnHeader'
+
+export {
+    Table,
+    TableBody,
+    TableCell,
+    TableColumn,
+    TableColumnGroup,
```
doistbot, line 254: <a href="#">P2</a>  The component's prop types (`TableProps`, `TableHeaderProps`, `TableBodyProps`, `TableRowProps`, `TableCellProps`, `TableColumnGroupProps`, `TableColumnProps`, `TableColumnHeaderProps`, `TableColumnWidth`) are defined but never exported, so `export * from './table'` in `src/index.ts` ships the components without their types. AGENTS.md requires exporting types explicitly (`export type { ButtonProps }`), and other components such as `Banner` follow this. Add an `export type { ... }` block here so consumers can type their own composition wrappers.

src/table/table.module.css
```diff
@@ -0,0 +1,167 @@
...
+    color: var(--product-library-display-tertiary-idle-tint);
+    opacity: 0;
+}
+
+.sortIndicatorDescending {
+    transform: rotate(180deg);
+}
+
+.align-start {
+    text-align: left;
```
doistbot: <a href="#">P2</a>  `align="start"`/`"end"` is a logical, RTL-aware API, but `.align-start` sets `text-align: left` and `.align-end` sets `text-align: right`. In RTL, `start` means right, so both `TableCell` and `TableColumnHeader` with `align="start"` render left-aligned and `align="end"` renders right-aligned — the opposite of the prop's meaning. Use `text-align: start` / `text-align: end` (already the pattern in `src/box/box.module.css`).

src/table/table.tsx
```diff
@@ -0,0 +1,266 @@
...
+            <th
+                {...headerProps}
+                ref={ref}
+                aria-sort={sortable ? ariaSortFor(sortDirection) : undefined}
+                className={classNames(
+                    styles.headerCell,
+                    styles[`align-${align}`],
+                    sortable && styles.headerCellSortable,
+                    exceptionallySetClassName,
+                )}
```
doistbot, line 221: <a href="#">P2</a>  Only expose `aria-sort` on the currently sorted header. When a table has several sortable but unsorted columns, this sets `aria-sort="none"` on every header, despite ARIA guidance to apply it to just one header. Leave it undefined when `sortDirection` is `null`; the button label already conveys that the column can be sorted.

src/table/table.tsx
```diff
@@ -0,0 +1,266 @@
...
+type TableRowProps = Omit<React.HTMLAttributes<HTMLTableRowElement>, 'className'> &
+    ObfuscatedClassName
+
+type TableCellProps = Omit<React.TdHTMLAttributes<HTMLTableCellElement>, 'align' | 'className'> &
+    ObfuscatedClassName & {
+        /** Horizontal alignment of the cell content. */
+        align?: 'start' | 'end'
+    }
+
+type TableColumnWidth =
```
doistbot, line 31: <a href="#">P2</a>  `TableColumnWidth` is an exact copy of the exported `ColumnWidth` from `src/columns` (same union and `width` prop semantics), and the mapping at line 138 (`styles[`columnWidth-${width.replace('/', '-')}`]`) reimplements the established pattern already used in `columns.tsx`: `getClassNames(styles, 'columnWidth', width.replace('/', '-'))`. Import `ColumnWidth` and use `getClassNames` from `../utils/responsive-props` so the width options and class-name convention can't drift between the two components.

doistbot (commented):
Adds a compositional, native-HTML `Table` component with primitives for header, body, rows, and cells, plus sort-indicator support and TanStack integration stories.

<details>

- **<a href="#">P3</a> src/table/table.module.css:3**: Route these Product Library values through `--reactist-table-*` variables declared in this file's `:root`, then consume those component variables throughout. The Table currently has no Reactist-specific token surface, so consumers cannot override its defaults independently and the styles bypass the repository token convention.
- **<a href="#">P3</a> src/table/table.stories.tsx:157**: `Default` and `NarrowViewport` render byte-identical four-column people tables, and `SelectedAndClickableRows` repeats the same markup with only the row `aria-selected`/`tabIndex`/handlers added. Extract one `renderPeopleTable()` helper (or a `PeopleTable` component) and reuse it across all three to cut ~120 duplicated lines without changing what each story demonstrates.
- **<a href="#">P3</a> src/table/table.stories.tsx:442**: The sortable and non-sortable branches of this ternary duplicate the `<TableColumnHeader>` wrapper and its `<Text>`/`flexRender` children. Hoist the rendered header content into a variable (`const content = <Text variant="body-2">{flexRender(...)}</Text>`) and only vary the sortable props between branches.
- **<a href="#">P3</a> src/table/table.test.tsx:76**: This test verifies that `TableCell` is transparent by checking CSS class names on the child `Text` component (`variant-callout-2`, `lineClamp-1`). Those classes belong to `Text`, not `TableCell` — if `Text` changes its CSS module naming, this table test breaks for no table-related reason. The `parentElement?.tagName` assertion alone already proves the TD wrapping. Drop the className assertions on the child, or replace them with a simpler child (e.g. a plain `<span data-testid="child">`) and assert it renders directly inside the `td`.

</details>

<sub>[Share Feedback](https://github.com/Doist/doistbot-app/issues/new?template=review-feedback.yml&title=Review+feedback+for+Doist%2Freactist&pr_url=https%3A%2F%2Fgithub.com%2FDoist%2Freactist%2Fpull%2F1102&datadog_url=https%3A%2F%2Fapp.datadoghq.com%2Flogs%3Fquery%3Dservice%253A%2528doistbot-app-sandbox%2520OR%2520doistbot-app%2529%2520%2540meta.deliveryId%253A43ece850-9ba3-11f1-8676-801626ae152d%26agg_m%3Dcount%26agg_m_source%3Dbase%26agg_t%3Dcount%26clustering_pattern_field_path%3Dmessage%26cols%3Dhost%252Cservice%26fromUser%3Dtrue%26messageDisplay%3Dinline%26refresh_mode%3Dsliding%26storage%3Dhot%26stream_sort%3Ddesc%26viz%3Dstream%26live%3Dtrue) • [Review Logs](https://app.datadoghq.com/logs?query=service%3A%28doistbot-app-sandbox%20OR%20doistbot-app%29%20%40meta.deliveryId%3A43ece850-9ba3-11f1-8676-801626ae152d&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&live=true)</sub>

src/table/table.tsx
frankieyan: No need for this right now. It's the same but they don't need to be.

src/table/table.tsx
frankieyan: ✅ c50e52a4

src/table/table.module.css
frankieyan: ✅ e0ed6343

src/table/table.tsx
frankieyan: ✅ d750fb0d

scottlovegrove (approved):

src/table/table.module.css
```diff
@@ -0,0 +1,167 @@
+.table {
+    width: 100%;
+    color: var(--product-library-display-primary-idle-tint);
+    font-family: var(--reactist-font-family);
```
pawelgrimm: Do we really need to specify the font family and color here? Can't we just rely on whatever is set by the consumer?

src/table/table.tsx
```diff
@@ -0,0 +1,275 @@
...
+    { exceptionallySetClassName, ...tableProps },
+    ref,
+) {
+    return (
+        <table
+            {...tableProps}
+            ref={ref}
+            className={classNames(styles.table, exceptionallySetClassName)}
+        />
+    )
```
pawelgrimm, line 111: We should use Box for the width and bg color. I'm sure you could do something similar for many of the components here.

```suggestion
    return (
        <Box
            ref={ref}
            as="table"
            width="full"
            background="primary"
            {...tableProps}
            className={classNames(styles.table, exceptionallySetClassName)}
        />
    )
```

src/table/table.module.css
```diff
@@ -0,0 +1,167 @@
...
+    vertical-align: middle;
+}
+
+.cell:first-child {
+    padding-left: var(--reactist-spacing-medium);
+}
+
+.cell:last-child {
+    padding-right: var(--reactist-spacing-medium);
+}
```
pawelgrimm, line 91: 💭 `padding-inline` has better RTL support (you know, just in case 😅):

```suggestion
.cell:first-child {
    padding-inline-start: var(--reactist-spacing-medium);
}

.cell:last-child {
    padding-inline-end: var(--reactist-spacing-medium);
}
```

src/table/table.tsx
```diff
@@ -0,0 +1,275 @@
...
+)
+
+Table.displayName = 'Table'
+TableColumnGroup.displayName = 'TableColumnGroup'
+TableColumn.displayName = 'TableColumn'
+TableHeader.displayName = 'TableHeader'
+TableBody.displayName = 'TableBody'
+TableRow.displayName = 'TableRow'
+TableCell.displayName = 'TableCell'
+TableColumnHeader.displayName = 'TableColumnHeader'
```
pawelgrimm: 💭 Why do we need to be explicit about setting `displayName`? Isn't that automatic?

src/table/table.tsx
```diff
@@ -0,0 +1,275 @@
...
+          sortDirection: 'asc' | 'desc' | null
+
+          /** Called when the sort control is activated. */
+          onSort: () => void
+
+          /** Complete localized label for the sort button. */
+          sortAriaLabel: string
+
+          /** Render a custom sort indicator icon. It should default to ascending and pointing upwards */
+          sortIcon?: React.ReactNode
```
pawelgrimm: 💭 Do we actually expect consumers to set custom icons per column? Or is this more to override the default, in which case, perhaps it would be better to set it via a Context?

src/table/table.tsx
```diff
@@ -0,0 +1,275 @@
...
+                        onClick={onSort}
+                    >
+                        {label}
+                        <span
+                            aria-hidden="true"
+                            className={classNames(
+                                styles.sortIndicator,
+                                indicatorDirection === 'desc' && styles.sortIndicatorDescending,
+                                !sortDirection && styles.sortIndicatorUnsorted,
+                            )}
```
pawelgrimm: 💭 IMHO, object form is more readable:

```suggestion
                            className={classNames(
                                styles.sortIndicator,
                                {
                                  [styles.sortIndicatorDescending]: indicatorDirection === 'desc',
                                  [styles.sortIndicatorUnsorted]: !sortDirection
                                },
                            )}
```

src/table/table.tsx
frankieyan: Honestly, probably not because this already comes from the product library, but I added it because of the conversation we had at the retreat. Let's take it out

src/table/table.tsx
frankieyan: I was torn on this one because of the feedback about splitting between Box and a classname being less readable, but splitting does give us better visibility on what Box can't achieve, even if not everything feel like they would belong to Box.

src/table/table.tsx
frankieyan: I kept them because we needed them for Box etc's stories: https://github.com/Doist/reactist/pull/1092

But it looks like that's a combination of using `polymorphicComponent` and arg stories, so they are safe to remove here.

src/table/table.module.css
frankieyan: It seems like only some of our apps sets a body colour, but in most cases I can see, the components themselves sets it, e.g. Text, form fields, prose, settings, search, etc. I guess it won't hurt to remove (if we add the default back to Comms), but we should also add the colour to storybook's preview.ts.

frankieyan: c8351005427642f873ccd0399ccc6aaa602fac1f

src/table/table.tsx
frankieyan: ef410edcb351cb156e9d7a9648dd0d238a59f215

src/table/table.module.css
frankieyan: c446b67fb880aebc14f9ad2620ed0c319a371a2a

src/table/table.tsx
frankieyan: 2545798e6497dd569e3b1936d6191874fd3c4d23

src/table/table.tsx
frankieyan: b7e5b581ebe0bf6bb365294bbc5d3629eb1ff0fe

It does complain about styles.x being potentially undefined, and the react compiler doesn't like `??` in computed properties, so the classnames have to be hoisted.

src/table/table.tsx
frankieyan: 5dc2f637b395105f5b6cbee7dd7f740fe67ba760

